PR #562, opened 2014-12-05
(gh-563) openstack provider for vagrant

mcanevet:


kevpl:
unfortunately, the tests are too old, so we'll have to have this retested.

kevpl:

kevpl:

anodelman:
Any reason that you can't just use our openstack provider, instead of adding a vagrant openstack provider?  We would prefer for openstack to be accessed through the library that already exists.  Is there some missing functionality?

mcanevet:
@anodelman I just can't get it work but I couldn't figure out why. Is it really working, because when trying to use it I found this bug https://github.com/puppetlabs/beaker/pull/564 that clearly prevent it to work properly?

stefanmortensen:
I got a running Beaker setup with OpenStack as a hypervisor after applying the network patch above and #621. My setup also assigned IPs from a different pool name than default which I fixed  by adding a parameter for the pool name to use with #619. Let me know what issues you are having, maybe I can help?

mcanevet:
@stefanmortensen is it possible to use OS_* environment variables with the native hypervisor? IIRC I was stuck with that.

lib/beaker/host_prebuilt_steps.rb
```diff
@@ -338,7 +338,7 @@ def enable_root_login host, opts
         if host['platform'] =~ /debian|ubuntu|cumulus/
           host.exec(Command.new("sudo su -c \"service ssh restart\""), {:pty => true})
         elsif host['platform'] =~ /centos|el-|redhat|fedora|eos/
-          host.exec(Command.new("sudo -E /sbin/service sshd restart"))
+          host.exec(Command.new("sudo -E /sbin/service sshd restart"), {:pty => true})
```
anodelman: I think that this could be split out as a separate patch, away from the vagrant-openstack request.

stefanmortensen:
@mcanevet no, you need to define that variables in your nodeset file. Mine look something like this:

HOSTS:
  ubuntu-1404-x64-master:
    platform: ubuntu-1404-x86_64
    hypervisor: openstack
    ip: 192.168.20.20
    flavor: m1.medium
    image: "Ubuntu 14.04"
    user: ubuntu
CONFIG:
   openstack_api_key: openstack_password
   openstack_username: openstack_user
   openstack_auth_url: http://127.0.0.1:5000/v2.0/tokens
   openstack_tenant: beaker
   openstack_network: beaker-net
   openstack_keyname: beaker-key

Would be kind of nice though to check the vars like username, key and auth url from env variable if they are unset in the nodeset file.

anodelman:
We are only going to be supporting one means of communicating with openstack and will not be merging vagrant-openstack support.  Any missing configuration options should be added to the beaker openstack hypervisor code.

mcanevet:
@stefanmortensen https://github.com/puppetlabs/beaker/pull/643
I'm still stuck on `copy_ssh_to_root` though...
I can't figure out how to use key pair.
If I specify an existing keypair in `openstack_keyname` I get `Invalid key_name provided.`

PR #571, opened 2014-12-11
WIP: (PDB-1034) Ezbake changes to get PDB source based builds working

kbarber:
This involves a rework of the ezbake_utils to provide more capability to fully
install PuppetDB using these helpers.

* In general, the main API for install_from_ezbake has been kept as it is, but
  some arguments are now optional for convenienence
* We now do the lein install process, but using a local maven repository to
  avoid collision with other projects running on the same host.
* install_termini_from_ezbake now added to install that component
* The make interface on ezbake has been cleaned up somewhat, so reduce the
  amount of extra work the helper has to do. This patch reflects those changes.
* Validate for support up front, this makes the tests easier and the failure
  more centralized.
* Automated the retrieval of project name and version if not provided to the
  helpers. This suits the case where a project is a singleton as opposed to
  a composition.
* ezbake_local_cmd has been added to generalize this kind of system invocation
  and simplify testing.
* ezbake_make generalizes the way we invoke make, and simplifies testing.
* install_ezbake_tarball_on_host has been created to generalize this step for
  the termini and service based installations to re-use. Its idempotent, in
  that it checks to ensure its not already installed.
* conditionally_clone was modified to support being passed a branch if required,
  so in the future we can simplify working on a development branch.
* Lots of yarddoc cleanups
* Full rspec coverage

### DIFF
--- a/lib/beaker/dsl/ezbake_utils.rb
+++ b/lib/beaker/dsl/ezbake_utils.rb
@@ -318,9 +318,8 @@ module Beaker
         # project_param_string if its provided? I'm guessing this is
         # true for other projects, but we should allow nil for the
         # sake of all-in-one projects like PDB.
-        ezbake_local_cmd "#{lein_prefix} with-profile ezbake ezbake build",
+        ezbake_local_cmd "#{lein_prefix} with-profile ezbake ezbake stage",
                          :throw_on_failure => true
-
 
         #staging_dir = File.join(ezbake_dir, 'target', 'staging')
         staging_dir = File.join('target','staging')
--- a/lib/beaker/dsl/ezbake_utils.rb
+++ b/lib/beaker/dsl/ezbake_utils.rb
@@ -282,7 +282,8 @@ module Beaker
         ezbake_tools_available?
         #conditionally_clone "gitmirror@github.delivery.puppetlabs.net:puppetlabs-ezbake.git",
         #                    ezbake_dir
-        # TODO: will need removal before merging
+        # TODO: will need removal before merging, since we'll want to
+        # work against a shipped version of ezbake
         conditionally_clone "git@github.com:kbarber/ezbake.git",
                             ezbake_dir, "pdb-1034-ezbake-pr-testing"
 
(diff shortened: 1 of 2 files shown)


lib/beaker/dsl/ezbake_utils.rb
```diff
@@ -117,27 +116,90 @@ def install_ezbake_deps host
...
 
         else
-          raise "No repository installation step for #{variant} yet..."
+          raise RuntimeError, "No repository installation step for #{variant} yet..."
+        end
+      end
+
+      # @!endgroup
+
+      REMOTE_PACKAGES_REQUIRED = ['make']
```
anodelman: For sets of requirements we've moved everything into options/presets.rb - so that we can track and view them all together in a single location.  I'm wary of having these set in the middle of a file where it would be difficult to monitor and can't be included in any options parsing.

lib/beaker/dsl/ezbake_utils.rb
```diff
@@ -145,74 +207,202 @@ def install_from_ezbake host, project_name, project_param_string, env_args={}, e
...
+      #
+      # @param [String] cmd command to execute
+      # @param [Hash] opts options
+      # @option opts [bool] :throw_on_failure If true, throws an
+      #   exception if the exit code is non-zero. Defaults to false.
+      # @return [bool] true if exit == 0 false if otherwise
+      # @raise [RuntimeError] if :throw_on_failure is true and
+      #   command fails
+      # @api private
+      def ezbake_local_cmd cmd, opts={}
```
anodelman: I would agree that this is a bigger issue (a simple wrapper around system is never simple :) ).  This might be a better option as a helper library instead of part of mainline beaker - until we decide what we are trying to accomplish and do a proper build out.

kbarber:
Closing for now, as I plan on doing some biggish changes anyway going forward, and to save us from noise and testing load I'm going to close this for now. I'll re-open when I'm ready for this to be reviewed again. @anodelman I've noted your comments as well.

PR #595, opened 2014-12-29
QENG-1668 - Enable the parsing of environment variables and add more options to be parsed from the ENV

cyberious:


anodelman:
Please associate your commit with a jira bug number.

anodelman:
Last comment still current.

cyberious:
There is no ticket for this, if you would like I can create and associate.

anodelman:
Yes, please file and associate with your commit for future tracking.

lib/beaker/tasks/rake_task.rb
```diff
@@ -79,13 +81,21 @@ def check_for_beaker_type_config
...
+      # tests, hosts, preserver_hosts, fail_mode, provision and log_level
+      # if they are not previously set by the calling rake task.
+      #
       #
       def check_env_variables
         if File.exists?(File.join(DEFAULT_ACCEPTANCE_ROOT, 'tests'))
           @tests = File.join(DEFAULT_ACCEPTANCE_ROOT, 'tests')
         end
         @tests = ENV['TESTS'] || ENV['TEST'] if !@tests
+        @hosts = ENV['BEAKER_setfile'] || ENV['BEAKER_set'] if !@hosts
```
anodelman: My only concern here is created another set of beaker environment variables, this one specific to the Rakefile.  I think that it would be simpler to have a single env var to read in additional options, and then use that beaker command line option.  Then this won't need constant updating to increase configurability.

cyberious: I agree but we need the configurable state.  The inconsistency between BEAKER and Beaker-rspec throws me for a loop I know.

anodelman: We will be merging beaker and beaker-rspec in the near future.  For now, instead of increasing env var clutter, let's just support a single env var that points to a file of a configuration hash - then send it to beaker with --options-file.

anodelman:
Last code comment still current.

anodelman:
Last code comment still current.

anodelman:
Closing out, lack of activity.

PR #597, opened 2014-12-30
(QENG-144) created masterless option, and support for masterless installs

kevpl:


anodelman:
@kevpl I'll assume that this passed local masterless installation testing?

kevpl:
@anodelman yes, it has

lib/beaker/dsl/install_utils.rb
```diff
@@ -453,12 +453,16 @@ def deploy_frictionless_to_master(host)
       # @api private
       #
       def do_install hosts, opts = {}
```
anodelman: Yard docs needs to be updated here to describe the new option.

kevpl: @anodelman `do_install` actually doesn't take the masterless option, it overwrites it from the main options hash.  This happens a few lines below this comment.

PR #598, opened 2014-12-31
(MAINT) also ignore the 'bundle' directory

nrvale0:
If the 'bundle' directory is present, ignore it during setup/scp. Often has self-referential symlinks which do NOT play will with the scp which does a dereference of all symlinks. This results in infinite copy during setup. There's a similar PR which was merged in Sept for the '.bundle' directory but I've never seen bundler create a dir by that name so that ignore may be unnecessary.

### DIFF
--- a/lib/beaker/dsl/install_utils.rb
+++ b/lib/beaker/dsl/install_utils.rb
@@ -26,7 +26,7 @@ module Beaker
       GitHubSig   = 'github.com,207.97.227.239 ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAq2A7hRGmdnm9tUDbO9IDSwBK6TbQa+PXYPCPy6rbTrTtw7PHkccKrpp0yVhp5HdEIcKr6pLlVDBfOLX9QUsyCOV0wzfjIJNlGEYsdlLJizHhbn2mUjvSAHQqZETYP81eFzLQNnPHt4EVVUh7VfDESU84KezmD5QlWpXLmvU31/yMf+Se8xhHTvKSCZIFImWwoG6mbUoWf9nzpIoaSjB+weqqUUmpaaasXVal72J+UX2B+2RPW3RcT0eOzQgqlJL3RKrTJvdsjE3JEAvGq3lGHSZXy28G3skua2SmVi/w4yCE6gbODqnTWlg7+wC604ydGXA8VJiS5ap43JXiUFFAaQ=='
 
       # The directories in the module directory that will not be scp-ed to the test system when using `copy_module_to`
-      PUPPET_MODULE_INSTALL_IGNORE = ['.bundle', '.git', '.idea', '.vagrant', '.vendor', 'vendor', 'acceptance', 'spec', 'tests', 'log']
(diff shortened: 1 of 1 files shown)


anodelman:
Please preface your commit with (MAINT).  Also, needs a rebase.

nrvale0:
ACK @anodelman

lib/beaker/dsl/install_utils.rb
```diff
@@ -26,7 +26,7 @@ module InstallUtils
       GitHubSig   = 'github.com,207.97.227.239 ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAq2A7hRGmdnm9tUDbO9IDSwBK6TbQa+PXYPCPy6rbTrTtw7PHkccKrpp0yVhp5HdEIcKr6pLlVDBfOLX9QUsyCOV0wzfjIJNlGEYsdlLJizHhbn2mUjvSAHQqZETYP81eFzLQNnPHt4EVVUh7VfDESU84KezmD5QlWpXLmvU31/yMf+Se8xhHTvKSCZIFImWwoG6mbUoWf9nzpIoaSjB+weqqUUmpaaasXVal72J+UX2B+2RPW3RcT0eOzQgqlJL3RKrTJvdsjE3JEAvGq3lGHSZXy28G3skua2SmVi/w4yCE6gbODqnTWlg7+wC604ydGXA8VJiS5ap43JXiUFFAaQ=='
 
       # The directories in the module directory that will not be scp-ed to the test system when using `copy_module_to`
-      PUPPET_MODULE_INSTALL_IGNORE = ['.bundle', '.git', '.idea', '.vagrant', '.vendor', 'vendor', 'acceptance', 'spec', 'tests', 'log']
+      PUPPET_MODULE_INSTALL_IGNORE = ['.bundle', '.git', '.idea', '.vagrant', '.vendor', 'acceptance', 'bundle', 'spec', 'tests', 'log']
```
anodelman: Any reason here that you are dropping the 'vendor' directory check?

kevpl:
The reason `.bundle` is ignored is that we use that as our bundle path for projects at Puppet.  I believe it was copied from the [historical default location](http://bundler.io/v0.9/bundle_install.html), but allowed to be local to allow differences between our projects.

My recommendation for this wouldn't be to add more locations that do the same thing, but to standardize on `.bundle/gems`, so that we don't end up trying to put all of these eventualities in all of the places they might be mentioned in the code.  This can be set either using the `BUNDLE_PATH` env var, or using the `--path` argument to bundler itself like below

    bundle install --path .bundle/gems/

anodelman:
My code comment is still applicable here.

lib/beaker/dsl/install_utils.rb
nrvale0: Oversite on my part. Will fix.

anodelman: Still pending this change.

anodelman:
We'll give it another week, then closing this out for inactivity.

nrvale0:
@anodelman fixed and repushed

anodelman:
Thanks!

PR #601, opened 2015-01-02
(QENG-1627) added spec tests to flush out host and ssh_connection testin...

kevpl:
...g


kevpl:
failures unrelated.

anodelman:
Considering that this passed spec, we're good here.

PR #602, opened 2015-01-02
(QENG-1628) Add bash completion scripts for beaker

er0ck:
using this script, one can press tab after any portion (or none)
of a beaker option/command and bash will complete it or offer
suggestions.

### DIFF
--- a/ext/completion/beaker-completion.bash
+++ b/ext/completion/beaker-completion.bash
@@ -1,0 +1,53 @@
+## bash completion script for beaker
+
+# The contained completion routines provide support for completing:
+#
+#    *) any option specified in beaker --help
+#       but not yet the arguments to the options
+
+# To use these routines:
+#
+#    1) Copy this file to somewhere (e.g. ~/.beaker-completion.sh).
+#    2) Add the following line to your .bashrc/.zshrc:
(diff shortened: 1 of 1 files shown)


er0ck:
i realized over the weekend that i could probably add some tests for this.  merging this won't really harm anything and i've manually tested it pretty thoroughly on osx and rhel7.   thinking of adding tests using shUnit

kevpl:
Hi @er0ck. I was able to get this to work in bash, but not in zsh (`complete` doesn't appear to exist by default, maybe there's something I can install for that).  Please update your script comment to show for this.

@anodelman, what do you think of the directory structure?  I know we have the `bin` directory for beaker's runner, but do we already have a place for external scripts?  I didn't see anything else, and I think this seems like a fine structure to start for them.  What do you think?

er0ck:
i chose this directory structure as it's used by puppet for external stuff.  i'm not married to it.   i could see other completion scripts for csh, et al.  and other external scripts orthogonal to completions...  vim-syntax for instance :-)

anodelman:
I'm fine with the proposed directory structure.

anodelman:
Let's merge this in and get more eyes on it.

PR #604, opened 2015-01-05
(QENG-1631) allowed exit code 0 for puppet agent call for EOS during install

kevpl:


kevpl:

Issue #503, opened 2014-10-21
Beaker copying module to wrong directory


petems:
Trying to get my module with just one fact to check shellshock

Code here: https://github.com/petems/petems-shellshock_facts

This fails, which is weird because it's so simple, so I did:

`BEAKER_destroy=no bundle exec rspec spec/acceptance/class_spec.rb`

but creates the following in `/etc/puppet`:

```
root@ubuntu-server-12042-x64:/etc/puppet/modules# tree shellshock_facts/
shellshock_facts/
|-- Gemfile
|-- Gemfile.lock
|-- lib
|   |-- facter
|   |   |-- facter
|   |   |   `-- shellshock.rb
|   |   `-- shellshock.rb
|   `-- lib
|       `-- facter
|           `-- shellshock.rb
|-- LICENSE
|-- manifests
|   |-- init.pp
|   `-- manifests
|       `-- init.pp
|-- Modulefile
|-- Rakefile
|-- README
`-- README.md

7 directories, 12 files
```

Compared to the tree from the repo itself:

```
peterso@name ~/Projects/petems-shellshock_facts (master $=) $ tree .
.
├── Gemfile
├── Gemfile.lock
├── LICENSE
├── Modulefile
├── README
├── README.md
├── Rakefile
├── lib
│   └── facter
│       └── shellshock.rb
├── manifests
│   └── init.pp
└── spec
    ├── acceptance
    │   ├── class_spec.rb
    │   └── nodesets
    │       ├── centos-64-x64-docker.yml
    │       ├── centos-64-x64.yml
    │       ├── default.yml
    │       └── ubuntu-server-12042-x64.yml
    ├── spec_helper.rb
    ├── spec_helper_acceptance.rb
    └── unit
        └── facter
            └── shellshock_facts_spec.rb

8 directories, 17 files
```

chris-reeves:
Pull request which resolves this submitted: #608

anodelman:
PR merged.

PR #608, opened 2015-01-07
(gh-503) (QENG-1485) Fix recursive copying in do_scp_to

chris-reeves:
Because of the functionality within do_scp_to which allows certain files and directories to be ignored when copying, each specific file/directory is copied individually. In the existing code the source and target paths are effectively the same, causing directories to be copied as subdirectories of themselves. This PR ensures that all files and directories are copied to their parent directories.

This PR also includes some additional expectations for the spec tests for do_scp_to. Some tests which are supposed to test whether files and directories are excluded from copies were only testing that non-excluded files *were* copied, and didn't test that excluded files *weren't* copied. The relevant commit explicitly tests that excluded files are not copied. Although not directly related to the original issue, it seemed sensible to roll this commit in with the others.

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -358,9 +358,9 @@ module Beaker
         dir_source.each do |s|
           s_path = Pathname.new(s)
           if s_path.absolute?
-            file_path = File.join(target, s.gsub(source,''))
+            file_path = File.join(target, File.dirname(s).gsub(source,''))
           else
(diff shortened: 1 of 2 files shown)


branan:
OK to test

anodelman:
ok to test

anodelman:
Thanks for the patch!  Glad to see this fixed!

PR #610, opened 2015-01-08
Add stderr output to stacktrace

electrical:
In some cases an underlying issue causes the vagrant hypervisor to fail but gives no usable output.
By adding the stderr output to the raise gives the user more info on why it failed.


anodelman:
ok to test

anodelman:
retest this please

PR #614, opened 2015-01-12
(QENG-1662) beaker smoketests failing on ec2 with security group error

anodelman:
- ensure when looking for an appropriate security group that you are
  looking in a) the provided vpc or b) in the default vpc - do not allow
  a mix of vpcs.


doug-rosser:
+1

PR #624, opened 2015-01-14
(QENG-1410) implemented provisioning log

kevpl:


anodelman:
Yard doc warnings/errors.

lib/beaker/cli.rb
```diff
@@ -46,6 +46,7 @@ def initialize
 
       @options[:log_dated_dir] = Beaker::Logger.generate_dated_log_folder(@options[:log_dir], @timestamp)
       @options[:xml_dated_dir] = Beaker::Logger.generate_dated_log_folder(@options[:xml_dir], @timestamp)
+      @options[:logger_provisioning] = Logger.new(File.join(@options[:log_dated_dir], @options[:log_provisioning]), { :quiet => true })
```
colinPL: Why is this `Logger` call not prefixed with `Beaker::Logger` like the previous 2 lines?

lib/beaker/network_manager.rb
```diff
@@ -91,10 +94,69 @@ def cleanup
...
+    # @return [String] the log line created for this event
+    def log_provisioning host, create
+      provisioning_logger = @options[:logger_provisioning]
+      time = Time.new
+      stamp = time.strftime('%Y-%m-%d %H:%M:%S')
+      verb = create ? '+' : '-'
+      line = "#{stamp}\t[#{verb}]\t#{host['hypervisor']}\t#{host['platform']}\t#{host}"
+      @options[:log_provisioning_facts] = '' unless @options.has_key?(:log_provisioning_facts)
+      @options[:log_provisioning_facts] << "#{line}\n"
+      provisioning_logger.notify line
```
colinPL: There is no validation around `provisioning_logger` not being `nil`.

kevpl: added that in

kevpl:
failures look unrelated to the change to me

lib/beaker/logger.rb
```diff
@@ -312,6 +312,13 @@ def Logger.generate_dated_log_folder(base_dir, timestamp)
       log_dir
     end
 
+    #Remove color codes from provided string.  Color codes are of the format /(\e\[\d\d;\d\dm)+/.
+    #@param [String] text The string to remove color codes from
+    #@return [String] The text without color codes
+    def Logger.strip_color_codes(text)
```
anodelman: Shouldn't this just be called 'strip_color_codes' instead of 'Logger.strip_color_codes'?  Looks like a copy/paste error.

lib/beaker/network_manager.rb
```diff
@@ -91,10 +94,70 @@ def cleanup
...
+      line
+    end
+
+    def log_provisioning_summary_xml
```
anodelman: Missing yard doc.

anodelman:
Could you provide me a sample of the junit xml output after running this?

lib/beaker/logger.rb
kevpl: I thought this would be better as a class method, since it looks like a pretty generic utility method.  If you don't think that's useful, I can pull this out to make it simpler.

kevpl:
Here is an example of the test suite junit xml created for provisioning:

    <testsuite name="provisioning.log"><properties><property name="perserve_hosts" value="never"/></properties><testcase classname="log/2015-01-22_10_56_24" name="provisioning.log"><system-out><![CDATA[2015-01-22 10:56:24 [+] vcloud  el-6-x86_64 n2qtfyksv4qu24h.delivery.pup
    2015-01-22 10:56:24 [+] vcloud  el-6-x86_64 si10i2mcihpxhsf.delivery.puppetlabs.net
    2015-01-22 10:56:24 [+] vcloud  debian-7-x86_64 lfdvxzmdpe19ly1.delivery.puppetlabs.net
    2015-01-22 10:56:27 [-] vcloud  el-6-x86_64 n2qtfyksv4qu24h.delivery.puppetlabs.net
    2015-01-22 10:56:27 [-] vcloud  el-6-x86_64 si10i2mcihpxhsf.delivery.puppetlabs.net
    2015-01-22 10:56:27 [-] vcloud  debian-7-x86_64 lfdvxzmdpe19ly1.delivery.puppetlabs.net
    ]]></system-out></testcase></testsuite>

anodelman:
Breaks beaker-rspec.

/lib/beaker/network_manager.rb:113:in `log_provisioning': log_provisioning called before provisioning logger created. skipping debian-6-64-1, true (ArgumentError)

kevpl:
failures unrelated to change

anodelman:
I'd like to run this locally before merge.

anodelman:
Junit output doesn't look quite right.  The provisioning log appears after the test run data and there are a bunch of undefined elements ('Elapsed Time', 'Attempted', etc).  Please clean this up.

screen shot 2015-01-27 at 10 39 45 am

kevpl:
Since these numbers don't apply in this case, because it's not a test case, I've zeroed them out, so that they show correctly.  A screenshot is below:
screen shot 2015-01-27 at 1 53 47 pm
Let me know if there are other changes that you want.

colinPL:
Why is the provisioning log being added to the junit output file?

anodelman:
The junit output is just another way of viewing all of the beaker output, including the provisioning log puts more data in one place which just makes things easier to understand.

kevpl:
failures due to unrelated issue.

colinPL:
I thought the junit output was intended to be consumed by xUnit parsers, such as Jenkin's JUnit Publisher. This would add a fake or false `testcase` to the results generated by those parsers.

anodelman:
The contents of the provisioning log seems really... incomplete?  I was hoping to also see the log scroll of the machines being provisioned and any messages generated from that.

anodelman:
@colinPL I would consider the provisioning step to be a pre pre-suite, since it's pass/fail state causes the other tests to be run or not.

kevpl:
@anodelman, the log is mostly for people to decide whether or not to reuse machines in later Beaker runs, and have them easily consumable, so that they could grep/awk out a hostname for re-use/ssh-ing into, as agreed to in the ticket.

The provisioning log can't happen until the of all tests, which includes their junit output, otherwise it won't include cleanup events.  I could do this, but it would require more code, and some sort of "orchestrator", that decides from all of the TestSuite objects what order they go in.  The LoggerJunit could fulfill this role if required, just storing up items, and then when a `close` type operation is given to it, write everything there.  But that would be error-prone, as intermittent failures that stop execution before this step won't be logged, leaving us with holes in reporting for those issues.

anodelman:
Then it shouldn't be called the provisioning log - since it doesn't log the provisioning. :)

How about sut.log or host_info or something something?

kevpl:
It does log provisioning events, as well as de-provisioning events, which should be included in any provisioning log.

What makes this not a provisioning log?  Does it need more information?  Is it the time of logging?

anodelman:
A provisioning log would contain the log generated during the provisioning stage of beaker execution, so any messages generated by vagrant or vsphere or fusion or whatever.

This only contains host name information and none of that.

colinPL:
Since this is no longer associated with apart of the provisioning process, can it be removed from the junit output file?

anodelman:
@colinPL true, it doesn't actually cover provision logging or test logging - might just be a nice file to have around called host_info.txt.

colinPL:
@anodelman Agreed on `host_info.txt` or `sut.log`. My concern is strictly in regards to the `junit.xml` document and whether the contents are representative of a test or pre-suite pre-test..

kevpl:
changed the log file to `sut.log`.  Also, pulled logic out of `junit.xml`.  Let me know if you still have any other issues with this.

anodelman:
I'll test locally and then we'll be good to go.

anodelman:
Looks good.

PR #627, opened 2015-01-15
(QENG-1677) Adds Fog Hypervisor

petems:
This adds a new generic fog hypervisor, which you can then specify to the cloud provider you pick. Right now it does some regex to check the cloud provider name to check the auth method and any special fields you need to add (not every provider has regions, not every provider has flavors etc.)

This means that instead of having to write out an entire new provider if someone asks for it for some of the other popular cloud services out there (Brightbox are pretty big in the UK, Linode are starting to get popular etc.) we can just extend how the fog provider works and use that.

It also means if someone prefers the fog way of doing things rather than the beaker implementation for some reason, they can use fog to run GCE, AWS, Openstack etc.

### DIFF
--- a/lib/beaker/hypervisor.rb
+++ b/lib/beaker/hypervisor.rb
@@ -129,6 +129,6 @@ module Beaker
   end
 end
 
-[ 'vsphere_helper', 'vagrant', 'vagrant_virtualbox', 'vagrant_parallels', 'vagrant_libvirt', 'vagrant_fusion', 'vagrant_workstation', 'fusion', 'aws_sdk', 'vsphere', 'vcloud', 'vcloud_pooled', 'aixer', 'solaris', 'docker', 'google_compute', 'openstack' ].each do |lib|
+[ 'vsphere_helper', 'vagrant', 'vagrant_virtualbox', 'vagrant_parallels', 'vagrant_libvirt', 'vagrant_fusion', 'vagrant_workstation', 'fusion', 'aws_sdk', 'vsphere', 'vcloud', 'vcloud_pooled', 'aixer', 'solaris', 'docker', 'google_compute', 'openstack', 'fog' ].each do |lib|
     require "beaker/hypervisor/#{lib}"
 end
--- a/lib/beaker/hypervisor/fog.rb
+++ b/lib/beaker/hypervisor/fog.rb
@@ -1,0 +1,105 @@
+require 'fog'
+
+module Beaker
+  #Beaker support for Fog
+  class Fog < Beaker::Hypervisor
+
+    SLEEPWAIT = 5
+
+    #Create a new instance of the Fog hypervisor object
+    #@param [<Host>] hosts The array of Fog hosts to provision
+    #@param [Hash{Symbol=>String}] options The options hash containing configuration values
+    def initialize(hosts, options)
+      @options = options
+      @logger = options[:logger]
+      @hosts = hosts
+      @vms = []
+
+      host = hosts.first
+
+      fog_auth = get_cloud_auth_method(hosts)
+
+      fog_compute_options = { :provider => host[:cloud_provider]}.merge( fog_auth )
+
+      @compute_client ||= ::Fog::Compute.new( fog_compute_options )
+
+      if not @compute_client
+        raise "Unable to create #{host[:cloud_provider]} Compute instance"
+      end
+    end
+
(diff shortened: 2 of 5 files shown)


lib/beaker/hypervisor.rb
```diff
@@ -127,6 +127,6 @@ def generate_host_name
   end
 end
 
-[ 'vsphere_helper', 'vagrant', 'vagrant_virtualbox', 'vagrant_libvirt', 'vagrant_fusion', 'vagrant_workstation', 'fusion', 'aws_sdk', 'vsphere', 'vcloud', 'vcloud_pooled', 'aixer', 'solaris', 'docker', 'google_compute', 'openstack' ].each do |lib|
+[ 'vsphere_helper', 'vagrant', 'vagrant_virtualbox', 'vagrant_libvirt', 'vagrant_fusion', 'vagrant_workstation', 'fusion', 'aws_sdk', 'vsphere', 'vcloud', 'vcloud_pooled', 'aixer', 'solaris', 'docker', 'google_compute', 'openstack', 'digital_ocean', 'fog' ].each do |lib|
```
anodelman: Unrelated 'digital_ocean' file inclusion here.

lib/beaker/hypervisor/fog.rb
```diff
@@ -0,0 +1,104 @@
...
+        @logger.notify "Provisioning #{host[:cloud_provider]} using Fog"
+        host[:vmhostname] = generate_host_name
+        @logger.debug "Provisioning #{host.name} (#{host[:vmhostname]})"
+
+        extras = provider_specific_fields(host)
+
+        default_server_options = {
+          :name => host[:vmhostname],
+          :image_id => host[:image_id],
+          :public_key_path => host[:public_key_path] || File.expand_path('~/.ssh/id_rsa.pub'),
```
anodelman: Default key values should be stored in options/presets.rb for easier tracking across the project.

lib/beaker/hypervisor/fog.rb
```diff
@@ -0,0 +1,104 @@
...
+    #Destroy any digitalocean instances
+    def cleanup
+      @logger.notify "Cleaning up Fog Created Servers"
+      @vms.each do |vm|
+        @logger.debug "Destroying host: #{vm.name}"
+        vm.destroy
+      end
+    end
+
+    def get_cloud_auth_method(hosts)
```
anodelman: Missing yard docs.

lib/beaker/hypervisor/fog.rb
```diff
@@ -0,0 +1,104 @@
...
+        fog_auth = {
+          :digitalocean_api_key => host[:digitalocean_api_key],
+          :digitalocean_client_id => host[:digitalocean_client_id],
+        }
+      else
+        raise NotImplementedError, "Don't know the cloud auth for #{provider}"
+      end
+    end
+
+    def provider_specific_fields(host)
```
anodelman: Missing yard docs.

lib/beaker/hypervisor.rb
petems: Fixed

anodelman:
Along with previous comments there are references to 'digitalocean' in here that don't belong.

petems:
@anodelman Removed 'digitalocean' references :+1:

lib/beaker/hypervisor/fog.rb
petems: Fixed

lib/beaker/hypervisor/fog.rb
petems: Fixed

lib/beaker/hypervisor/fog.rb
petems: Fixed

anodelman:
You've got a spurious commit mixed in there.

petems:
@anodelman fixed, rebased with master too.

petems:
So, I'm going to pull this out into it's own repo:
https://github.com/petems/beaker-fog

I've just tested it, and having it in the gemspec for a module works like a charm! :smile: 

I'll write up some stuff on the Wiki next week about how to make your own Hypervisor plugins as separate gems and such.

Me, @justinstoller and @liamjbennett talked a bit on how to do this for hosts and such in the future too, it seems harder because you have more classes to reference, but still possible :dancer:

PR #633, opened 2015-01-16
WIP: (PDB-1034) Ezbake changes to get PDB source based builds working

kbarber:
This involves a rework of the ezbake_utils to provide more capability to fully
install PuppetDB using these helpers.

* Some of the original API has been simplified, and the YAGNI parts removed
  since no one was using them today.
* We now do the lein install process, but using a local maven repository to
  avoid collision with other projects running on the same host.
* install_termini_from_ezbake now added to install that component
* I've switched to using the install.sh methodology on ezbake.
* This now is compatible with lein-ezbake.
* ezbake_local_cmd has been added to generalize this kind of system invocation
  and simplify testing.
* ezbake_installsh generalizes the way we invoke install.sh, and simplifies testing.
* install_ezbake_tarball_on_host has been created to generalize this step for
  the termini and service based installations to re-use. Its idempotent, in
  that it checks to ensure its not already installed.
* conditionally_clone was modified to support being passed a branch if required,
  so in the future we can simplify working on a development branch.
* Lots of yarddoc cleanups
* Some rspec coverage

### DIFF
--- a/lib/beaker/dsl/ezbake_utils.rb
+++ b/lib/beaker/dsl/ezbake_utils.rb
@@ -1,80 +1,158 @@
 require 'beaker/dsl/install_utils'
+require 'fileutils'
 
 module Beaker
   module DSL
-    #
     # This module contains methods to assist in installing projects from source
     # that use ezbake for packaging.
     #
     # @api dsl
     module EZBakeUtils
 
-      REMOTE_PACKAGES_REQUIRED = ['make']
+      # @!group Public DSL Methods
+
+      # Installs leiningen project with given name and version on remote host.
+      #
+      # @param [Host] host A single remote host on which to install the
+      #   specified leiningen project.
+      # @api dsl
+      def install_from_ezbake host
+        ezbake_validate_support host
+        ezbake_tools_available?
+        install_ezbake_tarball_on_host host
+        ezbake_installsh host, "service"
+      end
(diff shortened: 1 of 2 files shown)


kbarber:
@anodelman @waynr this needs to wait for the ezbake PR to get in, but feel free to review it now.

anodelman:
I'd like to get an ezbake smoketest of some sort set up before we do a merge on this - we don't have coverage in this area and shouldn't be merging more code without proof of functionality.

@waynr  @kbarber would you have time to assist in that?

waynr:
@anodelman @kbarber 
I just created QENG-1710 to track the need for this ezbake smoketest with some comments about my preferences about what it should look like.

kbarber:
So, I was going to put the last touches on the other side of this patch tomorrow, and suddenly found this comment. So, I won't have any time to do this, nor will we accept being blocked on this work for more than we need to. I think I'll just move this functionality into our own helper and not bother with the core change if we're suddenly going to make it difficult.

waynr:
@kbarber 
Are you willing to re-open this? @anodelman and I talked about this, we are willing to merge it without resolving QENG-1710 (actually QENG-1709, of which QENG-1710 was a duplicate) with the understanding that I will take on that ticket during my 2/25 sprint.

PR #636, opened 2015-01-20
(QENG-1684) Allow exit code of '1' when stopping puppet agent on centos4

branan:
When upgrading, the agent service is already stopped. On Centos4, this
causes a failure from the init script. We want to ignore that failure.


branan:
Failure appears unrelated

PR #638, opened 2015-01-21
(QENG-1690) added in helper installation method for packages

kevpl:
I had misunderstood the install_puppetlabs_dev_repo method before, thinking that it would install packages
from within the dev repo, but it's actually just installing the dev repo itself.  These changes add a
corresponding installation method to install the packages found in these repos after they've been installed

### DIFF
--- a/lib/beaker/dsl/install_utils.rb
+++ b/lib/beaker/dsl/install_utils.rb
@@ -1125,6 +1125,30 @@ module Beaker
         end
       end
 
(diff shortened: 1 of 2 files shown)


anodelman:
This passed local testing?

kevpl:
Yes, it did.  I created a simple test that, after these methods were called to install the puppet-agent packages, just did a simple puppet configprint call to verify that it was reachable on the system.

PR #640, opened 2015-01-22
(QENG-1695) Improves OSX Support

petems:
* Fixes for root key location copy
* Adds a Vagrant VB OSX option
* Fixes PermitRootLogin regex
* Adds option to enable GUI
* Raises error if version not given for DMG

Now works with an example OSX Virtualbox image :+1:


anodelman:
ok to test

petems:
Sorry I accidentally picked up a different commit whilst rebasing, just fixed.

Issue #642, opened 2015-01-23
Openstack hypervisor should lookup credential from environment variables


mcanevet:

anodelman:
Merged.

PR #643, opened 2015-01-23
(gh-642) Lookup Openstack credentials from ENV

mcanevet:


lib/beaker/hypervisor/openstack.rb
```diff
@@ -27,6 +27,11 @@ def initialize(openstack_hosts, options)
       @hosts = openstack_hosts
       @vms = []
 
+      @options[:openstack_api_key] ||= ENV['OS_PASSWORD']
+      @options[:openstack_username] ||= ENV['OS_USERNAME']
+      @options[:openstack_auth_url] ||= ENV['OS_AUTH_URL']
+      @options[:openstack_tenant] ||= ENV['OS_TENANT_NAME']
+
```
anodelman: We do all environment parsing in options/parser.rb (and options/presets.rb) so that we can easily track what env vars are supported by Beaker - this code should be shifted over there.

mcanevet:
@anodelman fixed

PR #645, opened 2015-01-26
(QENG-1714) fixed get_type for a machine with only non-agent roles

kevpl:
The issue here is that when you either have a machine with the roles master and agent, or agent, the AIO paths were set correctly when the Beaker run was given the `:type` parameter with the value 'aio'.

The issue is that if you have a machine that doesn't have the agent role at all, it is still getting the 'aio' paths.  In the simple case of having two separate master and agent machines, the master machine currently has an incorrect path.  This fix queries the roles, and if the system doesn't have the 'agent' role, it will come back as FOSS, which is the default for this setup.

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -49,6 +49,7 @@ module Beaker
       # related through 'type' and the differences between the assumption of our two
       # configurations we have for many of our products
       type = @options.get_type
+      type = :foss if type == :aio && !@options['HOSTS'][@name]['roles'].include?('agent')
       @defaults = merge_defaults_for_type @options, type
       pkg_initialize
     end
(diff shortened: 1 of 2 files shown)


kevpl:
failures unrelated to changes

lib/beaker/options/options_hash.rb
```diff
@@ -60,14 +62,18 @@ def is_pe?
...
-      def get_type
+      def get_type name
         case self[:type]
         when /pe/
           :pe
         when /foss/
           :foss
         when /aio/
-          :aio
+          if self['HOSTS'][name]['roles'].include? 'agent'
```
anodelman: This doesn't belong in OptionsHash - which is just an extension of hash and should not have included information as to the structure of what it is storing.  In some cases it will not have a 'HOSTS' section and then this will die.

anodelman:
This is probably a temporary patch until we resolve the pathing differences between pe/foss.

But, good for now. :)

Issue #393, opened 2014-08-07
Support snapshotting SUT


jantman:
It would be *really* nice to be able to snapshot SUTs. There's a Vagrant plugin for VirtualBox that does it quite nicely, though I haven't tried the cross-provider ones.

I *think* that the ideal implementation for me would be to just let the user handle all of it, i.e. simply expose the functionality via the DSL and let me, in my tests, ``take_snapshot(:step1)`` and ``restore_snapshot(:step1)``.

I know it's a pretty big undertaking and I hate submitting issues that I have little to no intent of trying to fix, but this came up when I was trying to test https://github.com/puppetlabs/puppetlabs-rabbitmq/pull/221 (where the package simply can't be cleanly upgraded/downgraded).

alexjfisher:
My use case is testing roles.  I followed http://tech.opentable.co.uk/blog/2014/09/01/testing-puppet-with-beaker-pt-dot-3-testing-roles/ but then wanted a way to reset my test environment between tests.

anodelman:
https://github.com/puppetlabs/beaker/pull/654 closed due to inactivity.

jantman:
So this is just off the table?

PR #654, opened 2015-01-28
SUT snapshot support for vagrant hypervisor

alexjfisher:
This is my attempt at addressing https://github.com/puppetlabs/beaker/issues/393

Perhaps it's OK to merge, but far more likely it needs more work first.

### DIFF
--- a/lib/beaker/dsl/helpers.rb
+++ b/lib/beaker/dsl/helpers.rb
@@ -1340,6 +1340,38 @@ module Beaker
         fact_on(default, name, opts)
       end
 
+      # Snapshot provided host
+      #
+      # Only vagrant hosts are supported
+      # and you must have https://github.com/scalefactory/vagrant-multiprovider-snap installed
+      #
+      # @param host [Host] The host to snapshot
+      # @param name [String] The name to use as the snapshot name
+      def take_snapshot_on(host, name)
+        host.do_take_snapshot(name)
+      end
+
+      # Snapshot default host
+      # @see #take_snapshot_on
+      def take_snapshot(name)
+        take_snapshot_on(default,name)
+      end
+
+      # Revert to a snapshot on provided host
+      #
+      # @param host [Host] The host being reverted
+      # @param name [String] The name of the snapshot to revert to
+      def restore_snapshot_on(host, name)
+        host.do_restore_snapshot(name)
+      end
+
+      # Revert to a snapshot on the default host
+      # @see #restore_snapshot_on
+      def restore_snapshot(name)
+        restore_snapshot_on(default,name)
+      end
+
+
       #Run a curl command on the provided host(s)
       #
       # @param [Host, Array<Host>, String, Symbol] host    One or more hosts to act upon,
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -379,6 +379,13 @@ module Beaker
       return result
     end
 
+    def do_take_snapshot snapshot_name
+      self[:hypervisor].take_snapshot(name, snapshot_name)
+    end
+
+    def do_restore_snapshot snapshot_name
+      self[:hypervisor].restore_snapshot(name, snapshot_name)
+    end
   end
 
   [ 'windows', 'unix', 'aix', 'mac' ].each do |lib|
--- a/lib/beaker/hypervisor.rb
+++ b/lib/beaker/hypervisor.rb
@@ -92,6 +92,15 @@ module Beaker
       nil
     end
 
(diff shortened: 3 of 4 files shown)


lib/beaker/hypervisor/vagrant.rb
```diff
@@ -160,6 +160,17 @@ def cleanup
       FileUtils.rm_rf(@vagrant_path)
     end
 
+    #snapshotting depends on https://github.com/scalefactory/vagrant-multiprovider-snap
+    def take_snapshot(host,snapshot_name)
+      @logger.debug "Creating snapshot of #{host}"
```
electrical: My best guess is that you want to use host.name in these parts to get the name of the host ?

alexjfisher: I'm calling take_snapshot from host.rb with the name.  I'm not passing the Host.
https://github.com/puppetlabs/beaker/pull/654/files#diff-25c9ddfb41958c160da1d3b679793ddaR382

Would it be better style to do "self[:hypervisor].take_snapshot(self, snapshot_name)" and call host.name in vagrant.rb as you've suggested?

alexjfisher: At the least, I should probably rename the parameter to hostname to avoid confusion.

anodelman:
Please mark your commits with (gh-xxx) indicating the issue that they are fixing.

alexjfisher:
@anodelman I've rewritten my commit messages with gh-393 (also rebased).

lib/beaker/host.rb
```diff
@@ -379,6 +379,13 @@ def do_scp_from source, target, options
       return result
     end
 
+    def do_take_snapshot snapshot_name
```
anodelman: I don't think that this belongs in the host code.  A host is pretty much unaware of what hypervisor is running it and I don't want them to be so coupled to their hypervisor.

lib/beaker/dsl/helpers.rb
```diff
@@ -1340,6 +1340,38 @@ def fact(name, opts = {})
...
 
+      # Snapshot provided host
+      #
+      # Only vagrant hosts are supported
+      # and you must have https://github.com/scalefactory/vagrant-multiprovider-snap installed
+      #
+      # @param host [Host] The host to snapshot
+      # @param name [String] The name to use as the snapshot name
+      def take_snapshot_on(host, name)
+        host.do_take_snapshot(name)
```
anodelman: Instead of sending this to the host, let's send to the hypervisor.  I think that we should split out a separate chunk of the DSL to be suts.rb or boxes.rb or nodes.rb that refers to actions that can be applied to hosts through their respective hypervisors.  Then we'll also need to make the hypervisor hash controlled by beaker available to those methods.  Then the call through system will work.

lib/beaker/hypervisor/vagrant.rb
```diff
@@ -161,6 +161,17 @@ def cleanup
       FileUtils.rm_rf(@vagrant_path)
     end
 
+    #snapshotting depends on https://github.com/scalefactory/vagrant-multiprovider-snap
```
anodelman: I know that the rest of this file isn't properly yard doc-ed, but please include yard docs for any new methods.

lib/beaker/hypervisor.rb
```diff
@@ -85,6 +92,15 @@ def cleanup
       nil
     end
 
+    #Take a snapshot of a specific host running on this hypervisor.
```
anodelman: Include yard docs, please.

lib/beaker/dsl/helpers.rb
alexjfisher: @anodelman Thanks for your feedback.  I was pretty sure I was going about things the wrong way, but I couldn't figure out how else I could get access to the hypervisor object from a host.

> we'll also need to make the hypervisor hash controlled by beaker available to those methods

How might you suggest this is done?  The hash is an instance variable of the NeworkManager object.  Will I need to patch rspec-beaker and add the hypervisor hash as a 'custom setting' there? (or have I completely misunderstood)?

anodelman: Yes, beaker-rspec will also need updating to support this new global variable.  I would expose it in the same way we have the 'hosts' and 'options' - see test_case.rb.

anodelman:
Needs a rebase, also low activity here.

Will close next week due to inactivity if there have been no updates.

anodelman:
Closed due to inactivity.

PR #659, opened 2015-01-29
(QENG-1758) beaker bash completion script should not prevent filename co...

er0ck:
...mpletion

This change specifies -o default to use the default completion method
when a completion is not found with this script. This allows completion
of filenames as well as beaker's options.


erikPrime:
Only works for bash? :(

er0ck:
here you go @erikPrime : https://askql.wordpress.com/2011/01/11/zsh-writing-own-completion/

Issue #660, opened 2015-01-29
Default AIO puppetpath should be /etc/puppetlabs/agent


johnduarte:
As of 5b050f6ccff, the default AIO puppetpath is set to /opt/puppetlabs/agent for unix. This value should be /etc/puppetlabs/agent

anodelman:
Please file this in Jira.

anodelman:
Ah, you already have a patch.  Leave this as-is then.  :)

anodelman:
Merged.

PR #661, opened 2015-01-30
(gh-660) Update default aio puppetpath path for unix

johnduarte:
This commit updates the default aio puppetpath to /etc/puppetlabs/agent


anodelman:
ok to test

kevpl:
@johnduarte and the other paths that still have /opt are ok?

johnduarte:
I have reviewed this with @rlinehan . Their is some uncertainty about distmoduledir, but currenly puppetpath is the only one that is definitely incorrect.

johnduarte:
@anodelman sorry, I already filed https://tickets.puppetlabs.com/browse/QENG-1759

PR #662, opened 2015-01-30
Upgrade google-api-client dependency to fix Ruby 1.9.x compatibility

kbarber:
Retriable was upgraded today (Jan 30th) and they shipped a version 2.0.0 that
requires Ruby >=2.0.0.

In google-api-client 0.7.1, we were pulling in version >= 1.4 of retriable. In 0.8.2
they have a fix that pulls in ~> 1.4 version of retriable, thus pinning the ceiling
on the upgrade to retriable 2.0.0.

You can see the failure here: https://jenkins.puppetlabs.com/job/platform_puppetdb_intn-sys_pr/506/BEAKER_CONFIG=ec2-west-el7-64mda-el7-64a,BEAKER_OPTIONS=postgres,label=beaker-ec2/console

### DIFF
--- a/beaker.gemspec
+++ b/beaker.gemspec
@@ -42,7 +42,7 @@ Gem::Specification.new do |s|
   # Optional provisioner specific support
   s.add_runtime_dependency 'rbvmomi', '~> 1.8'
   s.add_runtime_dependency 'fission', '~> 0.4'
-  s.add_runtime_dependency 'google-api-client', '~> 0.7'
+  s.add_runtime_dependency 'google-api-client', '~> 0.8'
   s.add_runtime_dependency 'aws-sdk', '~> 1.57'
   s.add_runtime_dependency 'docker-api'
   s.add_runtime_dependency 'fog', '~> 1.25'


kbarber:
Might possibly be better solved with a pin against retriable, but I thought upgrading google-api-client would be a better longer term solution.

rji:
retest this please

anodelman:
error in puppetdb smoketests:

    Fetching gem metadata from https://rubygems.org/......
    Resolving dependencies...
    Bundler could not find compatible versions for gem "activesupport":
      In Gemfile:
        beaker (>= 0) ruby depends on
          activesupport (>= 3.2) ruby
         activerecord (= 2.3.14) ruby depends on
         activesupport (2.3.14)

kbarber:
@anodelman we don't maintain these smoketests, and they differ dramatically from ours, I've told you this before ... if you differ from us you have to maintain these yourselves and sort out the errors. I'm getting tired of managing the differences in your testing, while you just ignore my advice to make sure they match exactly.

But anyway ... just do a --without test when you install the gems, like our tests do. That skips the stuff required by rspec. If you can't take responsibility for tests that differ from ours, because you choose to make them differ, I can't keep helping you with the differences:

https://github.com/puppetlabs/puppetdb/blob/master/ext/jenkins/beaker-tests-source.sh

Sort out the differences, and stop making it my problem please.

kbarber:
@anodelman I am deeply sorry about my last comment. It was uncalled for, and cruel and I hope you can forgive me. @djm68 told me off for it, and he was right. I don't mean to be callous, I just lose my cool sometimes, and I am sincerely sorry for treating a fellow colleague like yourself like this. We're all busy and working our hardest, and no one deserves to be snapped at like that. I won't say I won't do it again, because I know my stupid nature, but I hope you understand now I've thought about it, I realise it was wrong and uncalled for and I deeply regret it and will work hard to not be so harsh going forward. Please, your important to us all, and I'm sorry for treating you less than that.

Anyway ... as far as the problem. I'm not quite sure why this wasn't noticed before, but my best advice would be that you should avoid the 'test' group in our Gemfile. This contains rspec specific stuff, and in this case ActiveRecord which is required by our rspec tests, since we still test against the old ActiveRecord storeconfigs code in Puppet. This is deprecated, in fact its the 'old puppetdb' way, but we have a migration tool we still support that needs it. It will go away in Puppet 4.0.0, since legacy storeconfigs (ie. ActiveRecord) will finally be retired. Kind of a marking point in Puppet history also, this means PuppetDB is the only way left to do it.

We hit this earlier ourselves by the way, you can see our workaround here:

https://github.com/puppetlabs/puppetdb/blob/master/ext/jenkins/beaker-tests-source.sh#L33

By doing this, it avoids the 'test' group, and the problem with ActiveRecord 'disappears'. Which is good enough for our purposes (fwiw, I also do the reverse for running unit tests, that is I skip the acceptance group because beaker pulls in a lot of stuff). This way I can re-use 1 Gemfile for both test types basically, without pulling in croft I don't really need.

So I'm not sure what you're doing on the internal jenkins server (I don't have admin, so can't see the script) but consider switching to calling this script:

https://github.com/puppetlabs/puppetdb/blob/master/ext/jenkins/beaker-tests-source.sh

I recall you doing something special with images and using vsphere versus ec2, I don't know how to handle that in a standard way, but I wouldn't mind 1 place to manage this kind of thing. Although there is a lot to be said about using EC2 for smoke testing also, since that actually reflects our reality today, and I think we could pick up lots of bugs in the EC2 provider this way. Of course @waynr has a lot of ideas about test reproducibility as well, but is less of a fan of the 'checked in shell script', but he's doing the research for us in this arena, so I'd find out how we can make sure these kind of differences don't create differences between our jobs and this smoke test. That is, I trust he'll do something cool & creative that works.

Hope this helps :-).

anodelman:
I'll look over the smoketest based on your comments and see what I can get going.

anodelman:
We've been running with

bundle install --path vendor/bundle --without=test

Since early in January when I last looked these tests over.  I'll dig around some more and see what I can find.

anodelman:
Okay, now I'm confused.  Did you get this working with the --without?  The bundle man page tells me:

{noformat}
While the --without option will skip installing the gems in the specified groups, it will still download those gems and use them to resolve the dependencies of every gem in your Gemfile(5).
{noformat}

I can create a successful install if I comment out the 'test' group, so that is where the dependency conflict is being generated - but I don't see how --without will repair that since bundler still does dependency resolution using gems that are within an excluded group.

kbarber:
@anodelman well then I have little idea, besides jumping onto the box that is showing the problem and poking around. I certainly can't replicate it, our tests run many times a day without issue, so not sure.

anodelman:
Nertz.  I can reproduce on my laptop env, but that doesn't explain why you are getting green runs.  I'll poke around more and compare the two different jobs and see what I can find.

anodelman:
I still haven't found the difference, but the failure is from:

    gem 'activerecord', '2.3.14'

In the puppetdb Gemfile.  Do you need to pin activerecord here?  Could you update the pin?  If you either dropped the pin or updated it I think that we could get this worked out.

kbarber:
Yes, we do need the pin for Puppet related testing.

kbarber:
Okay, I know what is throwing us ... its google-api-client 0.8 that requires the latest activesupport:

```
Bundler could not find compatible versions for gem "activesupport":
  In Gemfile:
    beaker (>= 0) ruby depends on
      google-api-client (~> 0.8) ruby depends on
        activesupport (>= 3.2) ruby

    activerecord (= 2.3.14) ruby depends on
      activesupport (2.3.14)
```

For some reason that original failure didn't show google-api-client in the list so I thought this was a general problem with your tests, now I understand its just a conflict between the two.

kbarber:
@anodelman so going back to your original idea the pinning will be fine, but only for activerecord 3.2. Anything newer than that, Puppet will fail. Our PR is here: https://github.com/puppetlabs/puppetdb/pull/1250. We'll see if it passes CI.

anodelman:
Thanks for your work here @kbarber!  Nobody likes being stuck in dependency hell.

anodelman:
retest this please

anodelman:
Looks like the puppetdb fix has been applied to stable, but not master.  Currently, beaker runs against puppetdb master.  I'll see what I can do.

anodelman:
This is pending repair of the puppetdb smoketest.  Working with @waynr to diagnose.

anodelman:
Puppetdb now green.

anodelman:
retest this please

anodelman:
Final nit - please mark your commit with (MAINT) or associate with a bug number.

anodelman:
Moved to https://github.com/puppetlabs/beaker/pull/726 to meet contributor guidelines.

PR #665, opened 2015-02-02
(QENG-1757) use a different command to set the hostname for AWS el-7 ins...

doug-rosser:
...tances. My only concern is if this problem affects other hypervisors, this is not a general-case solution.

Finally, I have done some testing and do not notice the issue with el-6.


anodelman:
We don't currently execute any tests on el-7, did this pass testing on your end?

doug-rosser:
Yes.

anodelman:
Green tests, ready to go.

PR #666, opened 2015-02-03
(MAINT) Fix mkdir bug on PSWindows and fix test

petems:
Introduced in f551c0ad124bb8ab782f7dd90ab30ae9cc49fafa

This actually breaks Beaker, because the host variable is nil:

```
/opt/rubies/2.1.2/lib/ruby/gems/2.1.0/bundler/gems/beaker-136d9c3cccbc/lib/beaker/host.rb:305:in `mkdir_p': undefined local variable or method `host' for #<FreeBSD::Host:0x007f8abe0e20b8> (NameError)
	from /opt/rubies/2.1.2/lib/ruby/gems/2.1.0/bundler/gems/beaker-136d9c3cccbc/lib/beaker/host.rb:364:in `block in do_scp_to'
	from /opt/rubies/2.1.2/lib/ruby/gems/2.1.0/bundler/gems/beaker-136d9c3cccbc/lib/beaker/host.rb:361:in `each'
	from /opt/rubies/2.1.2/lib/ruby/gems/2.1.0/bundler/gems/beaker-136d9c3cccbc/lib/beaker/host.rb:361:in `do_scp_to'
	from /opt/rubies/2.1.2/lib/ruby/gems/2.1.0/bundler/gems/beaker-136d9c3cccbc/lib/beaker/dsl/helpers.rb:207:in `block in scp_to'
```

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -344,7 +344,8 @@ module Beaker
       if self.is_cygwin?
         cmd = "mkdir -p #{dir}"
       else
-        cmd = "if not exist #{dir.gsub!('/','\\')} (md #{dir.gsub!('/','\\')})"
+        windows_dirstring = dir.gsub('/','\\')
+        cmd = "if not exist #{windows_dirstring} (md #{windows_dirstring})"
       end
 
       result = exec(Beaker::Command.new(cmd), :acceptable_exit_codes => [0, 1])
--- a/spec/beaker/host_spec.rb
+++ b/spec/beaker/host_spec.rb
@@ -355,7 +355,7 @@ module Beaker
         allow( result ).to receive( :exit_code ).and_return( 0 )
         allow( host ).to receive( :exec ).and_return( result )
 
-        expect( Beaker::Command ).to receive(:new).with("if not exist test\\test\\test (md )")
+        expect( Beaker::Command ).to receive(:new).with("if not exist test\\test\\test (md test\\test\\test)")
         expect( host.mkdir_p('test/test/test') ).to be == true
 
       end


petems:
Any thoughts @liamjbennett? Maybe switching to explicit declaration of bitvise connection instead?

liamjbennett:
Good catch. :+1:

lib/beaker/host/pswindows.rb
```diff
@@ -28,46 +28,46 @@ def self.pe_defaults
...
-                  })
-                else
-                  h.merge({
-                    'puppetpath'    => "C:\\Program Files\\Puppet Labs\\Puppet\\etc",
-                    'hieraconf'     => "C:\\Program Files\\Puppet Labs\\Puppet\\etc\\hiera.yaml",
-                    'puppetvardir'  => 'C:\\Program Files\\Puppet Labs\\Puppet\\var',
-                    'puppetbindir'  => "C:\\Program Files\\Puppet Labs\\Puppet\\bin"
-                    })
-                  end
+      if h['platform'] && h['platform'].include?('amd64')
```
anodelman: I hit this myself, but I don't think that this is the right fix.  Here you are checking h for a platform setting, but you just created h - it will have no platform setting and you will never hit this chunk of code.

petems: This is the code already in trunk, I was just re-indenting it. I can take it out of the PR if needed?

anodelman: Then I believe that I have a patch that will stomp this one. :)

madAndroid: +1 I'm experiencing this as well, and have reverted to using `2.2.0` for now

lib/beaker/host.rb
```diff
@@ -300,7 +300,7 @@ def exec command, options={}
     # @param [String] dir The directory structure to create on the host
     # @return [Boolean] True, if directory construction succeeded, otherwise False
     def mkdir_p dir
-      if host['is_cygwin'].nil? or host['is_cygwin'] == true
+      if @options['is_cygwin'].nil? or @options['is_cygwin'] == true
```
petems: @anodelman this is the main culprit of the error.

lib/beaker/host/pswindows.rb
petems: Just pulled the re-indentation stuff into a new PR to keep this clean

petems: Moved into https://github.com/puppetlabs/beaker/pull/670

lib/beaker/host.rb
anodelman: Let's switch this to self['is_cygwin'] to make use of the hosts options accessor.

petems: You mean like this:

```ruby
def mkdir_p dir
      if self['is_cygwin']
        cmd = "mkdir -p #{dir}"
      else
        cmd = "if not exist #{dir.gsub!('/','\\')} (md #{dir.gsub!('/','\\')})"
      end

      result = exec(Beaker::Command.new(cmd), :acceptable_exit_codes => [0, 1])
      result.exit_code == 0
    end
```

anodelman:
Needs a rebase.

lib/beaker/host.rb
anodelman: Sorry, that should be self.is_cygwin?.

anodelman: Looking over this again, we should use the host object inheritance to avoid doing an is_cygwin at all.  The 'mkdir -p' should be used by unix type hosts and the 'if not exists' should be pushed into the pswindows host.

anodelman:
More review comments, also needs rebase.

petems:
@anodelman Just rebased.

You'd already fixed the `no local variable bit` in 9c32cac7a7a5bf52f21b628b14ddba9bfc44510c, I thought it would work on PSWindows, but I noticed the logic around creating a directory was wrong. This should fix it (tested on a Bitvise windows box)

lib/beaker/host.rb
```diff
@@ -344,7 +344,8 @@ def mkdir_p dir
       if self.is_cygwin?
         cmd = "mkdir -p #{dir}"
       else
-        cmd = "if not exist #{dir.gsub!('/','\\')} (md #{dir.gsub!('/','\\')})"
+        windows_dirstring = dir.gsub!('/','\\')
```
anodelman: If you are going to be storing the result use dir.gsub instead of dir.gsub!.

joshcooper: Also `gsub!` modifies the `dir` parameter passed into this method, and would be surprising to the caller.

petems: Fixed! :+1:

kevpl:
failure unrelated to change. :+1:

PR #671, opened 2015-02-05
Revert "(QENG-1784) updated FOSS group to be root instead of puppet"

kevpl:
Reverts puppetlabs/beaker#667

This is because the Puppet user was being confused for the Beaker user.  

This means that people were using `host['user']` and `host['group']` for puppet stuff incorrectly.


kevpl:

PR #672, opened 2015-02-06
(QENG-1795) added code & conf paths to non-unix hosts

kevpl:


kevpl:
failures unrelated to changes

PR #674, opened 2015-02-06
(maint) Add ruby to windows PATH

joshcooper:
Previously, beaker would add `puppetbindir` to the `~/.ssh/environment`,
which on Windows, excluded the ruby that the MSI installs. As a result,
puppet setup steps that rely on `gem`, e.g. when adding sources, would
fail.

This commit adds the ruby in the MSI to the `~/.ssh/environment` simply by
it being included in `puppetbindir`. Note the actual location depends on
whether we are installing x86 puppet on x64 windows, or not, so we
include both possibilities like we do for puppet's main bin directory.


joshcooper:
@kevpl can you take a look at this when you get a chance?

Iristyle:
:+1: to :ship:

kevpl:

anodelman:
Failure unrelated to patch.

PR #676, opened 2015-02-09
(QENG-1809) removed master path resets for AIO test setups

kevpl:
Originally, I had assumed that machines with the master role would keep their old paths, or be updated
with a completely new set.  It looks like the master pathing will actually be based upon the agent though,
since masters will be installed on top of agents now


anodelman:
Failures unrelated to patch.

PR #679, opened 2015-02-10
(QENG-1813) Use puppet-server service names

joshcooper:
Previously, when executing AIO acceptance tests, the beaker method
`with_puppet_running_on` would attempt to start and stop the
`puppetmaster` service. This is because the AIO defaults for
`puppetservice` and related conf.d were not specified, and would
fallback to the FOSS defaults.

This commit updates the AIO defaults to match the `puppetserver`
service, and related conf.d.


anodelman:
Failures unrelated to patch.

PR #683, opened 2015-02-13
(gh-687) [openstack] multi node root user enable

fiddyspence:
This PR enables root access on each node in turn, after it is provisioned.  Previously,
the code did all nodes on the first pass, before some were even a twinkle in Openstack's
eye.

### DIFF
--- a/lib/beaker/hypervisor/openstack.rb
+++ b/lib/beaker/hypervisor/openstack.rb
@@ -133,7 +133,7 @@ module Beaker
         @vms << vm
 
         #enable root if user is not root
-        enable_root_on_hosts()
+        enable_root(host)
       end
     end
 
@@ -152,8 +152,10 @@ module Beaker
       end
     end
 
-    # Enables root for instances with custom username like ubuntu-amis
-    #
+    # Enables root access for a host when username is not root
+    # This method ripped from the aws_sdk implementation and is probably wrong
+    # because it iterates on a collection when there's no guarantee the collection
+    # has all been brought up in openstack yet and will thus explode
     # @return [void]
     # @api private
     def enable_root_on_hosts
@@ -162,10 +164,8 @@ module Beaker
       end
     end
 
-    # Enables root access for a host when username is not root
-    #
-    # @return [void]
-    # @api private
+    # enable root on a single host (the current one presumably) but only
+    # if the username isn't 'root'
     def enable_root(host)
       if host['user'] != 'root'
         copy_ssh_to_root(host, @options)


lib/beaker/hypervisor/openstack.rb
```diff
@@ -152,27 +154,15 @@ def cleanup
...
     def enable_root(host)
-      if host['user'] != 'root'
-        copy_ssh_to_root(host, @options)
-        enable_root_login(host, @options)
-        host['user'] = 'root'
-        host.close
-      end
+      copy_ssh_to_root(host, @options)
+      enable_root_login(host, @options)
+      host['user'] = 'root'
```
anodelman: Any chance that this is going to execute on windows?  Then you'd want 'Administrator'?

anodelman:
This doesn't get exercised either in spec or acceptance testing - just need an answer to review comment and we can merge.

fiddyspence:
Hi - I just changed that code from what was there before - I am not doing anything windows personally, so this doesn't change behaviour from before.  If someone wanted Administrator support it would need significantly more messing with from what was there!!

kevpl:
Hi @fiddyspence ,

You raise a good point, and I think your code is good -- way to save us from iterating through the whole hosts set while we were already doing that!

A few things blocking progress of this:
1. `enable_root` is not as hardened as it was before.  If someone mistakenly calls it on a windows host, its behavior has changed.  Before, it would do nothing, because the if-check took place in the method.  Now, it can set a windows host's user to `root`, which would be terrible.  Moving the if-check back into the method saves us from this circumstance.
2. You've updated the PR title to have an issue prefixed to it, but we actually care about the git commit message itself having this prefix.

Once you've updated these, I look forward to getting your code tested and merged!

anodelman:

anodelman:
Review comment still current, will close out for inactivity next week.

anodelman:

fiddyspence:
@kevpl I changed the commit - brain fade there

In terms of the root stuff, I don't see any extra protection (unless I'm missing something) for windows, but I've put it back to as close the original as I can with the 2 methods - I haven't looked but I assume that's an overr

lib/beaker/hypervisor/openstack.rb
```diff
@@ -152,20 +152,13 @@ def cleanup
...
-      @hosts.each do |host|
-        enable_root(host)
-      end
-    end
-
     # Enables root access for a host when username is not root
     #
     # @return [void]
     # @api private
+    def enable_root_on_hosts(host)
```
anodelman: So, this still has a method name that allows for 'hosts' but you've changed the signature to accept a single host.  It then wraps a method that does take a single host.  I'm pretty confused as to the purpose of this.

fiddyspence:
@kevpl Was worried about the windows thing.  This fixes the problem (iterating on a collection that doesn't exist yet) but preserves the previous behaviour that apparently saved Windows from being trashed from the point of view of the administrator account.  Minimum change.

kevpl:
@fiddyspence this change is a little unclear to us.  It looks like you just pulled out the iteration through the hosts in the method, but you made it so that you pass in a host and didn't change the method name to match that fact.  

If I've got the situation right here, would you change the method name to make this more clear (just pulling out the plural would do it)?

Once that's done, I look forward to getting this pushed through our process & into the codebase

fiddyspence:
Method is now non-plural.

lib/beaker/hypervisor/openstack.rb
```diff
@@ -133,7 +133,7 @@ def provision
         @vms << vm
 
         #enable root if user is not root
-        enable_root_on_hosts()
+        enable_root_on_host(host)
```
kevpl: @fiddyspence, taking a look at this now that I'm more clear on what the change is, I'm realizing that this line is really just using the wrong method.  

`enable_root_on_hosts` shouldn't actually disappear (it might be needed for other things later), but this usage should actually just skip to using `enable_root`, and `enable_root_on_host` shouldn't exist, since it doesn't do anything more than pass the host to `enable_root`.

My apologies on coming to the main issue late like this, and asking you to do different things because I didn't understand what was going on here.

fiddyspence: You mean the original PR was right?  I get the feeling life is too short.

fiddyspence:
So -the original enable_root code was ripped straight from the aws_sdk.rb hypervisor, so it's arbitrary whether or not that's right - it did the root thing on a collection, and frankly who knows whether that works.

enable_root_on_hosts() as a method name is arbitrary and exists only here and in aws_sdk.rb - I'd bet a critical portion of anatomy that it's unlikely to be needed anywhere else later.

For the sake of consistency with that implementation, and for someone else to worry about later, have a new commit that does what you've asked.

kevpl:
@fiddyspence thanks for doing this, and apologies for requiring so many clarifications and changes on this.

kevpl:

PR #684, opened 2015-02-13
(QENG-1834) Regression in Beaker 2.4.0 using do_scp_to

anodelman:
- repair bug and add spec test coverage


colinPL:
I see specs for `true` and `false`, but not `nil`.

colinPL:
The tests were green prior to the most recent change, which just added a new spec test.

PR #697, opened 2015-02-17
(QENG-1735) set up local windows vm without cygwin and get it working

anodelman:
- ensure that you can install PE
- deal with setting PATH/RUBY env vars
- update DSL methods to work on windows without bash

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -145,6 +145,14 @@ module Beaker
       @options.is_pe?
     end
 
+    def is_cygwin?
+      self['is_cygwin'] == nil || self['is_cygwin'] == true
+    end
+
+    def platform
+      self['platform']
+    end
+
     # True if this is a pe run, or if the host has had a 'use-service' property set.
     def use_service_scripts?
       is_pe? || self['use-service']
@@ -203,8 +211,13 @@ module Beaker
     #Examine the host system to determine the architecture
     #@return [Boolean] true if x86_64, false otherwise
     def determine_if_x86_64
-      result = exec(Beaker::Command.new("arch | grep x86_64"), :acceptable_exit_codes => (0...127))
-      result.exit_code == 0
+      if is_cygwin?
+        result = exec(Beaker::Command.new("arch | grep x86_64"), :acceptable_exit_codes => (0...127))
+        result.exit_code == 0
+      else
+        result = exec(Beaker::Command.new("wmic os get osarchitecture"), :acceptable_exit_codes => (0...127))
+        result.stdout =~ /64/
+      end
     end
 
     #@return [Boolean] true if x86_64, false otherwise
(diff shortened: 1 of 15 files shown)


anodelman:
@liamjbennett My first pass at getting cygwin-less windows support in Beaker.

liamjbennett:
@anodelman awesome. I'll review + test tomorrow. I've been raising more windows PRs so we'll need to work it out.

anodelman:
@liamjbennett i did redesign your powershell wrapper to create an array instead of a string - to bring it into line with the rest of the wrappers.  Otherwise, the rest should 'just work'(tm).

lib/beaker/host.rb
```diff
@@ -202,8 +210,13 @@ def ip
     #Examine the host system to determine the architecture
     #@return [Boolean] true if x86_64, false otherwise
     def determine_if_x86_64
-      result = exec(Beaker::Command.new("arch | grep x86_64"), :acceptable_exit_codes => (0...127))
-      result.exit_code == 0
+      if is_cygwin?
```
liamjbennett: Basically the same change I made in [PR-694](https://github.com/puppetlabs/beaker/pull/694/files). 

Worth noting that this code is very similiar to [host/windows/pkg.rb#L47](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host/windows/pkg.rb#L47) and will need to be updated if we want to continue supporting windows 2003.

lib/beaker/host.rb
```diff
@@ -218,16 +231,22 @@ def is_x86_64?
...
+        #see if the key already exists
+        elsif exec(Beaker::Command.new("grep #{key} #{env_file}"), :acceptable_exit_codes => (0..255) ).exit_code == 0
+          exec(Beaker::SedCommand.new(self['HOSTS'][name]['platform'], "s/#{key}=/#{key}=#{escaped_val}:/", env_file))
+        else
+          exec(Beaker::Command.new("echo \"#{key}=#{val}\" >> #{env_file}"))
+        end
+      else #powershell windows
+        #see if the key/value pair already exists
+        exec(Beaker::Command.new("set #{key}=%#{key}%;#{val}"))
+        exec(Beaker::Command.new("setx #{key} %#{key}%;#{val}"))
```
liamjbennett: This will need testing because I had the same thing in my own branch and found that the setx command wasn't always working and required the ```host.close``` hack you have above in order to be available.

lib/beaker/host/pswindows.rb
```diff
@@ -28,52 +28,26 @@ def self.pe_defaults
...
-              end
-            end
+    def self.foss_defaults
+      h = Beaker::Options::OptionsHash.new
+      h.merge({
+        'user'              => 'Administrator',
+        'group'             => 'Administrators',
+        'distmoduledir'     => 'C:\\ProgramData\\PuppetLabs\\puppet\\etc\\modules',
+        'sitemoduledir'     => 'C:\\usr\\share\\puppet\\modules',
+        'hieralibdir'       => 'C:\\opt\\puppet-git-repos\\hiera\\lib',
```
liamjbennett: Need to take a look at some of these because I'm not sure they are correct.

lib/beaker/host/pswindows/file.rb
```diff
@@ -2,14 +2,23 @@ module PSWindows::File
   include Beaker::CommandFactory
 
   def tmpfile(name)
-    execute("echo C:\\Windows\\Temp\\#{name}.XXXXXX")
+    execute("echo C:\\Windows\\Temp\\#{name}.%RANDOM%")
```
liamjbennett: :+1: - Nice! Didn't even know about that environment variable.

lib/beaker/host/pswindows/pkg.rb
```diff
@@ -7,15 +7,21 @@ def check_for_command(name)
   end
 
   def check_for_package(name)
-    raise "Cannot check for package #{name} on #{self}"
+    #HACK NOOP
+    #raise "Cannot check for package #{name} on #{self}"
+    0
   end
 
   def install_package(name, cmdline_args = '')
```
liamjbennett: :+1: for adding this in. I can understand why this is a no-op at the moment, but this will give me (us) something to think about and extend for the different packing types available.

lib/beaker/host/pswindows/pkg.rb
```diff
@@ -7,15 +7,21 @@ def check_for_command(name)
   end
 
   def check_for_package(name)
```
liamjbennett: This can be implemented with the following powershell:

```PowerShell
Get-ItemProperty HKLM:\Software\Microsoft\Windows\CurrentVersion\Uninstall\* | Select-Object DisplayName | Where-Object { $_.DisplayName -eq "Microsoft .NET Framework 4.5.2" } | % { $_.DisplayName }
```

lib/beaker/host_prebuilt_steps.rb
```diff
@@ -505,19 +519,26 @@ def set_env host, opts
...
+          host.add_env_var('PATH', '$PATH')
+        end
+        #add the env var set to this test host
         env.each_pair do |var, value|
           host.add_env_var(var, value)
         end
 
-        host.exec(Command.new("cat #{host[:ssh_env_file]}"))
+        # print out the working env
+        if host.is_cygwin?
```
liamjbennett: Is this block really needed? I guess it's for debugging purposes right?

lib/beaker/host_prebuilt_steps.rb
```diff
@@ -296,9 +300,11 @@ def copy_ssh_to_root host, opts
       logger = opts[:logger]
       block_on host do |host|
         logger.debug "Give root a copy of current user's keys, on #{host.name}"
-        if host['platform'] =~ /windows/
+        if host['platform'] =~ /windows/ and host.is_cygwin?
           host.exec(Command.new('cp -r .ssh /cygdrive/c/Users/Administrator/.'))
           host.exec(Command.new('chown -R Administrator /cygdrive/c/Users/Administrator/.ssh'))
+        elsif host['platform'] =~ /windows/ and not host.is_cygwin
```
liamjbennett: This should be ```host.is_cygwin?``` (missing question mark)

lib/beaker/host_prebuilt_steps.rb
```diff
@@ -296,9 +300,11 @@ def copy_ssh_to_root host, opts
...
+          # POWERSHELL SOMETHING SOMETHING HACK HACK HACK
```
liamjbennett: This should do the trick ...

```puppet
host.exec(Command.new("if exist .ssh (xcopy .ssh C:\\Users\\Administrator\\.ssh /s /e)"))
````

liamjbennett:
Also need to fix ```enable_root_login``` here: [beaker/host_prebuilt_steps.rb](https://github.com/anodelman/beaker/blob/cygwin-less/lib/beaker/host_prebuilt_steps.rb#L343:L348)

lib/beaker/host_prebuilt_steps.rb
anodelman: Yeah, just to have the env printed out in the logs to make it clear what state beaker is running in.

anodelman:
Yeah, I do have to reconnect after puppet installation to make the pathing work - which I don't understand.  Wish that there was a better way to refresh, but haven't found anything in my digging.

anodelman:
Only known puppet foss debian 7 error.

PR #699, opened 2015-02-17
(QENG-1429) added clause to catch Beaker & Minitest assertions

kevpl:


kevpl:

lib/beaker/dsl/helpers.rb
```diff
@@ -564,6 +564,8 @@ def with_puppet_running_on host, conf_opts, testdir = host.tmpdir(File.basename(
 
           yield self if block_given?
 
+        rescue Beaker::DSL::Assertions, Minitest::Assertion => early_assertion
```
anodelman: I believe that you only need to rescue Beaker::DSL::Assertions, as they include Minitest::Assertions.

kevpl: I tried that, and it didn't work for this.  There's another use case for this that I changed to, and that fixed the issue: https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/structure.rb#L112

PR #701, opened 2015-02-18
(QENG-1852) install_puppet unable to install windows msi (regression)

anodelman:
- busted curl command on windows
- busted msiexec command for installation
- i cry myself to sleep every night

### DIFF
--- a/lib/beaker/dsl/install_utils.rb
+++ b/lib/beaker/dsl/install_utils.rb
@@ -807,8 +807,8 @@ module Beaker
         end
 
         if host['is_cygwin'].nil? or host['is_cygwin'] == true
-          dest = "/cygdrive/c/Windows/Temp/#{host['dist']}.msi"
-          on host, "curl -O #{dest} #{link}"
+          dest = "#{host['dist']}.msi"
+          on host, "curl -O #{link}"
 
           #Because the msi installer doesn't add Puppet to the environment path
           #Add both potential paths for simplicity
@@ -826,7 +826,7 @@ module Beaker
           on host, "if not exist #{host['distmoduledir']} (md #{host['distmoduledir']})"
         end
 
-        on host, "msiexec /qn /i #{dest}"
+        on host, "cmd /C 'start /w msiexec.exe /qn /i #{dest}'"
       end
 
       # Installs Puppet and dependencies from dmg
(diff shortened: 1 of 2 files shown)


kurtwall:
LGTM.

lib/beaker/dsl/install_utils.rb
```diff
@@ -807,8 +807,8 @@ def install_puppet_from_msi( host, opts )
         end
 
         if host['is_cygwin'].nil? or host['is_cygwin'] == true
-          dest = "/cygdrive/c/Windows/Temp/#{host['dist']}.msi"
-          on host, "curl -O #{dest} #{link}"
+          dest = "#{host['dist']}.msi"
```
sschneid: Is this definition needed since it doesn't look like it's being used anywhere?

lib/beaker/dsl/install_utils.rb
```diff
@@ -807,8 +807,8 @@ def install_puppet_from_msi( host, opts )
...
+          on host, "curl -O #{link}"
```
joshcooper: msiexec can install MSI's from a URL, so we don't actually need to curl it down.

Also `msiexec` will install puppet asynchronously. To block waiting for the install to complete you have to do:

`"cmd /c start /w msiexec /qn /i #{link}"`

lib/beaker/dsl/install_utils.rb
anodelman: Yup, it's used at the end of the method.

sschneid: Wouldn't it be out of scope at that point?

If I'm barking up nonexistent trees, ignore me.

anodelman:
Updated based upon review comments.

anodelman:
Failure caused by test timeout, not releated to patch.

kevpl:

lib/beaker/dsl/install_utils.rb
```diff
@@ -826,7 +826,7 @@ def install_puppet_from_msi( host, opts )
           on host, "if not exist #{host['distmoduledir']} (md #{host['distmoduledir']})"
         end
 
-        on host, "msiexec /qn /i #{dest}"
+        on host, "cmd /C 'start /w msiexec.exe /qn /i #{dest}'"
```
colinPL: `#{dest}` or `#{link}`? The `curl` command above drops the `#{dest}` portion, but this call is specifically using it.

anodelman: The previous command version was broken, curl on windows does not accept the second destination parameter.  So, we download the 'link' and it ends up in a destination 'dest', we use curl to download and msiexec to install.

Issue #704, opened 2015-02-19
Beaker puppet_module_install and scp_to is slow in larger environments


nemski:
This was originally raised in #143, but not implemented as the original speed issue was resolved by forcing a chunk_size. My problem is I have a large number of modules I need to copy across for testing (40+) and using puppet_module_install_on takes up-to 10 minutes!

There are two problems here:

* puppet_module_install_on uses the ignore method which actually slows down copying, because each individual file is scped one by one! The resulting spawning and reaping of child processes takes a long time.

* scp blind copies and uses encryption

I've implemented using rsync instead of scp and will submit a PR shortly, here's a comparison of time it takes to copy across modules in my environment (1GB connection and sub 1ms latency):

* puppet_module_install_on with scp => 463 seconds
* puppet_module_install_on with rsync (sync nothing) => 332 seconds
* scp_to => 179 seconds
* rsync_to (sync nothing) => 21 seconds
* rsync_to (sync all) => 19 seconds

Rsync speed starting with no files and all files are comparable, the speed increase seems to be lack of encryption/decryption. This is probably because the modules consist of lots of small files (around 4,000 in total).

Ultimately puppet_module_install_on with rsync should use the '--exclude' argument to ignore certain files, but it's unlikely I'll include this in my PR.

anodelman:
PR merged, closing.  https://github.com/puppetlabs/beaker/pull/711

PR #705, opened 2015-02-19
(gh-704) Add an option to do_scp_to to allow the use of rsync

nemski:
Without this patch copying a large number of files can take a long time. This patch allows the use of rsync, by adding a :copy_method argument to puppet_module_install_on or scp_to command, which speeds up file transfers. Syncing files instead of blind copying them itself does not provide a performance increase, but rather the lack of encryption significantly increases speed. Puppet_module_install_on is still slow (although faster than scp) with this patch, because the :ignore method is used with causes each file to be individually synced.

### DIFF
--- a/beaker.gemspec
+++ b/beaker.gemspec
@@ -46,6 +46,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'aws-sdk', '~> 1.57'
   s.add_runtime_dependency 'docker-api'
   s.add_runtime_dependency 'fog', '~> 1.25'
+  s.add_runtime_dependency 'rsync'
 
   # So fog doesn't always complain of unmet AWS dependencies
   s.add_runtime_dependency 'unf', '~> 0.1'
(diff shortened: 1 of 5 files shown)


electrical:
The only feature i see is missing for the rsync part is to ignore files?

nemski:
As per the host.rb file line 365 rsync will ignore files in the same way scp does. This is not the most efficient way to do it, it should be using the "--exclude" argument instead of syncing individual files, but it will work.

electrical:
Ah okay, sorry. missed that part. ( not enough tea )

jonnytpuppet:
I created my own PR for pretty much the same thing, without realising that this one was one already there! Oops! My one does have support for rsync's native --exclude, so it's nice and fast in that situation.

Perhaps we could find a best of both worlds solution, if mine is missing something?

https://github.com/puppetlabs/beaker/pull/711

nemski:
@jonnytpuppet I like yours better, you've better extracted the rsync function from the scp function. Closing this pull request.

Issue #700, opened 2015-02-18
Unable to set docker_options in order to override read_timeout


madAndroid:
I can't seem to see any obvious way of overriding the docker options such as `:read_timeout`.
In the code there is a section which ensures that `::Docker.config` options are not overwritten if they already exist => (from [docker.rb](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/hypervisor/docker.rb#L11))
```
 ::Docker.options = { :write_timeout => 300, :read_timeout => 300 }.merge(::Docker.options || {})
```

... but I can't see a (simple?) way of setting those config options - am I missing something?

Ideally, I'd like to add a `docker_options` line to the nodeset yaml to be able to override those default options, but that doesn't seem to work, as far as I can tell.

anodelman:
Merged, https://github.com/puppetlabs/beaker/pull/706.

PR #706, opened 2015-02-19
GH-700 - Allow Docker::options to be set from config hash

madAndroid:
Addresses #700 - This allows `docker_options` to be set in the nodeset YAML, allowing the timeouts to be easily extended.

### DIFF
--- a/lib/beaker/hypervisor/docker.rb
+++ b/lib/beaker/hypervisor/docker.rb
@@ -8,7 +8,9 @@ module Beaker
       @hosts = hosts
 
       # increase the http timeouts as provisioning images can be slow
-      docker_options = ::Docker.options.empty? ? @options[:docker_options] : ::Docker.options
+      unless ::Docker.options.nil?
+        docker_options = ::Docker.options.empty? ? @options[:docker_options] : ::Docker.options
+      end
       ::Docker.options = { :write_timeout => 300, :read_timeout => 300 }.merge(docker_options || {})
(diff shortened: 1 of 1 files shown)


electrical:
I think `::Docker.options` should still be used for the complete options hash to ensure we don't unset any options that have been set already.

I think the end result should be:

::Docker.options + default added options ( timeout stuff ) + optional settings from hosts file.

What do you think?

madAndroid:
Sounds good :) I wasn't quite sure how to go about setting `::Docker.options`: whether that would be done in the `Rakefile` or `spec_helper_acceptance.rb`, but neither of those options appeared to work

I'll have a look at amending what I've got into that suggested approach

madAndroid:
I've amended the merge, as suggested

electrical:
LGTM

anodelman:
Our internal testing will not touch docker code, so I'll merged based upon @electrical's :+1:

madAndroid:
Thanks Alice :-)

Just an idea -  is it worth setting up a set of travis jobs to run the
tests and gain some visibility outside of the puppetlabs test suite? :-)
might give some additional confidence, especially in instances such as
these, where there are no internal tests to cover the use case.

anodelman:
We used to use travis for unit testing, but quit because it wasn't configurable enough for our uses.  May need to re-examine.  We do have an internal goal of shifting the beaker pipeline to the public jenkins instance so that is is viewable for everyone - that will make everything better.

PR #708, opened 2015-02-20
(QENG-1847) changed internal host path attribute usage to host.puppet me...

kevpl:
...thod calls where possible

### DIFF
--- a/lib/beaker/dsl/helpers.rb
+++ b/lib/beaker/dsl/helpers.rb
@@ -553,7 +553,7 @@ module Beaker
         end
 
         begin
-          backup_file = backup_the_file(host, host['puppetconfdir'], testdir, 'puppet.conf')
+          backup_file = backup_the_file(host, host.puppet('master')['confdir'], testdir, 'puppet.conf')
           lay_down_new_puppet_conf host, conf_opts, testdir
 
           if host.use_service_scripts? && !service_args[:bypass_service_script]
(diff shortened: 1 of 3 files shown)


kevpl:

kevpl:

joshcooper:
The failures are caused because the puppet package is not creating the `/usr/share/puppet/modules` system module directory leading to:

```
Cannot create /usr/share/puppet/modules/crick; parent directory /usr/share/puppet/modules does not exist
```

This is filed as https://tickets.puppetlabs.com/browse/PUP-4001

lib/beaker/dsl/helpers.rb
```diff
@@ -553,7 +553,7 @@ def with_puppet_running_on host, conf_opts, testdir = host.tmpdir(File.basename(
         end
 
         begin
-          backup_file = backup_the_file(host, host['puppetconfdir'], testdir, 'puppet.conf')
+          backup_file = backup_the_file(host, host.puppet['confdir'], testdir, 'puppet.conf')
```
joshcooper: I believe `with_puppet_running_on` is only ever called on the master node, in which case, I think we always want to do `host.puppet('master')['confdir']`.

Currently the agent and master share the confdir (so it shouldn't matter), but the confdir may be different in the future, e.g. the master's vardir is changing https://tickets.puppetlabs.com/browse/SERVER-369

anodelman: If 'with_puppet_running_on' can only be called by the master we should put in a check for that and fail out otherwise.

kevpl:
review changes added

rji:
retest this please

rji:
@anodelman,@kevpl - All cells are green except the expected failure from FOSS Puppet on Debian.

PR #711, opened 2015-02-21
Added support for rsync

jonnytpuppet:
During my testing from Belfast, using vcloud+beaker is very, very slow over the built-in SCP. It is common for an acceptance test to take 5 or 10 minutes to start due to the module upload. Rsync reduced this to literally a few seconds!

As such, I have created some code to add rsync support for beaker. It requires the 'rsync' gem which is a wrapper for the system binary (thus, you need rsync already on your system). To use it, you simply use the following in your spec helpers:

```
copy_module_to(host, :source => proj_root, :module_name => 'firewall', :protocol => 'rsync')
```
or

```
# Same arguments as scp_to 
rsync_to host, proj_root, "#{host['distmoduledir']}/netscaler", {:ignore => [".bundle", ".git", ".vagrant"]}
```
or

```
create_remote_file(master, File.join(path, "site.pp"), pp, :protocol => 'rsync')
```

It's not fully comprehensive just yet, but I feel these are a good start, especially since the main slowdown is the initial module copy.

I have written some basic unit tests for my new code which work almost pretty well, except I'm having an issue which I'll make a separate comment for.

Thanks!

### DIFF
--- a/beaker.gemspec
+++ b/beaker.gemspec
@@ -38,6 +38,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'net-ssh', '~> 2.9'
   s.add_runtime_dependency 'net-scp', '~> 1.2'
   s.add_runtime_dependency 'inifile', '~> 2.0'
+  s.add_runtime_dependency 'rsync', '~> 1.0.9'
 
   # Optional provisioner specific support
   s.add_runtime_dependency 'rbvmomi', '~> 1.8'
--- a/lib/beaker/dsl/helpers.rb
+++ b/lib/beaker/dsl/helpers.rb
@@ -183,7 +183,7 @@ module Beaker
         end
       end
 
-      # Move a local file to a remote host
+      # Move a local file to a remote host using scp
       # @note If using {Beaker::Host} for the hosts *scp* is not
       #   required on the system as it uses Ruby's net/scp library.  The
       #   net-scp gem however is required (and specified in the gemspec.
@@ -206,6 +206,27 @@ module Beaker
           end
           @result = host.do_scp_to(from_path, to_path, opts)
           @result.log logger
+          @result
+        end
+      end
+
+      # Move a local file or directory to a remote host using rsync
+      # @note rsync is required on the local host.
+      #
+      # @param [Host, #do_scp_to] host A host object that responds like
+      #                                {Beaker::Host}.
+      # @param [String] from_path A local path to a file or directory.
+      # @param [String] to_path   A remote path to copy *from_path* to.
+      # @!macro common_opts
+      #
+      # @return [Result] Returns the result of the rsync operation
+      def rsync_to host, from_path, to_path, opts = {}
+        block_on host do | host |
+          if host['platform'] =~ /windows/ && to_path.match('`cygpath')
+            result = host.echo "#{to_path}"
+            to_path = result.raw_output.chomp
+          end
+          @result = host.do_rsync_to(from_path, to_path, opts)
           @result
         end
       end
@@ -238,13 +259,24 @@ module Beaker
       # @param [String] file_path A remote path to place *file_content* at.
       # @param [String] file_content The contents of the file to be placed.
       # @!macro common_opts
+      # @option opts [String] :protocol Name of the underlying transfer method.
+      #                                 Valid options are 'scp' or 'rsync'.
       #
       # @return [Result] Returns the result of the underlying SCP operation.
       def create_remote_file(hosts, file_path, file_content, opts = {})
         Tempfile.open 'beaker' do |tempfile|
           File.open(tempfile.path, 'w') {|file| file.puts file_content }
 
-          scp_to hosts, tempfile.path, file_path, opts
+          opts[:protocol] ||= 'scp'
+          case opts[:protocol]
+            when 'scp'
+              scp_to hosts, tempfile.path, file_path, opts
+            when 'rsync'
+              rsync_to hosts, tempfile.path, file_path, opts
+            else
+              logger.debug "Unsupported transfer protocol, returning nil"
+              nil
+          end
         end
       end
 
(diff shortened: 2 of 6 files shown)


spec/beaker/host_spec.rb
```diff
@@ -553,6 +553,32 @@ module Beaker
...
+
+        key = host['ssh']['keys']
+        if key.is_a? Array
+          key = key.first
+        end
+
+        rsync_args = [ 'source', 'target', ['-az', "-e \"ssh -i #{key} -p 22 -o 'StrictHostKeyChecking no'\"", "--exclude '.bundle'"] ]
+
+        expect( host ).to receive(:to_s).and_return('host.example.org')
+        #expect(Rsync.host).to eq('root@host.example.org')
```
jonnytpuppet: I'm having an issue to test that Rsync.host gets set to this hostname. During the test run, it comes back as nil, presumably due to an execution order issue.

Does anyone have any ideas?

jonnytpuppet:
I submitted this PR without realising that there is another one open (705) trying to do the same thing.

Mine does have support for using :ignore using rsync's native --exclude, so I guess it would be faster when wanting to ignore certain files?

jonnytpuppet:
Can anyone help me on why jenkins is failing? The unit tests all pass when I run them manually on my laptop..

Thanks

jonnytpuppet:
OK, so I fixed the docs issue, but jenkins is still failing and I'm not sure why. Any help would be appreciated, thanks!

electrical:
@jonnytpuppet it could be failing due to other reasons since it does more tests then just the unit tests you can run.
I'm sure later today @anodelman can have a look at it.

anodelman:
We have an unrelated failure in our smoketests that is causing your red.

lib/beaker/dsl/helpers.rb
```diff
@@ -210,6 +210,27 @@ def scp_to host, from_path, to_path, opts = {}
...
+      #                                {Beaker::Host}.
+      # @param [String] from_path A local path to a file or directory.
+      # @param [String] to_path   A remote path to copy *from_path* to.
+      # @!macro common_opts
+      #
+      # @return [Result] Returns the result of the rsync operation
+      def rsync_to host, from_path, to_path, opts = {}
+        block_on host do | host |
+          if host['platform'] =~ /windows/ && to_path.match('`cygpath')
+            result = on host, "echo #{to_path}"
```
anodelman: There is a built in 'echo' command per host (host.echo), let's use that here.

beaker.gemspec
```diff
@@ -38,6 +38,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'net-ssh', '~> 2.9'
   s.add_runtime_dependency 'net-scp', '~> 1.2'
   s.add_runtime_dependency 'inifile', '~> 2.0'
+  s.add_runtime_dependency 'rsync'
```
anodelman: Please give an initial major version so that we don't get weird bustage on version bumps in external projects.

lib/beaker/host.rb
```diff
@@ -389,6 +390,79 @@ def do_scp_from source, target, options
...
+      # We disable prompt when host isn't known
+      ssh_args << "-o 'StrictHostKeyChecking no'"
+
+      if not ssh_args.empty?
+        rsync_args << "-e \"ssh #{ssh_args.join(' ')}\""
+      end
+
+      if opts.has_key?(:ignore) and not opts[:ignore].empty?
+        opts[:ignore].map! do |value|
+          "--exclude '#{value}'"
```
anodelman: Does 'exclude' behave in the same way as 'ignore', just want to make sure that we don't slightly differ from expected behavior.

lib/beaker/host.rb
```diff
@@ -389,6 +390,79 @@ def do_scp_from source, target, options
...
+      ssh_args = []
+
+      if not File.file?(from_path) and not File.directory?(from_path)
+        raise IOError, "No such file or directory - #{from_path}"
+      end
+
+      # We enable achieve mode and compression
+      rsync_args << "-az"
+
+      if not ssh_opts.has_key?('user')
```
anodelman: Use host['user'] for this - it is assumed to be the host that we log onto SUTs with.

anodelman:
Please associate your commit with either a jira bug number (QENG-XXX) or a github issue number (gh-xxx) for future tracking.

jonnytpuppet:
Adding jira ticket ref: https://tickets.puppetlabs.com/browse/QENG-1878

jonnytpuppet:
@anodelman I've pushed some code to this PR as per the comments above.

Regarding your query about --exclude, unfortunately I do not believe it behaves the same way as the current scp/ignore implementation. Looking at the current codebase, it seems that there is some Regex going on to do some sort of pattern matching?

Rsync's --exclude uses it's own filter rules, which can be found here: http://linux.die.net/man/1/rsync (Look for the section called "Filter Rules").

We could add some similar code to the do_rsync_to method to perform a ruby Dir lookup on the source tree, and use it to generate an exclude array where each found directory that is applicable for excluding would have an array element prefixed with "--exclude"; however, that may lead to "riskier"/more unknown behaviour, given how long the rsync command would be. 

Alternatively, we could just make it a design paradigm that we state in the docs that if :ignore is used with rsync, then we have to expect rsync's native --exclude behaviour (and thus it's own pattern matching).

Let me know how you would like me to proceed, with regards to the above :)

I have tested my branch against some of the acceptance tests in puppetlabs-firewall, and it all seems ok...

anodelman:
Add the jira ticket to your commit as well (with QENG-xxx in the title).

anodelman:
Only failure now is a known issue with debian 7 foss puppet.

jonnytpuppet:
@anodelman I've added the ticket number to the commit, thanks :)

anodelman:
retest this please

anodelman:
Down to known debian 7 foss testing errors.

PR #713, opened 2015-02-24
(MAINT) fix broken spec tests

anodelman:
- accidental breakage of spec tests
- also removed deprecated rspec method call


kevpl:
Since this is only a spec file change, and specs have already passed, I'm merging this.

PR #714, opened 2015-02-24
(QENG-1721) added installation support of dev versions of windows puppet...

kevpl:
...-agent

### DIFF
--- a/lib/beaker/dsl/install_utils.rb
+++ b/lib/beaker/dsl/install_utils.rb
@@ -1276,6 +1276,11 @@ module Beaker
         when /^(debian|ubuntu|cumulus)$/
           release_path << "deb/#{codename}"
           release_file = "puppet-agent_#{opts[:version]}-1_#{arch}.deb"
+        when /^windows$/
+          release_path << 'windows'
+          onhost_copy_base = '`cygpath -smF 35`/'
+          arch_suffix = arch =~ /64/ ? '64' : '86'
+          release_file = "puppet-agent-x#{arch_suffix}.msi"
         else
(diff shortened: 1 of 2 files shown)


lib/beaker/dsl/install_utils.rb
```diff
@@ -1279,6 +1284,10 @@ def install_puppetagent_dev_repo( host, opts )
         when /^(debian|ubuntu|cumulus)$/
           on host, "dpkg -i --force-all #{onhost_copied_file}"
           on host, "apt-get update"
+        when /^windows$/
+          result = on host, "echo #{onhost_copied_file}"
```
anodelman: There is a built in host.echo command - you should use that.

lib/beaker/dsl/install_utils.rb
```diff
@@ -1279,6 +1284,10 @@ def install_puppetagent_dev_repo( host, opts )
...
+          onhost_copied_file = result.raw_output.chomp
+          on host, "cmd /C 'start /w #{onhost_copied_file}'"
```
anodelman: If you use {:cmdexe => true} in a command object it will correctly append 'cmd /C' to the command string.

joshcooper: I'm also not sure about the need for single quotes, we are not currently doing that in puppet: https://github.com/puppetlabs/puppet/blob/master/acceptance/lib/puppet/acceptance/install_utils.rb#L252

kevpl:
review feedback included

kevpl:
just the debian foss testing errors

anodelman:
Could you back fill some spec tests for this?

PR #717, opened 2015-02-26
(QENG-1891) added /opt/puppetlabs/bin directory to path for AIO

kevpl:
this is a stop-gap measure to unblock testing, as opposed to the larger work to update all paths to the latest spec, which will be done in QENG-1841.

### DIFF
--- a/lib/beaker/host/unix.rb
+++ b/lib/beaker/host/unix.rb
@@ -64,7 +64,7 @@ module Unix
       h = self.foss_defaults
       h['puppetserver-confdir'] = '/etc/puppetlabs/puppetserver/conf.d'
       h['puppetservice']  = 'puppetserver'
-      h['puppetbindir']   = '/opt/puppetlabs/agent/bin'
+      h['puppetbindir']   = '/opt/puppetlabs/agent/bin:/opt/puppetlabs/bin'
       h['puppetbin']      = "#{h['puppetbindir']}/puppet"
       h['puppetpath']     = '/etc/puppetlabs/agent'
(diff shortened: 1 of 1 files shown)


lib/beaker/host/unix.rb
```diff
@@ -64,7 +64,7 @@ def self.aio_defaults
       h = self.foss_defaults
       h['puppetserver-confdir'] = '/etc/puppetlabs/puppetserver/conf.d'
       h['puppetservice']  = 'puppetserver'
-      h['puppetbindir']   = '/opt/puppetlabs/agent/bin'
+      h['puppetbindir']   = '/opt/puppetlabs/agent/bin:/opt/puppetlabs/bin'
```
joshcooper: currently the puppet-agent package doesn't symlink files into /opt/puppetlabs/bin. Also many puppet acceptance tests that execute on Unix assume puppetbindir is a single directory (windows doesn't have this issue). So I think if we want to unblock testing, this should just be changed to `h['puppetbindir']   = '/opt/puppetlabs/puppet/bin`

joshcooper: Hmm just saw @jeffmccune's ticket. We need to talk about how to do this in a way that unblocks puppetserver, but doesn't break puppet.

kevpl:
@joshcooper are you referring to tests outside of the ones we cover? I'm only seeing the known debian failures in the latest Jenkins run.

kevpl:

kevpl:
since this change is also included in QENG-1841, I've decided to close this as a duplicate, and just get the other one in.

Issue #732, opened 2015-03-04
Better Openstack IP address assignment


fiddyspence:
When not using floating IPs, this code allows openstack to have automagically assigned an IP
which we then just use

anodelman:
In PR:

https://github.com/puppetlabs/beaker/pull/733

anodelman:
Follow up here:

https://tickets.puppetlabs.com/browse/BKR-28

PR #681, opened 2015-02-11
(gh-680)[openstack] enable support for multiple regions

fiddyspence:
Add optional :openstack_region config key to tell beaker which region to use.

Add support for non-floating IPs, if the machine is provisioned automagically with one.

Fixes #680

PR #733, opened 2015-03-04
(gh-732) Better Openstack IP address assignment

fiddyspence:
Allows us to have already used an Openstack assigned address, rather than relying on floating IPs

### DIFF
--- a/lib/beaker/hypervisor/openstack.rb
+++ b/lib/beaker/hypervisor/openstack.rb
@@ -115,13 +115,28 @@ module Beaker
         # Associate a public IP to the server
         # Create if there are no floating ips available
         #
-        ip = @compute_client.addresses.find { |ip| ip.instance_id.nil? }
-        if ip.nil?
-          @logger.debug "Creating IP for #{host.name} (#{host[:vmhostname]})"
-          ip = @compute_client.addresses.create
+        # Do we already have an address?
+        @logger.debug vm.addresses
+
+        begin
+          if vm.addresses[@options[:openstack_network]]
+            address = vm.addresses[@options[:openstack_network]].map{ |network| network['addr'] }.first
+          end
+        rescue NoMethodError
+          @logger.debug "No current address retrievable from OpenStack data"
         end
-        ip.server = vm
-        host[:ip] = ip.ip
+        unless address
+          ip = @compute_client.addresses.find { |ip| ip.instance_id.nil? }
+
+          if ip.nil?
+            @logger.debug "Creating IP for #{host.name} (#{host[:vmhostname]})"
+            ip = @compute_client.addresses.create
+          end
+          ip.server = vm
+          address = ip.ip
+        end
+        host[:ip] = address
+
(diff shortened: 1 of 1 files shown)


fiddyspence:
This code appeared in PR #681 first - broke out into its own PR

lib/beaker/hypervisor/openstack.rb
```diff
@@ -115,13 +115,27 @@ def provision
...
-          @logger.debug "Creating IP for #{host.name} (#{host[:vmhostname]})"
-          ip = @compute_client.addresses.create
+        # Do we already have an address?
+        @logger.debug vm.addresses
+
+        begin
+          if vm.addresses[@options[:openstack_network]]
+            address = vm.addresses[@options[:openstack_network]].map{ |network| network['addr'] }.first
+          end
+        rescue
```
anodelman: Let's figure out what exceptions you are really catching here, and also include some logger information on rescue.

anodelman:
Review comment still current.

anodelman:

anodelman:
Here's your associated beaker jira bug:

https://tickets.puppetlabs.com/browse/BKR-28

lib/beaker/hypervisor/openstack.rb
fiddyspence: All kinds of nil errors, i.e. if the hash we get back from Fog doesn't have an @options[:openstacknetwork] key (or if the select doesn't work because broken

anodelman: Please determine a reasonable set of catchable exceptions - we do not support empty rescue statements, especially with no error messaging.  It just makes bug tracking impossible in the long run.

fiddyspence: I've stuck in a rescue for NoMethodError with a debug message and updated the commit - that should catch the only problem I saw with it (i.e. a hash without those elements in it).

lib/beaker/hypervisor/openstack.rb
```diff
@@ -115,13 +115,28 @@ def provision
...
+          ip = @compute_client.addresses.find { |ip| ip.instance_id.nil? }
+
+          if ip.nil?
+            @logger.debug "Creating IP for #{host.name} (#{host[:vmhostname]})"
+            ip = @compute_client.addresses.create
+          end
+          ip.server = vm
+          address = ip.ip
         end
-        ip.server = vm
```
anodelman: ip.server used to be set for all cases - now it will not be set if the first being-rescue block is successful.  Is that the desired state here?

fiddyspence: Previously there was only one case - the aim of the code is to set the address of the host.  I don't think ip.server is relevant otherwise

anodelman:
test this please

mcanevet:
This PR totally broke my setup. Openstack does not allocate a floating IP anymore, and beaker tries to connect to VM's private IP, which obviously can't work...

PR #734, opened 2015-03-04
(BKR-5) added a default log_prefix of the hostfile name, and the --log-p...

kevpl:
...refix CLI option

### DIFF
--- a/lib/beaker/logger.rb
+++ b/lib/beaker/logger.rb
@@ -302,12 +302,18 @@ module Beaker
 
     # Utility method to centralize dated log folder generation
     #
-    # @param [String] base_dir path of the directory for the dated log folder to live in
-    # @param [Time] timestamp the timestamp that should be used to generate the dated log folder
+    # @param [String] base_dir Path of the directory for the dated log folder to live in
+    # @param [String] log_prefix Prefix to use for the log files
+    # @param [Time] timestamp Timestamp that should be used to generate the dated log folder
+    #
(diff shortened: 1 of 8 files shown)


kevpl:
failures unrelated to change

anodelman:
Let's do a few proof of concepts here:

- can execute two matching host files in different windows (same base dir)
- can ensure that LATEST is correct for new dir structure

kevpl:
added in work for `log_path`, that now doesn't use the latest symlink, but updates it correctly, and testing for that as well.

kevpl:

kevpl:
failures unrelated to change

anodelman:
This works fine locally.  It does change the path of junit output from junit/***time***/log.xml to junit/***config***/***time***/log.xml - this will involve updates to dependent jenkins jobs that publish junit output post beaker testing.

anodelman:
@kevpl now that everything is getting dropped into different log locations please add a line at the end of beaker execution to say "your log is here: ***, your junit output is here: ****" to make it easier to track.

anodelman:
retest this please

anodelman:
The new structure does appear to break our junit output pathing assumptions in jenkins.  Will look at this closer today and confer with @colinPL .

anodelman:
retest this please

anodelman:
I take it back, the jenkins junit pathing continues to work with this patch applied.

Issue #702, opened 2015-02-18
Beaker/vagrant ssh issues on centos6


lavaman:
I'm having a problem getting beaker tests to run on centos boxes. I did a clean clone of puppetlabs-apache on a new osx yosemite and after sshd_config gets the sed and restarts, ssh fails. A vagrant reload makes it work again, but I can't get single-run tests working. I've tried on centos-64 and centos-65 with the same results. I've verified this works fine on ubuntu-server-1404. Using included nodesets and puppetlabs vagrant boxes. Debug output in gist:
https://gist.github.com/lavaman/9ee69ee1075f6629a938

lavaman:
Tried putting in delay and doing a second restart here:
https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host_prebuilt_steps.rb#L347
Still no luck.

anodelman:
What vagrant version are you using?

lavaman:
latest. (1.7.2)

anodelman:
Attempting to reproduce now.

lavaman:
You may have to manually install nokogiri to point it at the right libxml2 dir:
http://stackoverflow.com/questions/19643153/error-to-install-nokogiri-on-osx-10-9-maverick
The first answer has the command for Yosemite.

anodelman:
My nokogiri installed no problem, beaker worked fine on vagrant centos5 box - testing centos6 now.

lavaman:
Here is my bundle list as well, just in case:
Using rake 10.4.2
Using CFPropertyList 2.3.0
Using i18n 0.7.0
Using json 1.8.2
Using minitest 5.5.1
Using thread_safe 0.3.4
Using tzinfo 1.2.2
Using activesupport 4.2.0
Using addressable 2.3.7
Using archive-tar-minitar 0.5.2
Using extlib 0.9.16
Using multi_json 1.10.1
Using autoparse 0.3.3
Using mini_portile 0.6.2
Using nokogiri 1.6.6.2
Using aws-sdk-v1 1.62.0
Using aws-sdk 1.62.0
Using excon 0.44.2
Using docker-api 1.19.0
Using fission 0.5.0
Using builder 3.2.2
Using formatador 0.2.5
Using mime-types 2.4.3
Using net-ssh 2.9.2
Using net-scp 1.2.1
Using fog-core 1.28.0
Using fog-xml 0.1.1
Using fog-atmos 0.1.0
Using fog-json 1.0.0
[...]

Issue #656, opened 2015-01-29
"sudo: sorry, you must have a tty to run sudo" when testing on el


mcanevet:

anodelman:
Merged.

PR #740, opened 2015-03-09
(gh-702) sshd restart doesn't work with some el6...

anodelman:
... variants especially centos - with vagrant provisioner

For centos6 & vagrant, this pull fixed it: https://github.com/puppetlabs/beaker/pull/545
& fixing https://github.com/puppetlabs/beaker/issues/656 re-broke it by undoing pull 545

- Instead of service sshd restart, service sshd reload fixes the issue.
  Even systemctl takes reload.


petems:
:+1:

PR #741, opened 2015-03-11
 (BKR-8) PE 3.8 on Sles11 can't find puppet in path after install...

anodelman:
... completes successfully.

- add the ~/.ssh/environment to a /etc/profile.d/beaker_env.sh script
- ensure that beaker_env.sh is kept up to date and correct with state of
  current env

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -230,6 +230,29 @@ module Beaker
       @x86_64 ||= determine_if_x86_64
     end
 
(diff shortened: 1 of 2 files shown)


ghoneycutt:
links are to an invalid domain

anodelman:
Our beaker jenkins instance is not publicly reachable (but we have plans to fix that in the near term).

Test failures are unrelated to patch.

PR #742, opened 2015-03-12
Create /etc/hosts for Docker instances

madAndroid:
This PR adds support for adding hosts to `/etc/hosts` on all docker instances in the nodeset; we cannot use the `hack_etc_hosts` method here, since the `host['ip']` for the docker instance is set to `0.0.0.0` and the ssh port which has been exposed for the particular container, therefore using `hack_etc_hosts` results in a hosts file which contains `0.0.0.0` for all entries. Instead, I've created a `hack_docker_etc_hosts` method, and used the correct entry in the `docker_container` key to set the IPs.

### DIFF
--- a/lib/beaker/host_prebuilt_steps.rb
+++ b/lib/beaker/host_prebuilt_steps.rb
@@ -321,11 +321,7 @@ module Beaker
     def hack_etc_hosts hosts, opts
       etc_hosts = "127.0.0.1\tlocalhost localhost.localdomain\n"
       hosts.each do |host|
-        if host['hypervisor'] == 'docker'
-          etc_hosts += "#{host['docker_container'].json["NetworkSettings"]["IPAddress"]}\t#{host.name}\n"
-        else
-          etc_hosts += "#{host['ip'].to_s}\t#{host[:vmhostname] || host.name}\n"
-        end
+        etc_hosts += "#{host['etc_host_ip'] || host['ip'].to_s}\t#{host[:vmhostname] || host.name}\n"
       end
       hosts.each do |host|
         set_etc_hosts(host, etc_hosts)
--- a/lib/beaker/host_prebuilt_steps.rb
+++ b/lib/beaker/host_prebuilt_steps.rb
@@ -321,7 +321,11 @@ module Beaker
     def hack_etc_hosts hosts, opts
       etc_hosts = "127.0.0.1\tlocalhost localhost.localdomain\n"
       hosts.each do |host|
-        etc_hosts += "#{host['ip'].to_s}\t#{host[:vmhostname] || host.name}\n"
+        if host['hypervisor'] == 'docker'
+          etc_hosts += "#{host['docker_container'].json["NetworkSettings"]["IPAddress"]}\t#{host.name}\n"
+        else
+          etc_hosts += "#{host['ip'].to_s}\t#{host[:vmhostname] || host.name}\n"
+        end
       end
       hosts.each do |host|
         set_etc_hosts(host, etc_hosts)
--- a/lib/beaker/hypervisor/docker.rb
+++ b/lib/beaker/hypervisor/docker.rb
@@ -66,18 +66,8 @@ module Beaker
         host['docker_image'] = image
       end
 
-      hack_docker_etc_hosts @hosts, @options
+      hack_etc_hosts @hosts, @options
 
-    end
-
-    def hack_docker_etc_hosts hosts, opts
-      etc_hosts = "127.0.0.1\tlocalhost localhost.localdomain\n"
-      hosts.each do |host|
-        etc_hosts += "#{host['docker_container'].json["NetworkSettings"]["IPAddress"]}\t#{host.name}\n"
-      end
-      hosts.each do |host|
-        set_etc_hosts(host, etc_hosts)
-      end
     end
 
     def cleanup
--- a/lib/beaker/hypervisor/docker.rb
+++ b/lib/beaker/hypervisor/docker.rb
@@ -65,6 +65,16 @@ module Beaker
         host['docker_container'] = container
         host['docker_image'] = image
       end
+
+      ## Set /etc/hosts :
+      etc_hosts = "127.0.0.1\tlocalhost localhost.localdomain\n"
+      @hosts.each do |host|
+        etc_hosts += "#{host['docker_container'].json["NetworkSettings"]["IPAddress"]}\t#{host[:vmhostname] || host.name}\n"
+      end
+      @hosts.each do |host|
+        set_etc_hosts(host, etc_hosts)
+      end
+
     end
 
     def cleanup
--- a/lib/beaker/hypervisor/docker.rb
+++ b/lib/beaker/hypervisor/docker.rb
@@ -66,15 +66,18 @@ module Beaker
         host['docker_image'] = image
       end
 
-      ## Set /etc/hosts :
+      hack_docker_etc_hosts @hosts, @options
+
+    end
+
+    def hack_docker_etc_hosts hosts, opts
       etc_hosts = "127.0.0.1\tlocalhost localhost.localdomain\n"
-      @hosts.each do |host|
-        etc_hosts += "#{host['docker_container'].json["NetworkSettings"]["IPAddress"]}\t#{host[:vmhostname] || host.name}\n"
+      hosts.each do |host|
+        etc_hosts += "#{host['docker_container'].json["NetworkSettings"]["IPAddress"]}\t#{host.name}\n"
       end
-      @hosts.each do |host|
+      hosts.each do |host|
         set_etc_hosts(host, etc_hosts)
       end
-
     end
 
     def cleanup
--- a/lib/beaker/hypervisor/docker.rb
+++ b/lib/beaker/hypervisor/docker.rb
@@ -64,6 +64,8 @@ module Beaker
         @logger.debug("node available as  ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@#{ip} -p #{port}")
         host['docker_container'] = container
         host['docker_image'] = image
+        host['etc_host_ip'] = container.json["NetworkSettings"]["IPAddress"].to_s
+
       end
 
       hack_etc_hosts @hosts, @options
--- a/spec/beaker/hypervisor/docker_spec.rb
+++ b/spec/beaker/hypervisor/docker_spec.rb
@@ -198,6 +198,15 @@ module Beaker
 
       end
 
+      it "should generate a new /etc/hosts file referencing each host" do
+        ENV['DOCKER_HOST'] = nil
+        docker.provision
+        hosts.each do |host|
+          expect( docker ).to receive( :set_etc_hosts ).with( host, "127.0.0.1\tlocalhost localhost.localdomain\n192.0.2.1\tvm1\n192.0.2.1\tvm2\n192.0.2.1\tvm3\n" ).once
+        end
+        docker.hack_docker_etc_hosts( hosts, options )
+      end
+
       it 'should record the image and container for later' do
         docker.provision
 
--- a/spec/beaker/hypervisor/docker_spec.rb
+++ b/spec/beaker/hypervisor/docker_spec.rb
@@ -10,7 +10,7 @@ end
 
 module Beaker
   describe Docker do
-    let(:hosts) { make_hosts(:hypervisor => 'docker') }
+    let(:hosts) { make_hosts }
 
     let(:logger) do
       logger = double('logger')
--- a/spec/beaker/hypervisor/docker_spec.rb
+++ b/spec/beaker/hypervisor/docker_spec.rb
@@ -10,7 +10,7 @@ end
 
 module Beaker
   describe Docker do
-    let(:hosts) { make_hosts }
+    let(:hosts) { make_hosts(:hypervisor => 'docker') }
 
     let(:logger) do
       logger = double('logger')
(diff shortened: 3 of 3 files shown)


madAndroid:
I've amended the approach somewhat, to create a new docker specific method to create the `/etc/hosts` file, since that can then be tested in the spec file.

lib/beaker/hypervisor/docker.rb
```diff
@@ -65,6 +65,19 @@ def provision
...
       end
+
+      hack_docker_etc_hosts @hosts, @options
+
+    end
+
+    def hack_docker_etc_hosts hosts, opts
+      etc_hosts = "127.0.0.1\tlocalhost localhost.localdomain\n"
+      hosts.each do |host|
+        etc_hosts += "#{host['docker_container'].json["NetworkSettings"]["IPAddress"]}\t#{host.name}\n"
```
anodelman: Looking at https://github.com/madAndroid/beaker/blob/hack_etc_hosts_docker/lib/beaker/hypervisor/docker.rb#L56 it looks like host['ip'] is being set per-host already, so you should be able to use the pre-existing hack_etc_host method.

madAndroid: Unfortunately not: this is always set to `0.0.0.0` if the `'DOCKER_HOST'` env_var isn't set - it get's set to the host machine's IP here => https://github.com/madAndroid/beaker/blob/hack_etc_hosts_docker/lib/beaker/hypervisor/docker.rb#L49 .. and AIUI, it get's set to the DOCKER_HOST's IP when that is set. This is to allow Beaker to connect via the port forward that Docker creates.

madAndroid: From my testing, when you use the `hack_etc_hosts

electrical: example entry of a docker instance:

```
# cat /etc/hosts
172.17.0.15     centos-6-x64
127.0.0.1       localhost
::1     localhost ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
```

This is with a remote docker server btw.
The hosts file looks good to me :-)

madAndroid: hmm.. yes, but the `hack_etc_hosts` method isn't present in the `provision` method, is it? :)
or have you added it? When I add `hack_etc_hosts` in the `provision` method in https://github.com/puppetlabs/beaker/blob/master/lib/beaker/hypervisor/docker.rb#L26, it creates a hostfile with `0.0.0.0` listed for all hosts

madAndroid: I'll gather some data, attempt using `hack_etc_hosts` again, and post back here :)

electrical:
@madAndroid this is with the latest master branch of beaker/beaker-rspec btw :-)

madAndroid:
When I have added `hack_etc_hosts` in docker's provision method:
```
node available as  ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@0.0.0.0 -p 49218

host1 18:05:42$ echo '127.0.0.1 localhost localhost.localdomain
0.0.0.0 host1
0.0.0.0 host2
0.0.0.0 host3
' > /etc/hosts
```

anodelman:
Hm, so your ip's are being incorrectly configured.  May need to fix the set ip line and then the rest will work.

madAndroid:
It's getting that from the provision method though, through a call to the docker API.. when I query the same fields directly using `docker inspect`, I get the same thing.. 
compare:
```
$ docker inspect [hash]
<snip>
    "NetworkSettings": {
        "Bridge": "docker0",
        "Gateway": "172.17.42.1",
        "IPAddress": "172.17.0.83",
        "IPPrefixLen": 16,
        "MacAddress": "02:42:ac:11:00:53",
        "PortMapping": null,
        "Ports": {
            "22/tcp": [
                {
                    "HostIp": "0.0.0.0",
                    "HostPort": "49217"
                }
            ]
        }
    },
<snip>
```
with: (taken from https://github.com/puppetlabs/beaker/blob/master/lib/beaker/hypervisor/docker.rb#L49)
```
        # Find out where the ssh port is from the container
        if ENV['DOCKER_HOST']
          ip = URI.parse(ENV['DOCKER_HOST']).host
          @logger.info("Using docker server at #{ip}")
        else
          ip = container.json["NetworkSettings"]["Ports"]["22/tcp"][0]["HostIp"]
        end
        port = container.json["NetworkSettings"]["Ports"]["22/tcp"][0]["HostPort"]

        forward_ssh_agent = @options[:forward_ssh_agent] || false

        # Update host metadata
        host['ip']  = ip
        host['port'] = port
        host['ssh']  = {
          :password => root_password,
          :port => port,
          :forward_agent => forward_ssh_agent,
        }
```
`host['ip']` is being set to the host machine's address ... hence the reason I sent the PR :)
The PR sets the hosts['ip'] for `set_etc_hosts` to:
```
.json["NetworkSettings"]["IPAddress"]
```

madAndroid:
Let me try rather with the latest HEAD of master, and try using `hack_etc_hosts` in my `spec_helper_acceptance.rb` rather, to see if that makes a difference ..

madAndroid:
Nope, same story .. without any call to `hack_etc_hosts` (anywhere), I get a host file that only references the localhost, none of the other nodes in the nodeset .. when I add `hack_etc_hosts` to the provision method in `docker.rb`, I get a hostfile that contains all nodes, but with their IPs set to `0.0.0.0` ... when I use my branch, I get hostfiles with the correct IPs:
```
node available as  ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@0.0.0.0 -p 49227
host1 18:32:19$ echo '127.0.0.1 localhost localhost.localdomain
172.17.0.91     host1
172.17.0.92     host2
172.17.0.93     host3
' > /etc/hosts
```

madAndroid:
@electrical - without `hack_etc_hosts` being called (which is the default for the docker hypervisor, AFAIK), the hostfile get's provisioned the same as the example you posted, but if you have multiple machines in the nodeset, each machine doesn't get provisioned with the other machine's IPs in `/etc/hosts`, unless `hack_etc_hosts` is called somewhere. The example you posted - is that from a multiple machine nodeset? It appears to only have one host listed, so I'm assuming not :)

anodelman:
Perhaps this line 

ip = container.json["NetworkSettings"]["Ports"]["22/tcp"][0]["HostIp"]

need to be updated to handle the case of being handed an ip of 0.0.0.0 - and then use 

host['docker_container'].json["NetworkSettings"]["IPAddress"]

?  

@electrical I don't have much insight here as to how docker handles ip address assignment, but does that seems like a reasonable approach?

madAndroid:
> Perhaps this line
> 
> ip = container.json["NetworkSettings"]["Ports"]["22/tcp"][0]["HostIp"]
> 
> need to be updated to handle the case of being handed an ip of 0.0.0.0 - and then use
> 
> host['docker_container'].json["NetworkSettings"]["IPAddress"]

I did try something along those lines when I was testing; the problem is that Beaker then attempts to connect on the nodes's IP instead of the `0.0.0.0:[forwarded_port]` combination, and it fails right at the beginning of the run (due to the incorrect port being used to connect, AFAICT). I'm not sure of the best option here, other then adding the docker specific method.. perhaps @electrical has a better approach in mind

electrical:
I'll give this some more thought. I must admit i never used multiple machine situations ( always been a single node ) so i can't give any insight on that. but that the hosts file is so different is spooky.

@madAndroid could you try something with a single node and see if it has the same issue?

madAndroid:
@electrical: Sure thing :)

This is the hosts file from a test run of the `puppetlabs-apache` acceptance tests (with no `hack_etc_hosts` present in [docker.rb](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/hypervisor/docker.rb) - which is the default) :
```
[root@centos-7-x64 ~]# cat /etc/hosts
172.17.0.95     centos-7-x64
127.0.0.1       localhost
::1     localhost ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
```

With `hack_etc_hosts` present (Beaker 2.6 installed using bundler, and then directly amended to add `hack_etc_hosts` in [docker.rb](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/hypervisor/docker.rb))
```
[root@centos-7-x64 ~]# cat /etc/hosts
127.0.0.1       localhost localhost.localdomain
0.0.0.0 centos-7-x64
```

With the Gemfile amended to use the `hack_etc_hosts_docker` branch of my fork:
```
[root@centos-7-x64 ~]# cat /etc/hosts
127.0.0.1       localhost localhost.localdomain
172.17.0.97     centos-7-x64
```

So the hosts file gets created normally when `hack_etc_hosts` is absent (the default, since `hack_etc_hosts` has never been included in beaker for the docker provider), which is how you're seeing your nodes created when using the HEAD of origin/master for your testing. When I've added a call to `hack_etc_hosts`, the host file doesn't get created with the correct IPs, which is what lead me to create the new method in [docker.rb](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/hypervisor/docker.rb)

The use case I'm testing with uses multiple nodes to test in both vagrant on OSX machines, and docker when run from Jenkins on CentOS7 - I noticed that the tests I added would work fine in vagrant, but fail in docker, due to the discrepancies in the hosts files... the only way I could get them to pass was to implement the new method which uses `set_etc_hosts` with the correct IPs collected from the entry in the `host['docker_container']` key.

electrical:
@madAndroid thank you very much for the detailed info.
So by default the docker stuff creates the `/etc/hosts` file correctly on its own, but as soon as it requires multiple hosts the `hack_etc_hosts` file is required.
With the current implementation of the ip address stuff it fails because we use either the forwarded port or `DOCKER_HOST` env var.

We could change the hosts file content to use the actual docker container IP when using the `hack_etc_hosts` action.
But that idea might fail if someone would use docker + other hypervisor. ( will that ever be used ? )

electrical:
Several more things to think about.
With the current situation:

When using a single docker host and without the etc hack it lets docker take care of the hosts file to ensure it has the IP of the container.
When using a single docker instance without the etc hack and `DOCKER_HOST` env var is set it still sets the right IP in the hosts file.

As soon as we enable the etc hack it gets a bit more difficult because we still have to make sure we actually use the internal IP's of the containers and not the `DOCKER_HOST` env var.
It gets even worse if we would use Swarm ( Docker's cluster software ) with multiple slaves ;-)

madAndroid:
> We could change the hosts file content to use the actual docker container IP when using the hack_etc_hosts action.

Yes, that's certainly a possibility :) From what I can see, there are other methods in [host_prebuilt_steps.rb](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host_prebuilt_steps.rb) which vary operations based on `host['platform']`, but nothing that varies on `host['hypervisor']` - I'm not sure if that's the correct direction to head in, but it's worth investigating IMO.

> But that idea might fail if someone would use docker + other hypervisor. ( will that ever be used ? )

I can see that being a use case, someone could attempt to take that path: I'll test with that as an option on a linux machine.

> As soon as we enable the etc hack it gets a bit more difficult because we still have to make sure we actually use the internal IP's of the containers and not the DOCKER_HOST env var.
> It gets even worse if we would use Swarm ( Docker's cluster software ) with multiple slaves ;-)

Very valid points :) I'll amend the current PR to use `hack_etc_hosts`, to ensure we keep the code DRY, perhaps that's the better way to go.

madAndroid:
Those changes appear to work in the spec tests, will be testing them at length in the morning :)

lib/beaker/host_prebuilt_steps.rb
```diff
@@ -321,7 +321,11 @@ def copy_ssh_to_root host, opts
     def hack_etc_hosts hosts, opts
       etc_hosts = "127.0.0.1\tlocalhost localhost.localdomain\n"
       hosts.each do |host|
-        etc_hosts += "#{host['ip'].to_s}\t#{host[:vmhostname] || host.name}\n"
+        if host['hypervisor'] == 'docker'
+          etc_hosts += "#{host['docker_container'].json["NetworkSettings"]["IPAddress"]}\t#{host.name}\n"
```
madAndroid: My understanding is that this will always be set to the actual containers IP, and never to the `DOCKER_HOST` IP if that env_var is set.

electrical:
@madAndroid that should work yeah. im not sure though about the hypervisor specific code in a generic helper function.
@anodelman what is your view on this?

anodelman:
Yeah, docker specific code should not appear in host_prebuilt_steps - that breaks our object model.

Maybe update hack_etc_hosts to accept an additional param...

madAndroid:
> Yeah, docker specific code should not appear in host_prebuilt_steps - that breaks our object model.

Fair enough :)

> Maybe update hack_etc_hosts to accept an additional param...

After some further testing, I've had a look at adding another `host` attribute rather: `etc_host_ip` - this opens up the possibility of using this for other providers that may differentiate between the SUTs ips/hostnames and the connection mechanism. For any providers that don't have this set, it falls back to the host['ip']. Does this sound like a better solution than provider specific code in `host_prebuilt_steps`?

anodelman:
Maybe it should be called 'vm_ip'?  I'm not sure if the name is totally clear as to what the difference between the two ip addresses associated with a host is.

Also, you should associate your commits with a beaker ticket number (BKR-xxx).

Also, please squash your commits - it will make reading beaker git log easier once this gets merged.

Thanks for sticking with this!

madAndroid:
> Also, you should associate your commits with a beaker ticket number (BKR-xxx).
> Also, please squash your commits - it will make reading beaker git log easier once this gets merged.

I've created a new PR which references the ticket at #770 - closing this PR in favour of that one.

> Thanks for sticking with this!

Thank you for all your help! :)

PR #743, opened 2015-03-12
(BKR-81) start using new public beaker jira project

anodelman:
- update contributor and readme doc to point to correct ticket tracker
  information


anodelman:
Failures definitely unrelated to patch, 'cause the patch only updates documentation.

PR #748, opened 2015-03-15
[BKR-181] docker hypervisor changes for centos/redhat images

n1tr0g:
Some minor changes to the centos/redhat docker images, so we can use them for testing. 

- by default we need to install the tar package on centos/redhat images, so we can use commands like 'puppet module install MODULENAME'

- Also we shouldn't restart the ssh service as defined in host_prebuilt_steps.rb if the hypervisor is docker and the image is centos/redhat based, because this images don't have the necessary packages to support that.

### DIFF
--- a/lib/beaker/host_prebuilt_steps.rb
+++ b/lib/beaker/host_prebuilt_steps.rb
@@ -508,7 +508,7 @@ module Beaker
           host.exec(Command.new("service ssh restart"))
         when /el-|centos|fedora|redhat|oracle|scientific|eos/
           host.exec(Command.new("echo '\nPermitUserEnvironment yes' >> /etc/ssh/sshd_config"))
-          host.exec(Command.new("/sbin/service sshd restart"))
+          host.exec(Command.new("/sbin/service sshd restart")) unless host['hypervisor'] == 'docker'
         when /sles/
           host.exec(Command.new("echo '\nPermitUserEnvironment yes' >> /etc/ssh/sshd_config"))
           host.exec(Command.new("rcsshd restart"))
--- a/lib/beaker/hypervisor/docker.rb
+++ b/lib/beaker/hypervisor/docker.rb
@@ -135,7 +135,7 @@ module Beaker
       when /^el-/, /centos/, /fedora/, /redhat/, /eos/
         dockerfile += <<-EOF
           RUN yum clean all
-          RUN yum install -y sudo openssh-server openssh-clients #{Beaker::HostPrebuiltSteps::UNIX_PACKAGES.join(' ')}
+          RUN yum install -y sudo openssh-server openssh-clients tar #{Beaker::HostPrebuiltSteps::UNIX_PACKAGES.join(' ')}
           RUN ssh-keygen -t rsa -f /etc/ssh/ssh_host_rsa_key
           RUN ssh-keygen -t dsa -f /etc/ssh/ssh_host_dsa_key
         EOF


lib/beaker/host_prebuilt_steps.rb
```diff
@@ -502,7 +502,7 @@ def set_env host, opts
           host.exec(Command.new("service ssh restart"))
         when /el-|centos|fedora|redhat|oracle|scientific|eos/
           host.exec(Command.new("echo '\nPermitUserEnvironment yes' >> /etc/ssh/sshd_config"))
-          host.exec(Command.new("/sbin/service sshd restart"))
+          host.exec(Command.new("/sbin/service sshd restart")) unless host['hypervisor'] == 'docker'
```
anodelman: If you are making hypervisor specific alterations to the workflow then it should occur in the docker.rb file by overloading the configuration/validation steps that occur post-provisioning.

kevpl:
needs contributor update: in particular, needs the issue prepended to the commit messages

lib/beaker/host_prebuilt_steps.rb
n1tr0g: easy I will alter the code to reflect that. 

So basically you want me to overload def configure in the docker.rb, so it can use a different set_env function?

n1tr0g:
@kevpl not sure what you exactly mean.

lib/beaker/host_prebuilt_steps.rb
anodelman: Yup, that's the right way to use the beaker object architecture.

anodelman:
Our Contributor doc describes that beaker commit messages need titles of the form:
```
(BKR-xxx) description of the jira bug
```

Your commit messages don't meet this standard.

lib/beaker/host_prebuilt_steps.rb
n1tr0g: cheers I will add that, but it will make the code really messy :)

n1tr0g:
ok, so for every PR you first need to create a ticket ? I can do that.

btw guys please don't forget that third party contributors are using their free time doing that, so words like "thank you", "please", etc go a long way :) 

I will update the PR tonight, after work in my free time :)

n1tr0g:
I am unable to create a Beaker ticket with my current permissions, however I am able to do that for the other projects.

screen shot 2015-03-18 at 8 45 13 pm

List of other projects:
screen shot 2015-03-18 at 8 50 25 pm

n1tr0g:
tnx ... looks like someone fixed my privileges.

anodelman:
Yup, the Jira permissions were all busted but should now all be correct.  Sorry for the confusion.

Still need updates based upon previous review comments.

anodelman:
Previous review comment still applies:

```
If you are making hypervisor specific alterations to the workflow then it should occur in the docker.rb file by overloading the configuration/validation steps that occur post-provisioning.
```

n1tr0g:
i will commit my change today/tomorrow, depends how much time I have to test it.

electrical:
I'm still :-1: on this change as i use docker differently and rely on sshd being reloaded. ( i run sshd as an actual service inside the docker container )

n1tr0g:
This should only apply on centos/redhat containers, because as it is now you can't use it with the default centos image. Do you have a better suggestion ?

electrical:
the change would still break things for me also on my centos container.
In my opinion it should be a configurable setting to restart sshd or not, and not just disable it by default when using docker.

n1tr0g:
Can you show me an example how this change would break your centos docker tests ? If you look at mine https://github.com/n1tr0g/golja-influxdb docker tests, they only work with the PR applied (both centos66 and centos7).

CentOS 7 container has a pre-installed fakesystemd, which does not implement the "service" and "systemctl" commands, so that is root of all problems :)

Btw I agree that this could be configurable, but the question is how @anodelman would like to do it.

electrical:
@n1tr0g by default docker starts the sshd daemon as only process ( main process ) and everything else spawns of there.
In my case i have /sbin/init as my start process which handles all other services ( sshd ) and lets me test service management. ( start/stop of a service ) which otherwise fails in most cases.
The restart of sshd allows some paths to be set correctly which if not set may cause weirdness.

n1tr0g:
Hmmm if that the case and from the sound of it your use case is quiet unique, why don't you just add an extra docker_image_commands where you restart the service as I did in https://github.com/n1tr0g/golja-influxdb/blob/master/spec/acceptance/nodesets/centos-7.yml, to install tar, so puppet module install works ?

electrical:
I already use special stuff like https://github.com/elastic/puppet-elasticsearch/blob/master/spec/acceptance/nodesets/centos-6-x64.yml ;-)

n1tr0g:
Hmmm I am missing something. If we know that centos7 images doesn't work with the current code why don't you just add the extra docker_image_commands: service restart sshd so we can fix both use cases ?

electrical:
The `docker_image_commands` are executed way before the ssh env setup is done, so that won't work either...

n1tr0g:
HeHe ops :) Maybe puppetlabs just need to release some official Docker images and everyone would be happy :)

anodelman:
I don't think that the right solution is to make docker not do the ssh restart - but have the option be configurable.

n1tr0g:
No problem. Where would you like to have this option added ? Not sure where to put it :)

anodelman:
Do you just want to skip the ssh restart, or setting env at all for docker?  There is already a 'set_env' variable defined that default to true.  If you set to false it will skip all the environment hacking during host setup.

anodelman:
Last comment is still current.

pauldavidgilligan-msm:
I tried the above for centos 7 but beaker now gives:

```
/home/dev-ops/.rvm/gems/ruby-2.1.5/gems/beaker-2.9.0/lib/beaker/host.rb:412:in `exec': Host 'centos-7-puppet' exited with 127 running: (Beaker::Host::CommandFailure)
```

a crude work around about to happen, any updates on this issue pls ? we are using centos6 currently but sure to move to 7 at some point.

n1tr0g:
Sorry guys I am back :)  @anodelman I am not sure what you mean with set_env is already defined ? Is this an setting flag ? If so could you point me to the documentation ?

In general I think the best solution would be to just provide puppetlabs beaker base docker images, similar as they did with the vagrant images in the past.

anodelman:
See the ability to switch on/off environment variable setting here:

https://github.com/puppetlabs/beaker/blob/master/lib/beaker/options/presets.rb#L150

anodelman:
@pauldavidgilligan-msm If you aren't hitting that code chunk right you may not have the right platform name associated with each of your hosts.

anodelman:
There's been no update to the PR since march 15 and the review comments are still current.  Will close this out next week if there has been no update.

anodelman:
Closing due to inactivity.

PR #749, opened 2015-03-16
(BKR-79) added deprecated warning to host defaults

kevpl:


lib/beaker/host.rb
```diff
@@ -59,6 +59,26 @@ def initialize name, options
...
 
+    # Builds a deprecated keys array, for checking to see if a key is deprecated.
+    # The recommended check after using this method is +result.includes?(key)+
+    #
+    # @note an unsupported host type (meaning it has no _aio_defaults_) will return
+    #   an empty hash
+    #
+    # @return [Array<Symbol>] An array of keys that are deprecated for a host
+    def build_deprecated_keys()
+      deprecated_keys_hash = self.class.send "foss_defaults".to_sym
```
anodelman: Let's move this into the begin block, in case someone out there is using a custom host type.

PR #750, opened 2015-03-18
(PE-8260) add answer for new NC migrate question

MosesMendoza:
This commit adds a 38-specific answer file to beaker, which introduces an
answer for whether to stop the upgrade and proceed to do the NC database
migration. It defaults to 'n' to enable upgrades to continue.

Signed-off-by: Moses Mendoza <moses@puppetlabs.com>

### DIFF
--- a/lib/beaker/answers/version38.rb
+++ b/lib/beaker/answers/version38.rb
@@ -1,0 +1,24 @@
+require 'beaker/answers/version34'
+
+module Beaker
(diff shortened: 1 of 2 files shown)


MosesMendoza:
@anodelman @kevpl can you take a look at this? I did my best to follow the pattern I've seen but I'm not sure this is right? This is related to https://github.com/puppetlabs/enterprise-dist/pull/2382. They should be merged at the same time.

PR #760, opened 2015-03-23
Add a libvirt hypervisor

nibalizer:
This commit adds a libvirt hypervisor that shells out to virsh.
This does not use the libvirt ruby gem. The libvirt ruby gem is
not well maintained and gets us into dependency problems with
nokogiri fast.

Dependency issues with nokogiri is what motivated me to write this
instead of using the existing libvirt-vagrant hypervisor

I'll likely write a follow on blog post about how I build my qcow
images and use this hypervisor.

### DIFF
--- a/lib/beaker/hypervisor/libvirt.rb
+++ b/lib/beaker/hypervisor/libvirt.rb
@@ -1,0 +1,152 @@
+# TODO: rework qcow handling to do this:
+# virsh vol-create-as default guest 2048MB
+# virsh vol-upload --pool default guest file.qcow2
+# virsh vol-delete --pool default guest
+module Beaker
+  class Libvirt < Beaker::Hypervisor
+
+    def initialize(hosts, options)
+      require 'tempfile'
+      @options = options
+      @logger = options[:logger]
+      @hosts = hosts
+      @logger.notify "Connecting to libvirt"
+
+    end
+
+    def provision
+      @logger.notify "Provisioning libvirt"
+      #conn = ::Libvirt::open('qemu:///system')
+
+      @hosts.each do |host|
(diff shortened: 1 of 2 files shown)


lib/beaker/hypervisor/libvirt.rb
```diff
@@ -0,0 +1,152 @@
...
+    <console type='pty'>
+      <target type='serial' port='0'/>
+    </console>
+  </devices>
+</domain>
+EOF
+        @logger.notify "Creating transient domain ruby-libvirt-tester"
+        file.write(new_dom_xml)
+        file.close
+        `virsh define #{file.path}`
```
electrical: Best to use something like popen. see https://github.com/puppetlabs/beaker/blob/master/lib/beaker/hypervisor/vagrant.rb#L67 for example.
Perhaps it could be created into a generic helper / util function.

anodelman:
We do not merge new hypervisor support unless we have a dedicated community member to monitor patches and testing - as we do not use libvrt internally it will not be exercised in our testing.

Otherwise, you can use --helper to add libraries to beaker at runtime.

nibalizer:
I'm happy with either approach. I'll take a stab at putting this in a library.

nibalizer:
@anodelman where is the canonical list of community supported hypervisors? I'd like to add mine there when it's published.

anodelman:
We don't have a doc or that yet, so there's something for me to take care of.  :)

So far it's just docker, supported by @electrical.

nibalizer:
And also fog from @petems and https://github.com/nibalizer/beaker-libvirt from me so the list is growing :)

PR #761, opened 2015-03-24
(MAINT) Show error from docker

petems:
As @richardc pointed out, the commit 25909d02608f9ec6fb69be972750d39ed08f2c41 actually removes being able to see the error raised.

This fixes that, so you can actually see what the issue is.


madAndroid:
:+1: this will come in handy :)

electrical:
LGTM

PR #762, opened 2015-03-27
(BKR-161) Beaker tagging improvements

sschneid:
- Use JOB_NAME for 'project' tag default if it exists
- Set 'beaker_version' tag in vmpooler hypervisor

### DIFF
--- a/lib/beaker/options/presets.rb
+++ b/lib/beaker/options/presets.rb
@@ -11,9 +11,12 @@ module Beaker
       # us to define multiple environment variables for the same
       # configuration value. They are checked in the order they are arrayed
       # so that preferred and "fallback" values work as expected.
+      #
+      # 'JOB_NAME' and 'BUILD_URL' envs are supplied by Jenkins
+      # https://wiki.jenkins-ci.org/display/JENKINS/Building+a+software+project
       ENVIRONMENT_SPEC = {
         :home                 => 'HOME',
-        :project              => ['BEAKER_PROJECT', 'BEAKER_project'],
+        :project              => ['BEAKER_PROJECT', 'BEAKER_project', 'JOB_NAME'],
         :department           => ['BEAKER_DEPARTMENT', 'BEAKER_department'],
         :jenkins_build_url    => ['BEAKER_BUILD_URL', 'BUILD_URL'],
         :created_by           => ['BEAKER_CREATED_BY'],
--- a/lib/beaker/hypervisor/vmpooler.rb
+++ b/lib/beaker/hypervisor/vmpooler.rb
@@ -111,6 +111,7 @@ module Beaker
       @logger.notify 'Tagging vmpooler VMs'
 
       tags = {
+        'beaker_version' => Beaker::Version::STRING,
         'jenkins_build_url' => @options[:jenkins_build_url],
         'department' => @options[:department],
         'project' => @options[:project],


rji:
:+1:

sschneid:
Tagging result as illustrated by the Beaker acceptance run:

````json
{
  "ok": true,
  "w5p94l212w5fphs": {
    "template": "centos-7-x86_64",
    "lifetime": 12,
    "running": 0.01,
    "tags": {
      "jenkins_build_url": "http://jenkins-beaker.delivery.puppetlabs.net/job/qe_beaker_intn-sys_beaker-accceptance-base-vpool/",
      "department": "eso-dept",
      "beaker_version": "2.8.0",
      "created_by": "jenkins",
      "project": "Beaker Smoketest - PE Extended Platforms Smoke Test"
    },
    "domain": "delivery.puppetlabs.net"
  }
}
````

**Comment by sschneid** [comment]
This shouldn't be merged until we have a discussion/decision about whether to overload 'project'.  (cc: @waynr @rji)

File: lib/beaker/options/presets.rb [hunk 1]
```diff
@@ -110,7 +110,7 @@ def env_vars
       def presets
         h = Beaker::Options::OptionsHash.new
         h.merge({
-          :project                => 'Beaker',
+          :project                => ENV['JOB_NAME'] || 'Beaker',
```
waynr: Pulling environment variables in here may be somewhat confusing for people reading the code since there is another place where we pull in environment variables:
https://github.com/sschneid/beaker/blob/bkr-161_tagging/lib/beaker/options/presets.rb#L16

If you pull in JOB_NAME there you can be more explicit about which variables take precedence. If I understand correctly, setting it here will definitely allow it to be overridden by BEAKER_project or BEAKER_PROJECT.

However, I am not generally familiar with recent Beaker development goals and I see that other variables are pulled in within this code block so *shrug*.

sschneid: Yep... this particular block is setting defaults, the block you referenced then overrides the defaults with the 'BEAKER_' ENV params.

anodelman: @waynr is correct - we try and pull in all the env vars in one chunk at one go so that it is easier to track what variables are supported.

anodelman: Also, beaker env vars should be of the form BEAKER_*** - just for consistency.

waynr: @anodelman JOB_NAME is set by Jenkins itself (it'd probably be helpful to make note of this in a comment)

lib/beaker/options/presets.rb
```diff
@@ -13,7 +13,7 @@ class Presets
       # so that preferred and "fallback" values work as expected.
       ENVIRONMENT_SPEC = {
         :home                 => 'HOME',
-        :project              => ['BEAKER_PROJECT', 'BEAKER_project'],
+        :project              => ['BEAKER_PROJECT', 'BEAKER_project', 'JOB_NAME'],
```
anodelman: Is 'JOB_NAME' a requirement here?  Beaker env vars are of the form BEAKER_*** (unless they are historic and pending deprecation).

sschneid: Like BUILD_URL a few lines below, this comes from Jenkins and is used to set the project name if BEAKER_PROJECT (or _project) is unset.

sschneid:

PR #615, opened 2015-01-13
(QENG-1663) Add FreeBSD support

petems:
Full FreeBSD 9 and 10 support

I've got it working with https://github.com/puppetlabs/puppetlabs-inifile

Check it out here: https://github.com/petems/puppetlabs-inifile/tree/freebsd_beaker_tests

Full test suite working when FreeBSD specific changes made to the acceptance tests :+1:

PR #764, opened 2015-03-30
(BKR-113) Add FreeBSD host class

petems:
Breaking up #615 into easier to merge chunks :+1: 

* Main change is location of Puppet data under `/usr/local/etc` rather than `/etc/`

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -39,6 +39,8 @@ module Beaker
         Aix::Host.new name, options
       when /osx/
         Mac::Host.new name, options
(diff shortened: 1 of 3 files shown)


anodelman:
This should have the platform.rb changes to go along with it - otherwise you'll never be able to create a host of this type.

anodelman:
Thanks for breaking up the PRs!  This is a lot easier to review!

petems:
@anodelman Added! :+1:

anodelman:
ok to test

PR #765, opened 2015-03-30
(BKR-113) Adds FreeBSD Vagrantfile logic

petems:
With Vagrant we use rsync for folder sync, as nfs has a character restriction of  88 characters. So depending on the module name and where it's being stored, it can error out with:

```
==> default: Mounting NFS shared folders...
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

mount -t nfs '10.0.1.1:/Users/petersouter/projects/reallylongpathnameover88characterssothatmountfswillfail12345678910111213141516' '/vagrant'

Stdout from the command:

Stderr from the command:

mount_nfs: 10.0.1.1:/Users/petersouter/projects/reallylongpathnameover88characterssothatmountfswillfail12345678910111213141516: File name too long
```

Further reading if interested:
http://www.secnetix.de/olli/FreeBSD/mnamelen.hawk
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=167105

### DIFF
--- a/lib/beaker/hypervisor/vagrant.rb
+++ b/lib/beaker/hypervisor/vagrant.rb
@@ -46,6 +46,25 @@ module Beaker
           v_file << "    v.vm.synced_folder '.', '/vagrant', :nfs => true\n"
         end
 
+        if /freebsd/i.match(host['platform'])
+          v_file << "    v.ssh.shell = 'sh'\n"
+          v_file << "    v.vm.guest = :freebsd\n"
+
+          # FreeBSD NFS has a character restriction of 88 characters
+          # So you can enable it but if your module has a long name it probably won't work...
+          # So to keep things simple let's rsync by default!
+          #
+          # Further reading if interested:
+          # http://www.secnetix.de/olli/FreeBSD/mnamelen.hawk
+          # https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=167105
+          #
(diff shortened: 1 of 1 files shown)


anodelman:
We'll look at this post https://github.com/puppetlabs/beaker/pull/764 merge.

lib/beaker/hypervisor/vagrant.rb
```diff
@@ -46,6 +46,20 @@ def make_vfile hosts, options = {}
...
         end
 
+        if /freebsd/i.match(host['platform'])
+          v_file << "    v.ssh.shell = 'sh'\n"
+          v_file << "    v.vm.guest = :freebsd\n"
+
+          # FreeBSD NFS has a character restriction of 88 characters
+          # So you can enable it but if your module has a long name it probably won't work...
+          # Further links and reading in git commit message
+          if host['vagrant_freebsd_nfs'].nil?
```
anodelman: Might as well put the notes and links right in the code.  :)

petems: Hmm, honestly I prefer having notes in git commits, as git commits are more permanent and searchable than code, but I can do both :+1:

petems: Fixed! :+1:

PR #768, opened 2015-03-30
(QENG-2083) share out beaker's history file generation tool

anodelman:
- remove the history file used exclusively by beaker in favor of the
  new, general use tool


anodelman:
Failure unrelated to patch.

PR #769, opened 2015-03-31
(maint) Allow installing 32-bit Puppet-Agent on Windows

Iristyle:
- Previously, install_utils was modified to be able to install PE
   based on a configuration switch called `install_32` in
   6f8deb7dbc7bc75b8624df754c261f4d2b58b999

   This allows a 64-bit Windows OS to ask to install a 32-bit Ruby /
   Puppet version.  This is important given 64-bit support wasn't
   added to Puppet or PE until 3.7.

   When installing the puppet-agent through install_puppetagent_dev_repo
   helper, allow for selecting the build in a similar fashion.

### DIFF
--- a/lib/beaker/dsl/install_utils/puppet_utils.rb
+++ b/lib/beaker/dsl/install_utils/puppet_utils.rb
@@ -347,7 +347,8 @@ module Beaker
           # - we do not have install_32 set on host
           # - we do not have install_32 set globally
           version = opts[:version]
-          if !(version_is_less(version, '3.7')) and host.is_x86_64? and not host['install_32'] and not opts['install_32']
+          is_config_32 = true == (host['ruby_arch'] == 'x86') || host['install_32'] || opts['install_32']
+          if !(version_is_less(version, '3.7')) && host.is_x86_64? && !is_config_32
             host['dist'] = "puppet-#{version}-x64"
           else
             host['dist'] = "puppet-#{version}"
@@ -693,7 +694,8 @@ module Beaker
           when /^windows$/
             release_path << 'windows'
             onhost_copy_base = '`cygpath -smF 35`/'
-            should_install_64bit = host.is_x86_64? && !host['install_32'] && !opts['install_32']
+            is_config_32 = true == (host['ruby_arch'] == 'x86') || host['install_32'] || opts['install_32']
+            should_install_64bit = host.is_x86_64? && !is_config_32
             # only install 64bit builds if
             # - we do not have install_32 set on host
             # - we do not have install_32 set globally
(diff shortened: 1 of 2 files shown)


lib/beaker/dsl/install_utils/puppet_utils.rb
```diff
@@ -693,7 +693,11 @@ def install_puppetagent_dev_repo( host, opts )
           when /^windows$/
             release_path << 'windows'
             onhost_copy_base = '`cygpath -smF 35`/'
-            arch_suffix = arch =~ /64/ ? '64' : '86'
+            should_install_64bit = host.is_x86_64? && !host['install_32'] && !opts['install_32']
```
Iristyle: Would prefer this setting to be `:ruby_arch`, which is what we've use in Puppet since the introduction of Ruby 2.0.0 / x64 support (which involved build pipeline changes from @rji)

See https://tickets.puppetlabs.com/browse/PUP-2388 for the history around `:ruby_arch` or the associated PR at https://github.com/puppetlabs/puppet/pull/2903

I've used `:install_32` for consistency now with what I see in https://github.com/puppetlabs/beaker/commit/6f8deb7dbc7bc75b8624df754c261f4d2b58b999 but would prefer to see this changed to `:ruby_arch` if we know we don't have downstream consumers here.

If you look at https://github.com/puppetlabs/puppet/tree/master/acceptance/config/nodes you'll see all the existing Windows node definitions use `:ruby_arch`


/cc @joshcooper @justinstoller @kevpl @anodelman

joshcooper: Can we modify beaker to accept either `:ruby_arch` or `:install_*` (in that order), and deprecate the latter?

Iristyle: That's certainly the correct approach with respect to backward compatibility.

Iristyle:
@joshcooper I added 489f008 to allow us to use `:ruby_arch`.

anodelman:
So :ruby_arch is a new addition to the host support?

Please update the yard docs for do_install to reflect this new option.

anodelman:
Request to update yard docs still current.

Iristyle:
@anodelman these settings are on the `host`, not as part of the `opt` hash passed into the method.  There are no yard docs for `install_32` anywhere that I could find FWIW.

anodelman:
Please add a ```@note``` to the yard docs for that method indicating the interaction between ruby_arch and install_32 for windows hosts.

anodelman:
@Iristyle Just blocked on last review comment.

anodelman:
@kevpl can you get this over the finish line?

kevpl:
closing in favor of https://github.com/puppetlabs/beaker/pull/807

PR #770, opened 2015-03-31
(BKR-165) hack_etc_hosts method doesn't work for Docker provider

madAndroid:
This replaces #742 - with a new branch tied to an issue in JIRA

This PR adds support for adding hosts to `/etc/hosts` on all docker instances in the nodeset; using the hack_etc_hosts method without this change results in a hosts file which contains `0.0.0.0` for all entries, since the `host['ip']` for the docker instance is set to `0.0.0.0` and the ssh port which has been exposed for the particular container. Instead, we set a new key, `host['vm_ip']` in the provision method, and set the entry in the host file to this if it's present.

### DIFF
--- a/lib/beaker/host_prebuilt_steps.rb
+++ b/lib/beaker/host_prebuilt_steps.rb
@@ -313,15 +313,18 @@ module Beaker
       end
     end
 
-    #Update /etc/hosts to make it possible for each provided host to reach each other host by name.
-    #Assumes that each provided host has host[:ip] set.
+    # Update /etc/hosts to make it possible for each provided host to reach each other host by name.
+    # Assumes that each provided host has host[:ip] set; in the instance where a provider sets
+    # host['ip'] to an address which facilitates access to the host externally, but the actual host
+    # addresses differ from this, we check first for the presence of a host['vm_ip'] key first,
+    # and use that if present.
     # @param [Host, Array<Host>] hosts An array of hosts to act upon
     # @param [Hash{Symbol=>String}] opts Options to alter execution.
     # @option opts [Beaker::Logger] :logger A {Beaker::Logger} object
     def hack_etc_hosts hosts, opts
       etc_hosts = "127.0.0.1\tlocalhost localhost.localdomain\n"
       hosts.each do |host|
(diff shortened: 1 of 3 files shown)


electrical:
Code and logic looks good. Would love to see the '.to_s' in the hack function like the ip key has.
will make some time today to test it out at work.

madAndroid:
> Would love to see the '.to_s' in the hack function like the ip key has.

Tried that, but it failed in the spec test with:
```
  1) Beaker::Vagrant can generate a new /etc/hosts file referencing each host
     Failure/Error: vagrant.hack_etc_hosts( @hosts, options )
       #<Beaker::Vagrant:0x007fbfa2032b90> received :set_etc_hosts with unexpected arguments
<--snip-->
"127.0.0.1\tlocalhost localhost.localdomain\nip.address.for.vm1\tvm1\nip.address.for.vm2\tvm2\nip.address.for.vm3\tvm3\n")
got:
<--snip-->
"127.0.0.1\tlocalhost localhost.localdomain\n\tvm1\n\tvm2\n\tvm3\n")
```
so I moved the `.to_s` into the docker provisioner at https://github.com/madAndroid/beaker/blob/BKR-165-hack_etc_hosts_docker/lib/beaker/hypervisor/docker.rb#L67 ... not sure why it failed - it appears to have nullified the variable if it's not set to string in `host['vm_ip']`

electrical:
Ah okay, good reason then :-)

electrical:
Doing a test run with the custom branch now to see if it breaks anything.

electrical:
Did some local testing with a single node def.

Before the fix:
```
172.17.0.44     centos-6-x64
127.0.0.1       localhost
::1     localhost ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
```

After the fix:
```
127.0.0.1       localhost localhost.localdomain
172.17.0.43     centos-6-x64
```

The missing IPV6 stuff is fine. IP address of the container is passed on correctly in my tests.

@anodelman Perhaps we could add the IPV6 to beaker in a separate PR or leave it out?

Either case i'm :+1:  on merging this. Thanks for the work @madAndroid !

anodelman:
ok to test

PR #771, opened 2015-03-31
(QENG-2096) Update dev repo for new AIO repo name

justinstoller:
Soon AIO builds will only be published to the repo "PC1" (not
products/devel or main as previously). This allows Beaker to install
from both new AIO repos or legacy repos.

### DIFF
--- a/lib/beaker/dsl/install_utils/puppet_utils.rb
+++ b/lib/beaker/dsl/install_utils/puppet_utils.rb
@@ -551,7 +551,7 @@ module Beaker
           platform_configs_dir = File.join(repo_configs_dir, variant)
 
           # some of the uses of dev_builds_url below can't include protocol info,
-          # pluse this opens up possibility of switching the behavior on provided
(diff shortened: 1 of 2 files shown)


justinstoller:
Recent failure caused by connectivity issues to Github when downloading lein for a couple of cells in the PDB integration test matrix (unrelated to this patch). Everything else is green.

kevpl:
It looks like the tests here don't test the functionality added.  @justinstoller, would you add a unit test or two to verify this behavior, so that when I make changes later and break this (hopefully by accident), it'll scream in my face about it?

justinstoller:
@kevpl, I added three tests, two to test the conditional logic of whether or not we're using PC1 vs main in debian, and one to test the order of our link checking for EL. I reorganized the tests so that the set up code was being shared as much as possible (without stubbing and trying to accurately unstub things)

justinstoller:
Build failed because I rebased my PR and Jenkins' Git plugin did not like that.

kevpl:

PR #775, opened 2015-04-03
(BKR-4) Beaker now dumps re-usable hosts file when SUTs are preserved

kevpl:


lib/beaker/cli.rb
```diff
@@ -126,6 +126,24 @@ def execute!
           if @network_manager
             @network_manager.cleanup
           end
+        else
+          preserved_hosts_filename = File.join(@options[:log_dated_dir], File.basename(@options[:hosts_file]))
```
anodelman: let's cut this into a new method.

kevpl:
@anodelman updated.  rebased & pulled functionality into its own method.  It should be ready for you to look at now.

kevpl:
failures unrelated.

lib/beaker/cli.rb
```diff
@@ -187,13 +220,35 @@ def print_env_vars_affecting_beaker( log_level )
...
+            has_supported_hypervisor = true
+          when /docker/
+            set_docker_warning = true
+          end
+        end
+        if has_supported_hypervisor
+          @logger.send(log_level, "\nYou can re-run commands against the already provisioned SUT(s) by following these steps:\n")
+          @logger.send(log_level, '(docker support is untested for this feature. please reference the docs for more info)') if set_docker_warning
+          @logger.send(log_level, "- change the hosts file to #{@options[:hosts_preserved_yaml_file]}")
+          @logger.send(log_level, '- use the --no-provision flag')
```
waynr: I think it would be useful to also indicate which hosts have been preserved and how to log in to them manually...if this isn't done elsewhere?

lib/beaker/cli.rb
```diff
@@ -187,13 +220,35 @@ def print_env_vars_affecting_beaker( log_level )
...
+          when /vagrant|fusion|vmpooler|vcloud/
+            has_supported_hypervisor = true
+          when /docker/
+            set_docker_warning = true
+          end
+        end
+        if has_supported_hypervisor
+          @logger.send(log_level, "\nYou can re-run commands against the already provisioned SUT(s) by following these steps:\n")
+          @logger.send(log_level, '(docker support is untested for this feature. please reference the docs for more info)') if set_docker_warning
+          @logger.send(log_level, "- change the hosts file to #{@options[:hosts_preserved_yaml_file]}")
```
anodelman: I would prefer that it would just give you the whole command line to execute instead of having the user have to correctly copy/paste bits and pieces and add the --no-provision.

anodelman:
It was weird to me that when I ran this it generated a new config file every time, even if i was already using a generated config file.  Instead of generating per preserve-hosts run, generate per preserve-hosts run without --no-provision.

kevpl:

anodelman:
I think that this is an excellent first step to better host reuseability!

PR #776, opened 2015-04-03
(BKR-70)changed up the close method so that it is more robust

erikPrime:


anodelman:

lib/beaker/ssh_connection.rb
```diff
@@ -62,8 +62,14 @@ def connect
...
-      rescue
-        @ssh.shutdown!
+      rescue => e
+        @logger.warn "Attemped ssh.close.  Caught an error: #{e.message}  Attempting ssh.shutdown!..."
+        begin
+          @ssh.shutdown!
+        rescue => e
+          @logger.warn "Attemped ssh.shutdown!.  Caught an error: #{e.message}. Giving up and destorying ssh."
+          @ssh = nil
+        end
```
joshcooper: would be good to have a test for the case when `shutdown!` raises

anodelman: I would prefer to see catching a specific error instead of the general 'rescue' - we are moving away from empty rescues to explicit rescue conditions.

jeffmccune: Please also check the caught error and make sure it's the one we're expecting.  If not, please re-raise the error.

jeffmccune:
This pull request blocks the puppet-server acceptance test [PR 464](https://github.com/puppetlabs/puppet-server/pull/464) because we'll need to update the [pin on beaker](https://github.com/puppetlabs/puppet-server/blob/5ad6dd7bffe31ab4fc4efad7aa071830def81f6a/Gemfile#L18) once this pull request is merged into master.

This will enable the puppet-server test to behave as desired.

anodelman:
Broke spec:

```
Failures:

  1) Beaker::SshConnection close sets the @ssh variable to nil
     Failure/Error: expect( connection.instance_variable_get(:@ssh) ).to be_nil
       expected: nil
            got: #<Object:0x0000000419bd20>
     # ./spec/beaker/ssh_connection_spec.rb:47:in `block (2 levels) in <module:Beaker>'

  2) Beaker::SshConnection close calls ssh shutdown if ssh close fails
     Failure/Error: allow( mock_ssh ).to receive( :close ) { raise Error }
     NameError:
       uninitialized constant Beaker::Error
     # ./spec/beaker/ssh_connection_spec.rb:52:in `block (3 levels) in <module:Beaker>'
     # ./lib/beaker/ssh_connection.rb:64:in `close'
```

erikPrime:
I'm not convinced this PR moves things in the right direction.  It is possible that we should instead be investigating the registration of a "on_close" call back, so that when the far end closes a channel, we gracefully close on our side before encountering the IOError: Stream End, and that if we are closing the last channel in the Session, we close the session.

PR #777, opened 2015-04-06
(BKR-184) generalized repo selection code for install_puppetlabs_dev_repo

kevpl:


lib/beaker/dsl/install_utils/puppet_utils.rb
```diff
@@ -583,25 +588,22 @@ def install_puppetlabs_dev_repo ( host, package_name, build_version,
...
-              link = "%s/%s/%s/repos/%s/%s%s/products/%s/" %
-                [ dev_builds_url, package_name, build_version, variant,
-                  fedora_prefix, version, arch ]
-            end
-
-            if not link_exists?( link )
-              link = "%s/%s/%s/repos/%s/%s%s/devel/%s/" %
+            link = nil
+            package_repos = options[:dev_builds_repos].nil? ? [] : [options[:dev_builds_repos]]
+            package_repos.push(['PC1', 'products', 'devel']).flatten!
```
joshcooper: I don't think PC1 should be present in beaker, even as a fallback. The reason is that I can imagine a scenario where a caller of this method, e.g. puppetserver, wants to use puppet-agent from PC2, but forgets to specify `dev_builds_repos`, and pulls in the version of PC1, and things just might "work" and fail sometime later due to versioning issues between puppet-agent and puppetserver. I think it would be better for the caller to always specify the PCN repos, and maybe fall back to "products and devel"... /cc @haus @jeffmccune

haus: I agree. Ideally projects will know what collection they expect to be testing against.

lib/beaker/dsl/install_utils/puppet_utils.rb
```diff
@@ -545,6 +545,11 @@ def install_puppetlabs_release_repo ( host )
         #
         # @note This method only works on redhat-like and debian-like hosts.
         #
+        # @note You can pass in additional package_repos to try using the global
+        #   options hash with the +:dev_builds_repos+ key.  This can be a
+        #   string value, or an array of repos to look through.  It will look
+        #   through them in the order given.
+        #
         def install_puppetlabs_dev_repo ( host, package_name, build_version,
                                   repo_configs_dir = 'tmp/repo_configs' )
```
anodelman: Let's pass in the options here with opts = options instead of using the global options directly in the method.

kevpl:
failures unrelated

anodelman:
Thanks for the update, looks good.

lib/beaker/dsl/install_utils/puppet_utils.rb
```diff
@@ -584,25 +589,22 @@ def install_puppetlabs_dev_repo ( host, package_name, build_version,
...
-              link = "%s/%s/%s/repos/%s/%s%s/products/%s/" %
-                [ dev_builds_url, package_name, build_version, variant,
-                  fedora_prefix, version, arch ]
-            end
-
-            if not link_exists?( link )
-              link = "%s/%s/%s/repos/%s/%s%s/devel/%s/" %
+            link = nil
+            package_repos = opts[:dev_builds_repos].nil? ? [] : [opts[:dev_builds_repos]]
+            package_repos.push(['products', 'devel']).flatten!
```
justinstoller: I feel a little dense. How does this continue to work `pc1` repos?

haus: @justinstoller looks like you would need to set opts[:dev_builds_repos] to include PC1 to me? then it is included in the package_repos to iterate over. I *think*.

kevpl: @justinstoller @haus is right about how it should be working.

I have questions for you on this, I'll add them to the ticket momentarily

PR #781, opened 2015-04-07
(BKR-192) can't run beaker without a host file specified

anodelman:
- the hosts file was being defaulted to to provide a log name prefix -
  it needs a default value for when there is no user provided value plus
  no hosts file


rji:
Patch looks good to me, waiting on CI

rji:
:tada: CI is green. :+1:

PR #783, opened 2015-04-08
(BKR-70) Make SshConnection.close more robust...

anodelman:
...(so that on(host, "reboot") works correctly)

- catch IOErrors
- added additional execution option ':expect_connection_failure' for
  operations that you believe should result in the connection to the
  host being dropped or broken
- tested with rebooting centos7 box, successfully rebuilds connection
  and continues test execution

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -396,10 +396,19 @@ module Beaker
         unless options[:silent]
           # What?
           result.log(@logger)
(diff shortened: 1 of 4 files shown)


anodelman:
Failure unrelated to patch (timeout during single test)

sschneid:
:+1: looks straightforward to me.

PR #784, opened 2015-04-10
(BKR-182) Disable check_for_updates by default with beaker

anodelman:
- added to version 3.2+


stahnma:
:+1: from me, pending travis.

PR #787, opened 2015-04-16
(BKR-209) check for package and host prebuild sles10

er0ck:
this change fixes check_for_package on sles10 in which zypper always
returns 0, even if a package is not found.
This change also removes ntp package from the host pre-build steps.
sles10 comes with ntp_command installed.


anodelman:
retest this please

PR #791, opened 2015-04-22
(BKR-217) aws_sdk: given CONFIG subnet_ids, try each in turn

rbrw:
(BKR-217) aws_sdk: given CONFIG subnet_ids, try each in turn

    This should help alleviate the EC2 InsufficientInstanceCapacity.

    The new policy will only be applied to a host if there isn't a CONFIG
    subnet_id and there isn't a host-specific subnet_id.  When it does
    apply, try each subnet in subnet_ids once for each relevant host.

    The new option can be used like this:

      CONFIG
        vpc_id: foo
        subnet_ids:
          - x
          - y
          - z

### DIFF
--- a/lib/beaker/hypervisor/aws_sdk.rb
+++ b/lib/beaker/hypervisor/aws_sdk.rb
@@ -225,10 +225,9 @@ module Beaker
     #
     # @return [AWS::EC2::Instance)]
     # @api private
-    def create_instance(host, ami_spec)
+    def create_instance(host, ami_spec, subnet_id)
       amitype = host['vmname'] || host['platform']
       amisize = host['amisize'] || 'm1.small'
-      subnet_id = host['subnet_id'] || @options['subnet_id'] || nil
       vpc_id = host['vpc_id'] || @options['vpc_id'] || nil
 
       if vpc_id and !subnet_id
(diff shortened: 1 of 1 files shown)


rbrw:
I tried to break up the work into more easily digested commits, though in some cases, ignoring whitespace may help (i.e. in github, by adding ?w=1 to the URL).  In any case, assuming the general idea seems reasonable, I'm happy to adjust the work as you like.

We've tested this by running it through the normal PuppetDB acceptance tests with a vpc_id and subnet_ids setting, and since we can't easily force a capacity failure, I tested that locally (manual beaker runs) using some temporary instrumentation that would throw the InsufficientCapacity exception whenever the subnet was named 'no-capacity'.  I did something similar to test the more general instance teardown code.

Finally, I ran the tests through Travis, and they seemed fine on everything except Ruby 1.8.7 where they failed with "Gem::InstallError: rubyzip requires Ruby version >= 1.9.2.".

lib/beaker/hypervisor/aws_sdk.rb
```diff
@@ -69,6 +66,20 @@ def provision
       nil #void
     end
 
+    # Kill all instances.
+    #
+    # @param instances [Enumerable<EC2::Instance>]
+    # @return [void]
+    def kill_instances(instances)
+      instances.each do |instance|
+        if instance.exists?
```
anodelman: You dropped the instance.nil? check, please add that back.

lib/beaker/hypervisor/aws_sdk.rb
```diff
@@ -220,125 +221,171 @@ def kill_zombie_volumes
...
+    # Create EC2 instances for all hosts, tag them, and wait until
+    # they're running.
+    #
+    # @return [void]
+    # @api private
+    def launch_all_nodes
+      @logger.notify("aws-sdk: launch all hosts in configuration")
+      ami_spec = YAML.load_file(@options[:ec2_yaml])["AMI"]
+      global_subnet_id = @options['subnet_id']
+      global_subnets = @options['subnet_ids']
```
anodelman: What's the different between subnet_id and subnets?  Just a single subnet vs. array?  Because that could be held in the same variable.

lib/beaker/hypervisor/aws_sdk.rb
```diff
@@ -220,125 +221,171 @@ def kill_zombie_volumes
...
+      if global_subnet_id && global_subnets
+        raise RuntimeError, 'Config specifies both subnet_id and subnet_ids'
+      end
+      any_subnet_hosts, other_hosts = @hosts.partition do |host|
```
anodelman: Could we have a different name than 'other_hosts', it's not very descriptive of why they are 'other'.

lib/beaker/hypervisor/aws_sdk.rb
```diff
@@ -220,125 +221,171 @@ def kill_zombie_volumes
...
+      end
+      instances = [] # Each element is {:instance => i, :host => h}
+      begin
+        @logger.notify("aws-sdk: launch instances not particular about subnet")
+        # Shuffle the subnets so we don't always hit the same one
+        # first, and cycle though the subnets independently of the
+        # host, so we stick with one that's working.  Try each subnet
+        # once per-host.
+        subnet_i = 0
+        if global_subnets
```
anodelman: If you are only going through hosts with subnets assigned do you really need to first check if global_subnets is set?  The hash of hosts will be empty so you don't need this check?

lib/beaker/hypervisor/aws_sdk.rb
```diff
@@ -220,125 +221,171 @@ def kill_zombie_volumes
...
+          shuffnets = global_subnets.shuffle
+          any_subnet_hosts.each do |host|
+            instance = nil
+            shuffnets.length.times do
+              begin
+                subnet_id = shuffnets[subnet_i]
```
anodelman: Doing the create/push of an instance could live in a separate method and just called with appropriate parameters.

anodelman:
retest this please

lib/beaker/hypervisor/aws_sdk.rb
rbrw: @anodelman Certainly.  (For what it's worth, if I recall correctly, I had done that because all of the callers guarantee that now themselves.)

lib/beaker/hypervisor/aws_sdk.rb
rbrw: @anodelman Ahh, right, thanks - should we be doing any other sanity checking on the value, i.e. make sure it's enumerable or similar?  Or is that unnecessary?

lib/beaker/hypervisor/aws_sdk.rb
rbrw: @anodelman assuming I understand you correctly, then I"m not sure - we're also using this to distinguish the cases where we want the new behavior vs the old.  And while I think we might be able to treat a subnet_id spec as a degenerate case of subnet_ids, I think doing so might add some extraneous complexity to that code path.

Let's see what you think after I post the next version, which will hopefully address your other concerns.

lib/beaker/hypervisor/aws_sdk.rb
rbrw: @anodelman ...and indeed, now I recall why I had that there -- it's to guard the subnets.shuffle (if nothing else) since subnets will be nil when not specified.  But I can change it to something like "shuffnets = !any_subnet_hosts.empty? && subnets.shuffle" or similar if you prefer.

rbrw: @anodelman actually, maybe nevermind.  In the new version, where I broke out the code into a separate function, we can just short-circuit at the top, i.e.:

      if subnets.nil? or subnets.empty?:
           return

rbrw:
@anodelman OK, I hope I've addressed most of your concerns in the update I just pushed.  I've also pushed a rebased branch for the previous version in case you want to see the relative changes.  i.e. via "git diff ticket/master/bkr-217-rotate-ec2-subnets-v1-rebased..ticket/master/bkr-217-rotate-ec2-subnets" with respect to my repository.

anodelman:
I'm going to merge this - could you update the wiki to describe the new feature added:

https://github.com/puppetlabs/beaker/wiki/EC2-Support

PR #795, opened 2015-04-27
(BKR-225) rsync module_utils method source is always nil

fiddyspence:
The rsync method uses 'source' as the variable to rsync from, but sadly it's nil


kevpl:
:+1:

kevpl:

PR #796, opened 2015-04-27
(BKR-226) - rsync fails because of NATted ports

fiddyspence:


anodelman:
ok to test

anodelman:
Breaks spec:

```
Failures:

  1) Beaker::Host do_rsync_to do_rsync_to logs info and call Rsync class
     Failure/Error: expect(Rsync.host).to eq('root@host.example.org')
       
       expected: "root@host.example.org"
            got: "root@default.ip.address"
       
       (compared using ==)
     # ./spec/beaker/host_spec.rb:578:in `block (3 levels) in <module:Beaker>'
```

fiddyspence:
I've fixed the test by making it test equality with the default ip set by the make_host method, which is actually a decent test because it sets it to a string.

anodelman:
Breaks spec ruby 2.0+:

```
  1) Beaker::Host do_rsync_to do_rsync_to logs info and call Rsync class
     Failure/Error: expect( host ).to receive(:to_s).and_return('host.example.org')
       (host.example.org).to_s(*(any args))
           expected: 1 time with any arguments
           received: 0 times with any arguments
     # ./spec/beaker/host_spec.rb:575:in `block (3 levels) in <module:Beaker>'
```1

anodelman:
Still breaking spec.

anodelman:
Last comment still current.

anodelman:

anodelman:
Looks good.

PR #798, opened 2015-04-28
(BKR-223) Change order of how we assign IPs to openstack instances

fiddyspence:
This reverts much of the previous behaviour and only falls back to introspecting the existing
data from openstack to assign an address (this is it failing over) - previously we'd have got a response from excon:

    /Users/fids/.rbenv/versions/1.9.3-p545/lib/ruby/gems/1.9.1/gems/excon-0.44.1/lib/excon/middlewares/expects.rb:10:in `response_call': Expected([200, 202]) <=> Actual(404 Not Found) (Fog::Compute::OpenStack::NotFound)
    excon.error.response
      :body          => "{\"itemNotFound\": {\"message\": \"FloatingIpPoolNotFound: Floating ip pool not found.\", \"code\": 404}}"
      :headers       => {
        "Content-Length"       => "97"
        "Content-Type"         => "application/json; charset=UTF-8"
        "Date"                 => "Tue, 28 Apr 2015 08:35:28 GMT"
        "X-Compute-Request-Id" => "req-df60a05d-930a-4bcc-b741-34076b6faf47"
      }
      :local_address => "10.12.216.212"
      :local_port    => 63301
      :reason_phrase => "Not Found"
      :remote_ip     => "10.52.131.38"
      :status        => 404

After this PR

    Beaker::Hypervisor, found some openstack boxes to create
    Provisioning OpenStack
    Warning: Try 1 -- Host 10.52.203.29 unreachable: Connection refused - connect(2)
    Warning: Trying again in 1 seconds

### DIFF
--- a/lib/beaker/hypervisor/openstack.rb
+++ b/lib/beaker/hypervisor/openstack.rb
@@ -14,7 +14,7 @@ module Beaker
     #@option options [String] :openstack_auth_url The URL to access the OpenStack instance with (required)
     #@option options [String] :openstack_tenant The tenant to access the OpenStack instance with (required)
     #@option options [String] :openstack_network The network that each OpenStack instance should be contacted through (required)
-    #@option options [String] :openstack_keyname The name of an existing key pair that should be auto-loaded onto each 
+    #@option options [String] :openstack_keyname The name of an existing key pair that should be auto-loaded onto each
     #                                            OpenStack instance (optional)
     #@option options [String] :jenkins_build_url Added as metadata to each OpenStack instance
     #@option options [String] :department Added as metadata to each OpenStack instance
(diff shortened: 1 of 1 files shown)


mcanevet:
@fiddyspence @anodelman I tested this patch and it solves the bug I reported.

anodelman:
@mcanevet thanks for doing some testing on this.

rafaduran:
Hi all,

After installing a new beaker version, with this change included, I'm getting issues with floatings IPs. With previous versions, if something were wrong I was getting a FloatingIpPoolNotFound error. That wasn't perfect but current situation isn't better, since now I'm getting timeouts due to beaker trying to connect the VM private network interface.

I've also realized that this call is wrong:
https://github.com/puppetlabs/beaker/blob/master/lib/beaker/hypervisor/openstack.rb#L139

And I think it should be replaced by something like

```ruby
ip = @compute_client.addresses.create :name => 'your_floating_ip_network'
```

Adding a new parameter "floating_ip_openstack_network" and updating the code to make the right create call should be enough.

anodelman:
@rafaduran could you please open a beaker jira bug with your comment?

rafaduran:
@anodelman Ok, I will do it.

PR #799, opened 2015-04-28
(BKR-218) Support using vmpooler API tokens

sschneid:


lib/beaker/hypervisor/vmpooler.rb
```diff
@@ -19,6 +19,18 @@ def initialize(vmpooler_hosts, options)
       @options = options
       @logger = options[:logger]
       @hosts = vmpooler_hosts
+      @credentials = load_credentials(@options[:dot_fog])
+    end
+
+    def load_credentials(dot_fog = '.fog')
```
colinPL: What about `~/.fog`? It is common for the file to be under the user's home directory.

sschneid: It's actually loading from the `@options[:dot_fog]` location, this is just a default (stolen almost completely from the AWS hypervisor).

sschneid:
I've generated a token for the `jenkins` user and added it to `${secure_server}/secure/delivery/fog`file.  This updated credentials file should be deployed to Jenkins slaves as they perform Puppet runs.

sschneid:

lib/beaker/hypervisor/vmpooler.rb
```diff
@@ -19,6 +19,18 @@ def initialize(vmpooler_hosts, options)
       @options = options
       @logger = options[:logger]
       @hosts = vmpooler_hosts
+      @credentials = load_credentials(@options[:dot_fog])
```
anodelman: This will add a dependency on the presence of a .fog file for executing this code.  I think that we could reasonable add that (and update documentation to reflect it), or we need to add a switch to beaker to attempt loading in the case of 'vpool_auth = true' or some such.

sschneid: Is there somewhere that specifies which hypervisors need access to a credentials file?  I didn't see it anywhere in documentation so I'm not sure where to update?

anodelman: The way this is written it will always attempt to load the .fog since beaker provides a default value for options[:dot_fog] - so, I believe, that this will now throw an error if one isn't specified.

I know that it is hacky, but we may need an additional switch param here to indicate that we are going to be running authorized so that we can avoid always attempting to open the dot fog.

colinPL: Let it try to load the file? Have `load_credentials` successfully catch exceptions and print a line in the log that it failed to load `.fog` and is continuing without credentials.

sschneid: Validated that this does indeed fail without a `.fog` file present...

````
Hypervisor for debian-7-x86_64-1 is vmpooler
Beaker::Hypervisor, found some vmpooler boxes to create
Failed: errored in CLI.provision
#<Errno::ENOENT: No such file or directory - /Users/sschneider/.fog>
````

Comment (sschneid) [inline lib/beaker/hypervisor/vmpooler.rb]: > Let it try to load the file? Have load_credentials successfully catch exceptions and print a line in the log that it failed to load .fog and is continuing without credentials.

Seems logical... I'll work on this and update the PR when I have it functioning as desired.

Comment (sschneid) [inline lib/beaker/hypervisor/vmpooler.rb]: > I'll work on this and update the PR when I have it functioning as desired.

Updated PR.  Manually tested `.fog` with `:vmpooler_token`, `.fog` without `:vmpooler_token`, and no `.fog`.

**Comment by kevpl** [comment]
:+1: 

@anodelman would you verify that your concerns are addressed here?

File: lib/beaker/hypervisor/vmpooler.rb [hunk 3]
```diff
@@ -19,6 +19,22 @@ def initialize(vmpooler_hosts, options)
...
+    def load_credentials(dot_fog = '.fog')
+      creds = {}
+
+      begin
+        fog = YAML.load_file(dot_fog)
+        default = fog[:default]
+
+        creds[:vmpooler_token] = default[:vmpooler_token]
+      rescue Errno::ENOENT
+        @logger.notify "Credentials file (#{@options[:dot_fog]}) not found; proceeding without authentication"
```
anodelman: Nit - can you make this a warning?

anodelman:
One final nit, otherwise :+1:

lib/beaker/hypervisor/vmpooler.rb
sschneid: Changed from `notify` to `warning`.

kevpl:

sschneid:
@puppetlabs-jenkins you're a jerk.

anodelman:
Failing tests are using ec2, not vpooler.  This is safe to merge.

cyberious:
@anodelman where is the ticket to document this new change? I just got hit with but figured out by code diving.

sschneid:
@cyberious Can you provide more details on what sort of troubles you ran into?

djm68:
https://groups.google.com/a/puppetlabs.com/forum/#!topic/dev-team/zcZRteV0ens

PR #801, opened 2015-04-28
(BKR-215) Document all of the options to #on and friends

anodelman:
- updates shared options to be current (with info for
  accept_all_exit_codes and friends)
- yard doc does a single pass and you can't guarantee file ordering, so
  you have to have the common options macro in each file that needs it -
  duplication yay!
- update usage of acceptable_exit_codes to accept_all_exit_codes when we are just passing acceptable_exit_codes a mega array


anodelman:
Failure unrelated to patch.

PR #802, opened 2015-04-30
(BKR-191) replace is_cygwin? with a more understandable name

anodelman:
- create is_powershell?, update usages of is_cygwin as appropriate
- move powershell only code into pswindows host object
- fixes related to breakages discovered when this code was moved

PR #806, opened 2015-05-04
(BKR-240) added a reason code on reboot for windows

kevpl:


kevpl:
@anodelman I don't think support from the OS is the issue here.  According to (their docs)[https://technet.microsoft.com/en-us/library/cc732503.aspx?f=255&MSPPError=-2147217396], it's supported in both win2003r2 & win2008r2.

Checking into the Jenkins failure, it's only the echo command that's failing after, the execution of the shutdown command doesn't appear to have a problem.  I can run the command locally, and get the result that an echo issued right after a reboot will succeed, but if I preserve the host and try to ssh into it after, I can't get to it (a vmpooler host) for another minute or two.  I get the same type of failure seen in the Jenkins if I sleep during the test before running the echo command.  I suspect that this is a timing issue with the reboot, and think the solution here is a long back off to let the reboot finish, or to block the return until the ssh connection successfully dies and restarts after reboot finishes.

anodelman:
I think that the solution to windows reboot is in:

https://github.com/puppetlabs/beaker/pull/802

I discovered the issue when I was playing around with is_cygwin and fixed it so that I could get my local testing running.

kevpl:
#802 just merged, so I'll kick off another cycle of testing here to see if that solves this issue.

kevpl:

anodelman:
retest this please

kevpl:
failure unrelated

PR #810, opened 2015-05-05
(BKR-186) added current test info to the options hash for querying

kevpl:


lib/beaker/dsl/structure.rb
```diff
@@ -37,6 +37,11 @@ module Structure
       # @param [Proc] block The actions to be performed in this step.
       def step step_name, &block
         logger.notify "\n  * #{step_name}\n"
+        if @options
```
anodelman: I'm not sure about putting this in the options.  I would see it living in an object and then being queried on through a class - a more object oriented, ruby solution.

PR #811, opened 2015-05-07
(BKR-305)Adding function to have custom docker options per host

fiddyspence:
Sometimes, one needs to pass custom options to Docker when creating a host.

This patch allows a :dockeropts key to be added to the nodeset, which is passed to the Docker API

### DIFF
--- a/lib/beaker/hypervisor/docker.rb
+++ b/lib/beaker/hypervisor/docker.rb
@@ -65,10 +65,24 @@ module Beaker
         end
 
         @logger.debug("Creating container from image #{image_name}")
-        container = ::Docker::Container.create({
-          'Image' => image_name,
-          'Hostname' => host.name,
-        })
+        image_hash = {
+           'Image' => image_name,
+           'Hostname' => host.name,
+        }
+
+        if options[:dockeropts]
+          options[:dockeropts].each do |k,v|
+            image_hash[k] = v
+          end
+        end
+
+        if host[:dockeropts]
+          host[:dockeropts].each do |k,v|
+            image_hash[k] = v
+          end
+        end
+
+        container = ::Docker::Container.create(image_hash)
 
         @logger.debug("Starting container #{container.id}")
         container.start({"PublishAllPorts" => true, "Privileged" => true})


lib/beaker/hypervisor/docker.rb
```diff
@@ -33,10 +33,19 @@ def provision
...
         @logger.debug("Creating container from image #{image.id}")
-        container = ::Docker::Container.create({
+
+        image_hash = {
           'Image' => image.id,
           'Hostname' => host.name,
-        })
+        }
+
+        if host[:dockeropts]
```
anodelman: This is only per-host, want to do a ```host[:dockeropts] || options[:dockeropts]``` so that you can set a default set of options globally.

anodelman:
Review comment still current.  Also, commit should reference a jira ticket number like (BKR-xxx).

anodelman:

anodelman:

lib/beaker/hypervisor/docker.rb
```diff
@@ -65,10 +65,24 @@ def provision
...
+           'Hostname' => host.name,
+        }
+
+        if options[:dockeropts]
+          options[:dockeropts].each do |k,v|
+            image_hash[k] = v
+          end
+        end
+
+        if host[:dockeropts]
```
kevpl: Hi @fiddyspence,

This will be a little more maintainable if you eliminate the duplicate code block.  You can do that with these lines changed to something like this:

    if host[:dockeropts] || options[:dockeropts]
      dockeropts = host[:dockeropts] ? host[:dockeropts] : options[:dockeropts]
      dockeropts.each do |k,v|

Also, please add a unit test, so that we can verify that we don't break this new functionality accidentally in the future.

Let me know if you have any questions about this, and we look forward to getting this fix in!

Thanks,
Kevin

anodelman:
Review comment still current.

anodelman:
Will close out next week due to inactivity.

anodelman:
Closing due to inactivity.

PR #812, opened 2015-05-07
(BKR-249) Add OpenBSD support

bodgit:


lib/beaker/dsl/install_utils/puppet_utils.rb
```diff
@@ -197,7 +197,7 @@ def install_puppet(opts = {})
 
             # Certain install paths may not create the config dirs/files needed
             on host, "mkdir -p #{host['puppetpath']}" unless host[:type] =~ /aio/
-            on host, "echo '' >> #{host.puppet['hiera_config']}"
+            #on host, "echo '' >> #{host.puppet['hiera_config']}"
```
petems: This should probably be changed to a `printf` or something that works on both rather than commented out.

lib/beaker/host_prebuilt_steps.rb
```diff
@@ -373,6 +380,8 @@ def enable_root_login host, opts
           host.exec(Command.new("sudo -E /sbin/service sshd reload"), {:pty => true})
         elsif host['platform'] =~ /freebsd/
           host.exec(Command.new("sudo /etc/rc.d/sshd restart"))
```
petems: Since both freebsd and openbsd have the same sshd restart command, this could probably be changed to:

```ruby
elsif host['platform'] =~ /(free|open)bsd/
           host.exec(Command.new("sudo /etc/rc.d/sshd restart"))
end
```

petems:
You're going to need to squash commits and add (BKR-249) to the start of commit messages also :+1:

lib/beaker/host_prebuilt_steps.rb
```diff
@@ -555,6 +567,10 @@ def set_env host, opts
           host.exec(Command.new("touch #{host[:ssh_env_file]}"))
           #add the constructed env vars to this host
           host.add_env_var('PATH', '$PATH')
+          # FIXME
+          if host['platform'] =~ /openbsd-([^-]+)-(.+)/
+            host.add_env_var('PKG_PATH', "http://ftp.openbsd.org/pub/OpenBSD/#{$1}/packages/#{$2}/")
```
petems: This should probably auto-translate x64 -> amd64 (ie. http://ftp.openbsd.org/pub/OpenBSD/5.6/packages/amd64/)

petems:
I got this (kinda) working with the following hostfile:

```ruby
HOSTS:
  openbsd-56-amd64:
    roles:
      - master
    platform: openbsd-5.6-amd64
    box : tmatilai/openbsd-5.6
    box_url : https://vagrantcloud.com/tmatilai/boxes/openbsd-5.6/versions/1.0.0/providers/virtualbox.box
    hypervisor : vagrant
CONFIG:
  type: git
```

However, the module wasn't found, probably a pathing thing...

bodgit:
Yes, this was something I mentioned in the Jira ticket. Because it pulls in the latest Puppet gem it now gets 4.x.x installed which has changed a bunch of paths, the `copy_module_to` helper puts the module in the old pre-4.x.x location. Using `$BEAKER_{provision,destroy}` appropriately I can manually move the module and resume the test and get things to pass but I think there's some work required to make things JFW on both < 4.x.x and >= 4.x.x.

I've also created a PR to make some OpenBSD 5.7 images, it seems Puppet can't manage `service` resources on earlier releases.

lib/beaker/dsl/install_utils/puppet_utils.rb
bodgit: This was due to it trying to echo to `/etc/puppetlabs/code/hiera.yaml` which if installing by gem lacks all of that directory structure. Commenting it out didn't break anything so I'm not sure if it even needs to be there.

Possibly the `install_puppet_from_gem` helper should be enhanced to ensure post-installation that directories such as `/etc/puppetlabs/puppet`, `/etc/puppetlabs/code`, `/var/log/puppetlabs/puppet`, etc. should exist.

lib/beaker/host_prebuilt_steps.rb
bodgit: Sure, makes sense.

anodelman:
@KevinCorcoran - can you add a comment here about 4.x support in beaker?

petems:
Hey @jasperla! @adrienthebo said you might be interested in this: OpenBSD support in Beaker. Any strong feelings about the way the install or setup is being done?

jasperla:
Excellent! I'm currently very short on time so I cannot run-verify this patch, but a quick note on paths:

third party packages should be installed into `/usr/local`, so the path to puppet will be `/usr/local/bin/puppet`, however configuration lives in `/etc`, e.g. `/etc/puppet/puppet.conf`. Also, system users and groups associated with a specific piece of software or service, are prefixed with an underscore e.g. `_puppet` and `_bind`. Although this is something currently enforced in the OpenBSD port but not applied upstream.

anodelman:
Shout out to @kevpl, previous question to you still current.

bodgit:
> third party packages should be installed into /usr/local, so the path to puppet will be /usr/local/bin/puppet, however configuration lives in /etc, e.g. /etc/puppet/puppet.conf. Also, system users and groups associated with a specific piece of software or service, are prefixed with an underscore e.g. _puppet and _bind. Although this is something currently enforced in the OpenBSD port but not applied upstream.

At the moment I've not used the puppet port, only the (newest) ruby port and then continued with gems to get the rest of the environment installed, (like how I assumed the FreeBSD support did it). I don't know what the policy is here, should I prefer the OS-provided puppet package if one is available?

The disadvantage here is that there is generally a 1:1 mapping of OpenBSD version to Puppet version as the port rarely changes (if at all) for the lifetime of the OS. Currently the latest version of OpenBSD has version 3.7.4 of Puppet. Relying on the OS port means I can't easily test a matrix of OS and Puppet versions, (although for my own personal use I will probably stick with the port for my own systems). That can also be an advantage that if I wish to simply have a "works on OpenBSD" tick against my module; I just test that one version per OS release. I can also fix the paths to work with how the port lays files out and ignore any 4.x.x-related issues as it won't be a problem until the port gets updated.

I would appreciate @kevpl to chime in here.

kevpl:
Hi @bodgit,

If I'm understanding correctly, the problem that you're seeing above is a pathing issue with Beaker for the newer releases of Puppet (4+).  Since the pathing has changed, you'll need to use the `--type` option, and set it to `aio` (All-in-One Agent).  Beaker should work fine for the 4.x version once you've done that.  I was delaying until I got around to writing the proper doc on that, but I've waited too long, so here's me just saying it :-/.  Let me know if you're still having issues once you set the type parameter.

For the OS issue, I'm not familiar with BSD, but the new pathing for 4.x is defined [here](https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md). If you have more questions about BSD interacting with Puppet, I would suggest asking the Puppet teams in #puppet-dev.

bodgit:
Thanks @kevpl, setting `type: aio` at the bottom of my OpenBSD nodeset made things work.

anodelman:
We do not have any openbsd boxes to test this against.  I'm okay with merging that, as long as we have a maintainer who is willing to review and test additional patches.  @bodgit ?

bodgit:
Sure, happy to help. I'm using beaker to test modules targeted for OpenBSD so I have a vested interest for it to keep working.

Incidentally, I've uploaded my later 5.7 vagrant images to vagrant-cloud/atlas if that's any use?

jasperla:
@bodgit sounds worth sharing :)

anodelman:
I'm going to update the wiki indicating ownership of various sections of the code, then I'll get this merged up.

kevpl:
@bodgit I've added a doc on the [AIO Type](https://github.com/puppetlabs/beaker/wiki/Types,-Puppet-4,-and-the-All-In-One-Agent), so hopefully we'll have less of this confusion going forward.

anodelman:
Needs a rebase.

bodgit:
Ok, I'll try and rebase shortly. Hopefully it's not diverged too much

anodelman:
@bodgit i have added you to https://github.com/puppetlabs/beaker/wiki/Beaker-Owners-&-Reviewers in preparation for this being landed.

bodgit:
Rebased, bear with me while I just check it all still works.

bodgit:
Went down a bit of a rabbit hole:

* I found https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/install_utils/foss_utils.rb#L247 which means that Puppet 3.8.1 is the default gem-installed version.
* My logic for installing Ruby was to basically install the latest available and as OpenBSD has packages for all rubies ranging from 1.8 through to 2.2 it was choosing 2.2. Turns out Puppet 3.x doesn't work with 2.2 and I found PUP-3796 which confirms that'll likely never change now. I've added logic that prefers the newest ruby that is also older than 2.2 so 2.1 is now preferred and Puppet 3.x should be happy.
* However it turns out that stock 3.x Puppet just doesn't work on OpenBSD; there isn't a valid `user` provider available so Puppet bails out unable to apply anything. It's in 4.x and I think the OpenBSD port backports those changes.

So in the end I've just updated my `spec_helper_acceptance.rb` with:

```ruby
install_puppet({ :default_action => 'gem_install', :version => '4' })
```

So I get the latest 4.x gem installed. My tests work so from that I think my changes have been successfully rebased :-)

lib/beaker/host/openbsd.rb
```diff
@@ -0,0 +1,18 @@
+[ 'host', 'command_factory' ].each do |lib|
+  require "beaker/#{lib}"
+end
+
+module OpenBSD
+  class Host < Unix::Host
+
+    def self.foss_defaults
```
anodelman: These are now known and 'platform_defaults'.

lib/beaker/host_prebuilt_steps.rb
```diff
@@ -359,6 +364,8 @@ def enable_root_login host, opts
           host.exec(Command.new("sudo sed -i '' 's/#PermitRootLogin yes/PermitRootLogin Yes/g' /etc/sshd_config"))
         elsif host['platform'] =~ /freebsd/
           host.exec(Command.new("sudo sed -i -e 's/#PermitRootLogin no/PermitRootLogin yes/g' /etc/ssh/sshd_config"), {:pty => true} )
+        elsif host['platform'] =~ /openbsd/
+          host.exec(Command.new("sudo perl -pi -e 's/^PermitRootLogin no/PermitRootLogin yes/' /etc/ssh/sshd_config"), {:pty => true} )
```
anodelman: Will openbsd consistently have perl installed?  We try and avoid adding additional dependencies to the SUTs.

lib/beaker/host_prebuilt_steps.rb
```diff
@@ -523,6 +530,9 @@ def set_env host, opts
         when /freebsd/
           host.echo_to_file('\nPermitUserEnvironment yes', '/etc/ssh/sshd_config')
           host.exec(Command.new("sudo /etc/rc.d/sshd restart"))
+        when /openbsd/
+          host.exec(Command.new("sudo perl -pi -e 's/^#?PermitUserEnvironment no/PermitUserEnvironment yes/' /etc/ssh/sshd_config"), {:pty => true} )
```
anodelman: Another perl pie, same question as before.

lib/beaker/dsl/install_utils/foss_defaults.rb
```diff
@@ -24,6 +24,21 @@ module FOSSDefaults
             'distmoduledir'     => '/usr/local/etc/puppet/modules',
             'sitemoduledir'     => '/usr/share/puppet/modules',
           },
+          'openbsd' => {
```
anodelman: Please make similar changes to pe_defaults.rb & aio_defaults.rb.

lib/beaker/host_prebuilt_steps.rb
petems: Most of the OpenBSD and FreeBSD boxes I've seen have perl installed, as it's a requirement to run VMWare tools and Virtualbox stuff. 

All of the openbsd boxes on VagrantCloud had it that I tested (4 separate) and the two cloud providers that I tried (Cloudsigma and M5Cloud) both had perl on the base image.

jasperla: perl is part of the OpenBSD base installation. There is no way not to install perl.

bodgit: It's part of the base system; the package/port tools `pkg_*` are all written in Perl.

petems: Awesome :smile:

bodgit: It was that or try and battle around the fact OpenBSD `sed` doesn't support `-i` :wink:

jasperla: It'd be great to have `sed -i` on OpenBSD, so if you feel bored feel free to submit a patch ;-)

bodgit:
I'm tempted to add similar logic here for OpenBSD as @petems has proposed for FreeBSD in #849. 

The problem I've found is installing by gem method currently doesn't pull in auxiliary gems for things like augeas which requires both installing the necessary gems but also the various C libraries which might be provided by the native OS packaging.

The gem install method changes can stay as they're still of use if you want to deviate from the version provided by native OS packaging although I've found trying to install gem versions of Puppet prior to 4.x practically unusable without patching.

petems:
Yeah, I feel like using the port makes more sense than the gem install.

anodelman:
Review comments (minus perl question) still current.

bodgit:
Rebased again. I've added support to prefer using the OS packages which means we get Puppet 3.7.4 but it has all the right dependencies for augeas, etc. pulled in. I've also dropped subclassing the unix host type as now the paths are defined elsewhere there's no difference. For PE and AIO I've gone with the defaults, given Puppetlabs don't provide PE or AIO packages for OpenBSD it's speculative anyway.

I can foresee some problems with the paths, particularly AIO; they tend to differ based on type/version *and* installation method. For example AIO implies Puppet 4.x, but Puppet 4.x doesn't necessarily imply AIO if it happened to be installed from a gem. This means that both installation methods will be looking for modules in `/etc/puppetlabs/code/modules` but the binaries will likely be in different places with AIO putting them under `/opt/puppetlabs/bin` and the gem putting them probably somewhere like `/usr/local/bin`. Maybe it's just gem installs in general that are the odd one out here?

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -681,15 +704,23 @@ def install_puppet_from_gem_on( host, opts )
...
 
           ver_cmd = opts[:version] ? "-v '#{opts[:version]}'" : ''
-          on host, "gem install puppet #{ver_cmd} --no-ri --no-rdoc"
+          on host, "gem install puppet #{ver_cmd} --no-ri --no-rdoc --no-format-executable"
+
+          # A gem install might not necessarily create these
+          ['confdir', 'logdir', 'codedir'].each do |key|
+            if host.puppet.has_key?(key)
+              dir = host.puppet[key]
+              on host, "[ -d #{dir} ] || mkdir -p #{dir}"
```
anodelman: This won't work on non-bash supported system.  Use host.mkdir_p(dir) so that the correct code is run on the correct platform.

anodelman:
@bodgit You can specify what paths are installed using host['type'] = 'aio' to force the use of AIO (until such a time as that becomes the reasonable default.  As is, you will not be able to use AIO defaults (or pe defaults) as all and installation methods that depend upon that pathing will fail.

lib/beaker/host_prebuilt_steps.rb
```diff
@@ -523,6 +530,9 @@ def set_env host, opts
         when /freebsd/
           host.exec(Command.new("sudo perl -pi -e 's/^#?PermitUserEnvironment no/PermitUserEnvironment yes/' /etc/ssh/sshd_config"), {:pty => true} )
           host.exec(Command.new("sudo /etc/rc.d/sshd restart"))
+        when /openbsd/
+          host.exec(Command.new("sudo perl -pi -e 's/^#?PermitUserEnvironment no/PermitUserEnvironment yes/' /etc/ssh/sshd_config"), {:pty => true} )
+          host.exec(Command.new("sudo /etc/rc.d/sshd restart"))
```
kevpl: is this any different than the freebsd case above it?  If so, you should be sure to consolidate them like you've been doing with your `(free|open)bsd` matches

bodgit:
@anodelman I'm not sure I follow. The AIO defaults logic assumes anything is 'unix' if it isn't a windows platform so that catches OpenBSD there and I've added logic to the PE defaults so it also uses the unix paths there.

I can adds defaults for AIO but they'd be based on what the gem install does as that's all I can test and wouldn't match what a real AIO package would probably do.

anodelman:
The fall through doesn't work correctly for pe_defaults - it falls through to platform 'windows'.  aio_defaults will work as expected.

So, I believe that you only need to fix up pe_defaults and you should be good to go.

bodgit:
I did make some changes to `{add,remove}_pe_defaults_on()` so it should pick 'unix' same as FreeBSD. Anyway it's rebased again against master, there were some conflicts in `install_puppet_from_gem_on()` where the `block_on` loop was added, it should hopefully be good to go now.

anodelman:
ok to test

anodelman:
Errors during yard documentation generation:
```
[warn]: @param tag has unknown parameter name: host 
    in file `lib/beaker/dsl/install_utils/foss_utils.rb' near line 660
Files:         112
Modules:        69 (   39 undocumented)
Classes:        61 (   30 undocumented)
Constants:      86 (   64 undocumented)
Methods:       789 (  182 undocumented)
 68.66% documented
```

bodgit:
Ok, I /think/ this was a missing 's' in the comments. I've pushed a fix for it, unless you can shed some light on that warning/error.

anodelman:
Spec failures:
```
Failures:

  1) Beaker::Host#add_env_var does nothing if the key/value pair already exists
     Failure/Error: host.add_env_var('key', '/my/first/value')
       #<Beaker::Command (class)> received :new with unexpected arguments
         expected: ("grep KEY=.*\\/my\\/first\\/value ~/.ssh/environment")
              got: ("grep ^KEY=.*\\/my\\/first\\/value ~/.ssh/environment")
     # ./lib/beaker/host/unix/exec.rb:89:in `add_env_var'
     # ./spec/beaker/host_spec.rb:192:in `block (3 levels) in <module:Beaker>'

  2) Beaker::Host#add_env_var adds new line to environment file if no env var of that name already exists
     Failure/Error: host.add_env_var('key', '/my/first/value')
       #<Beaker::Command (class)> received :new with unexpected arguments
         expected: ("grep KEY=.*\\/my\\/first\\/value ~/.ssh/environment")
              got: ("grep ^KEY=.*\\/my\\/first\\/value ~/.ssh/environment")
     # ./lib/beaker/host/unix/exec.rb:89:in `add_env_var'
     # ./spec/beaker/host_spec.rb:203:in `block (3 levels) in <module:Beaker>'

  3) Beaker::Host#add_env_var updates existing line in environment file when adding additional value to existing variable
     Failure/Error: host.add_env_var('key', '/my/first/value')
```

bodgit:
Should be fixed now.

anodelman:
retest this please

bodgit:
I forgot I added the anchors to the grep patterns which would probably (and did) break any tests regarding them, but I've ran the tests locally now and they all pass so I'm not sure what the failure is.

anodelman:
Needs a rebase.

bodgit:
Done

kevpl:

bodgit:
Could someone let me know what it's failing on? Is it something I need to fix?

anodelman:
retest this please

anodelman:
puppet gem installation is failing on solaris:
```
haj4gfvxc2swai6.delivery.puppetlabs.net (solaris11-64-2) 09:55:33$ gem install puppet -v '3.7.5' --no-ri --no-rdoc --no-format-executable
Successfully installed facter-2.4.4
Successfully installed json_pure-1.8.2
Successfully installed hiera-1.3.4
Successfully installed puppet-3.7.5
4 gems installed

haj4gfvxc2swai6.delivery.puppetlabs.net (solaris11-64-2) executed in 53.13 seconds

haj4gfvxc2swai6.delivery.puppetlabs.net (solaris11-64-2) 09:56:26$ puppet agent --configprint all
bash: puppet: command not found
```

anodelman:
Post gem installation steps on sles failing:
```
ujyzhc7m98lhdad.delivery.puppetlabs.net (sles12-64-2) 09:52:37$ ln -s /usr/bin/facter.ruby* /usr/bin/facter
ln: failed to create symbolic link ‘/usr/bin/facter’: File exists

ujyzhc7m98lhdad.delivery.puppetlabs.net (sles12-64-2) executed in 0.19 seconds
Exited: 1
#<Beaker::Host::CommandFailure: Host 'ujyzhc7m98lhdad.delivery.puppetlabs.net' exited with 1 running:
 ln -s /usr/bin/facter.ruby* /usr/bin/facter
Last 10 lines of output were:
	ln: failed to create symbolic link ‘/usr/bin/facter’: File exists>
```

anodelman:
Looks like the addition of `--no-format-executable` to gem installation is causing the problem.

bodgit:
Ok, I've added logic for the gem flags so it only adds `--no-format-executable` for OpenBSD

anodelman:
retest this please

anodelman:
retest this please

anodelman:
Still wonky on sles/solaris:
```
aua6yhemnaaltm8.delivery.puppetlabs.net (sles12-64-2) 11:22:46$ gem install puppet -v '3.7.5' --no-ri --no-rdoc
Successfully installed facter-2.4.4
Successfully installed json_pure-1.8.2
Successfully installed hiera-1.3.4
Successfully installed puppet-3.7.5
4 gems installed

aua6yhemnaaltm8.delivery.puppetlabs.net (sles12-64-2) executed in 6.94 seconds

aua6yhemnaaltm8.delivery.puppetlabs.net (sles12-64-2) 11:22:53$ puppet agent --configprint all
bash: puppet: command not found

aua6yhemnaaltm8.delivery.puppetlabs.net (sles12-64-2) executed in 0.03 seconds
Exited: 127
#<Beaker::Host::CommandFailure: Host 'aua6yhemnaaltm8.delivery.puppetlabs.net' exited with 127 running:
 puppet agent --configprint all
Last 10 lines of output were:
	bash: puppet: command not found>
```

bodgit:
Ok, I'm a bit puzzled, I'm no longer adding `--no-format-executable` for SLES and Solaris but it is trying to invoke `puppet` and assume it's in `$PATH`. I've reordered the code slightly so for Solaris at least it symlinks `{facter,hiera,puppet}` into `$PATH` before trying to do the `--configprint all` bit. But I'm confused why SLES might still be bust, I can't see anywhere that does symlink trickery for that platform.

anodelman:
Hm yeah, it's still sles dying post gem installation.

bodgit:
Is there a particular SLES vagrant box I can try and replicate this on? Looking at `install_puppet_from_gem_on()` there's no SLES conditionals anywhere AFAIK so it drops through without any special tweaks, yet things aren't in the `$PATH` by the end.

anodelman:
Doesn't look like we provide a pre-built vagrant box for sles.  Let me try and carve out some time this week to investigate.

bodgit:
I found a SLES 11 vagrant box which has a 1.8.7 Ruby. `/usr/bin/gem install puppet -v 3.8.1 --no-ri --no-rdoc` JFW's and creates `/usr/bin/{facter,hiera,puppet}`. I couldn't find a SLES 12 box to see how it differs.

petems:
@bodgit I was playing around with this with a Jenkins module and I was getting the error when trying to run the code:

```
Error: /Stage[main]/Java/Package[java]/ensure: change from absent to present failed: No valid installpath found in /etc/pkg.conf and no source was set
```

Would it be an idea to add some code like this as a helper:
`echo "# OpenBSD pkg.conf\ninstallpath=http://ftp.openbsd.org/pub/OpenBSD/5.7/packages/amd64/" > /etc/pkg.conf`

bodgit:
Ok, I figured out what was happening. In `acceptance/pre_suite/puppet_gem/install.rb` it was special casing SLES and doing the `ln -s` there which is why I couldn't find it anywhere in `lib`! I think it should therefore be safe to re-enable `--no-format-executable` and it should work everywhere with it being reordered slightly for Solaris' benefit.

jasperla:
@bodgit Fwiw, OpenBSD 5.8 will come with `sed -i`, I committed that earlier this week.

bodgit:
@jasperla Yes, I saw that commit and figured you'd probably mention it at some point ;-)

anodelman:
Thanks for sticking with this!

Passing tests, looks good.

PR #817, opened 2015-05-19
(maint) Relax constraint on hocon

justinstoller:
Prior to this we were using the hocon gem and pinning to a z release.

This commit relaxes the version dependency to pull in an pre-1.0 hocon
release.

This is necessary for testing the hocon module itself.

### DIFF
--- a/beaker.gemspec
+++ b/beaker.gemspec
@@ -33,7 +33,7 @@ Gem::Specification.new do |s|
   # Run time dependencies
   s.add_runtime_dependency 'minitest', '~> 5.4'
   s.add_runtime_dependency 'json', '~> 1.8'
-  s.add_runtime_dependency 'hocon', '~> 0.0.4'
+  s.add_runtime_dependency 'hocon', '~> 0.1'
   s.add_runtime_dependency 'net-ssh', '~> 2.9'
   s.add_runtime_dependency 'net-scp', '~> 1.2'
   s.add_runtime_dependency 'inifile', '~> 2.0'


fpringvaldsen:
@justinstoller I believe the hocon module is actually using version 0.9.0 of the hocon gem, would it be possible to update this dependency so it'll pull in that version?

anodelman:
Failures unrelated to patch.

anodelman:
Please associated your commit with BKR-287.

justinstoller:
@fpringvaldsen sorry, this slipped though my radar. I was trying to make Beaker as loosely coupled to the hocon module it's testing as possible (while trying to save some semblance of semantic version pinning). We own hocon though, eh? Would it be possible to see where 1.0 is on the timeline for hocon and make sure the hocon folks sync up with the Beaker folks?

fpringvaldsen:
@justinstoller No problem! The Puppet Server Jade team maintains the hocon gem (we ported it down from the original Java hocon library so that we could write the hocon module). I'll find out what our plans are for a 1.0 release of the gem and then sync up with the beaker folks.

justinstoller:
Fixed the commit message.

I've tested this locally, however I'm really not sure about the API changes between '0.0.x' and '0.9.x' of hocon. I haven't seen any breakages, but I would appreciate a careful look by those that understand hocon, better.

fpringvaldsen:
@justinstoller There shouldn't have been any API breakages between 0.0.x and 0.9.x. Essentially, a while back we decided we needed to get the hocon gem close to feature parity with the upstream Java HOCON library in order for the hocon module to be where we needed it.

As a result, 0.9.0 features some new APIs, a number of bug fixes, and support for a number of things in the HOCON language spec that the hocon gem was previously unable to parse. The only difference in the previously existing APIs between 0.0.x and 0.9.x is that the parser can now parse more things, so there shouldn't be any breakage there.

fpringvaldsen:
There's an 0.1.0 release as well, which is the same as the 0.9.0 release except there was a minor bug that caused some of the new API functions to not work (we fixed this bug, which was the only change between 0.1.0 and 0.9.0).

anodelman:
Failures unrelated to patch.

PR #821, opened 2015-05-28
(BKR-273) Allow Setting Type by Host

anodelman:
- no longer use 'type' to control path/default host settings
- set environment and defaults post pe/foss installation
- dump more information post testing so that you can correctly pick up
  the host again


lib/beaker/dsl/helpers/host_helpers.rb
```diff
@@ -508,6 +504,19 @@ def create_tmpdir_on(host, path_prefix = '', user=nil)
           end
         end
 
+        # 'echo' the provided value on the given host
+        # @param [Host] host The host to execute the 'echo' on
+        # @param [String] val The string to 'echo' on the host
+        # @return [String] The echo'ed value
+        def echo_on_host host, val
```
colinPL: How is this different than the `echo` method provided by `pswindows/exec.rb`?

anodelman: That is a chunk of code moved from host_prebuilt_steps to its new home in host_helpers.  I'm not keen on altering it as a ride along, but it should be noted and filed as a separate issue.

MikaelSmith: Why is it `echo_on_host` rather than `echo_on`, as other methods are named?

anodelman: I believe to differentiate between it and host.echo.  I still believe that this can be taken on in a separate bug as the issue at hand in this PR is not that code chunk itself (which is just moving from one file to another) but the removal of 'type'.

anodelman:
retest this please

lib/beaker/answers.rb
```diff
@@ -46,7 +80,20 @@ def self.create version, hosts, options
     # @param [String] default Should there be no user value for the provided question name return this default
     # @return [String] The answer value
     def answer_for(options, q, default = nil)
-      options[:answers][q] ? options[:answers][q] : default
+      answer = nil
```
colinPL: This can be `answer = DEFAULT_ANSWERS[q]`. If `DEFAULT_ANSWERS` does not contain `q`, the value will be `nil`. This would not impact setting the value to `default` later in the method.

anodelman:
retest this please

kevpl:
@anodelman needs rebase

anodelman:
retest this please

anodelman:
retest this please

anodelman:
retest this please

anodelman:
retest this please

anodelman:
Failures unrelated to patch.

PR #822, opened 2015-05-28
(maint) Remove step dsl usage from inside Beaker

justinstoller:
Prior to this we would use the `step` DSL method in a couple of places
inside Beaker.  This can be problematic when using Beaker as a library
and requires loading more of Beaker than is necessary for executing
certain helper methods.  This commit removes the internal usage of step
and replaces it with calls to the the logger that will output the same
style of notification that calling step does.


anodelman:
retest this please

anodelman:
retest this please

PR #824, opened 2015-05-29
(BKR-239) Remove dependency on puppet-dashboard rake tasks...

anodelman:
... for shallow gravy.

- do that

### DIFF
--- a/lib/beaker/dsl/install_utils/pe_utils.rb
+++ b/lib/beaker/dsl/install_utils/pe_utils.rb
@@ -390,12 +390,15 @@ module Beaker
               wait_for_host_in_dashboard(host)
             end
 
-            if pre30master
-              task = 'nodegroup:add_all_nodes group=default'
-            else
-              task = 'defaultgroup:ensure_default_group'
+            # only appropriate for pre-3.9 builds
+            if version_is_less(master[:pe_ver], '3.99')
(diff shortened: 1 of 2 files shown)


anodelman:
retest this please

anodelman:
Failures unrelated to patch.

kevpl:
@anodelman don't we need to keep this to work with older versions?

anodelman:
We do not currently have any pipelines using anything pre 3.7 (other than beaker, that does a 3.3.2 -> 3.7.2 upgrade test that is passing).  So there is a risk here, but it is low.  I think that we can clean up breakage in the future, but we need this fix right away.

electrical:
Will this break anything for PE 3.3.x test runs? ( I'm starting to drop 3.1.x and 3.2.x )

anodelman:
I did a successful 3.3.2 install, so there's a high chance of success.  If we just consider this too risky we could wrap the call in a version check?

electrical:
I'm assuming its causing issues for newer PE versions?
We could do either of 2 things.
1. wrap it in a version check
2. do some testing ( acceptance testing ) to see if it breaks testing runs on PE 3.1.x

anodelman:
Wrapped in a version check, backwards compatible.

PR #825, opened 2015-06-01
(BKR-300) Always manage_etc_hosts in cloud-init

mcanevet:
Some cloud-init configuration does not have `manage_etc_host: true` so the VM does not have a valid FQDN which is quite annoying because some services refuse to start without a valid FQDN.

This patch force cloud-init to manage `/etc/host` and thus set a valid fqdn.


anodelman:
This won't get exercised in our testing, so we're just going to hope for the best!

PR #826, opened 2015-06-01
(BKR-292) Cisco platform changes + prepend command functionality

anodelman:
- add cisco to supported platforms
- add ability to send a prepend_cmd string to a command object

### DIFF
--- a/lib/beaker/command_factory.rb
+++ b/lib/beaker/command_factory.rb
@@ -16,7 +16,11 @@ module Beaker
     end
 
     def execute(command, options={}, &block)
-      result = self.exec(Command.new(command), options)
+      cmd_create_options = {}
+      exec_opts = options.dup
+      cmd_create_options[:prepend_cmds] = exec_opts.delete(:prepend_cmds) || nil
+      cmd_create_options[:cmdexe] = exec_opts.delete(:cmdexe) || false
+      result = self.exec(Command.new(command, [], cmd_create_options), exec_opts)
 
       if block_given?
         yield result
--- a/lib/beaker/command.rb
+++ b/lib/beaker/command.rb
@@ -50,6 +50,7 @@ module Beaker
       @args    = args
       @environment = {}
       @cmdexe = @options.delete(:cmdexe) || false
+      @prepend_cmds = @options.delete(:prepend_cmds) || nil
 
       # this is deprecated and will not allow you to use a command line
       # option of `--environment`, please use ENV instead.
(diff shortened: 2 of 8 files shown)


anodelman:
Failures unrelated to patch.

anodelman:
@mikewiebe can you confirm that this handle your test cases?

mikewiebe:
@anodelman yes, I will confirm these changes work with our tests.

mikewiebe:
One small request.  Could we modify the following line in lib/beaker/platform.rb in this commit?

@@ -171,7 +179,7 @@ module Beaker
     def request_terminal_for channel, command
       channel.request_pty do |ch, success|
         if success
-          @logger.info "Allocated a PTY on #{@hostname} for #{command.inspect}"
+          @logger.debug "Allocated a PTY on #{@hostname} for #{command.inspect}"
         else

Change to a debug message so that we only see if if --log-level is set.

lib/beaker/command_factory.rb
```diff
@@ -16,7 +16,11 @@ def assertions
     end
 
     def execute(command, options={}, &block)
-      result = self.exec(Command.new(command), options)
+      cmd_create_options = {}
+      exec_opts = options.dup
+      cmd_create_options[:prepend_cmds] = exec_opts.delete(:prepend_cmds) || nil
```
mikewiebe: I added this comment earlier today but not under this pull request.  Adding it again here.



@anodelman While running my tests I stumbled on a problem. This line is actually modifies the the variable that stores my options outside of this method. The quick irb session below displays the problem.

1) First define a variable 'options' that contains my options including :prepend_cmds

irb(main):001:0> options = {:prepend_cmds => "foo bar baz", :option2 => "option"}
=> {:prepend_cmds=>"foo bar baz", :option2=>"option"}
irb(main):002:0> puts(options)
{:prepend_cmds=>"foo bar baz", :option2=>"option"}
=> nil

2) Now define a quick method to process the options and remove :prepend_cmds if it exists and store in a new var.

irb(main):003:0> def process(options={})
irb(main):004:1> new_opts = {}
irb(main):005:1> new_opts[:prepend_cmds] = options.delete(:prepend_cmds) || nil
irb(main):006:1> end
=> nil

3) Quick test to make sure my options var outside the process method still contains :prepend_cmds

irb(main):007:0> puts(options)
{:prepend_cmds=>"foo bar baz", :option2=>"option"}
=> nil

4) Now call the process method on my options

irb(main):008:0> process(options)
=> "foo bar baz"

5) Unexpectedly (IMO) my variable 'options' outside the method is modified.
irb(main):009:0> puts(options)
{:option2=>"option"}
=> nil

This problem is causing my prepended commands to be stripped in the following case:

opts = {:pty => true, :prepend_cmds => pp(agent)}
agent.install_package(package, '', nil, opts)
agent.check_for_package("puppet", opts)

When I call agent.install_package my opts are intact, but when I then call agent.check_for_package the :prepend_cmds have been removed causing the command to fail.

anodelman: This should no longer be an issue because I've updated the code to dup the options hash so that it remains untouched.

anodelman:
Failures unrelated to patch.

lib/beaker/command_factory.rb
mikewiebe: Excellent.  I have tested the latest diffs and they work for my scenarios.

mikewiebe:
@anodelman Thanks for pulling this in! Appreciate the support here.

PR #827, opened 2015-06-01
(BKR-239) Remove dependency on puppet-dashboard rake tasks for...

anodelman:
... shallow gravy.

- wrap rake test call in version check, only run if pre 3.99

### DIFF
--- a/lib/beaker/dsl/install_utils/pe_utils.rb
+++ b/lib/beaker/dsl/install_utils/pe_utils.rb
@@ -345,7 +345,10 @@ module Beaker
               if host['roles'].include?('frictionless') &&  (! version_is_less(version, '3.2.0'))
                 # If We're *not* running the classic installer, we want
                 # to make sure the master has packages for us.
-                deploy_frictionless_to_master(host)
+                # Only for pre 3.99
+                if version_is_less(version, '3.99')
+                  deploy_frictionless_to_master(host)
+                end
                 on host, installer_cmd(host, opts)
               elsif host['platform'] =~ /osx|eos/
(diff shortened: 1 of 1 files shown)


lib/beaker/dsl/install_utils/pe_utils.rb
```diff
@@ -345,7 +345,10 @@ def do_install hosts, opts = {}
               if host['roles'].include?('frictionless') &&  (! version_is_less(version, '3.2.0'))
                 # If We're *not* running the classic installer, we want
                 # to make sure the master has packages for us.
-                deploy_frictionless_to_master(host)
+                # Only for pre 3.99
+                if version_is_less(version, '3.99')
+                  deploy_frictionless_to_master(host)
```
jpartlow: Using the puppet-dashboard rake tasks to classify a pe_repo platform only works prior to 3.99, but you still need to perform this classification on 4.0.  You just can't use the rake tasks for it.

jpartlow:
Also the deploy_frictionless_to_master() step can be skipped if the agent is the same platform as the master.

lib/beaker/dsl/install_utils/pe_utils.rb
anodelman: @jpartlow Can you give me the correct steps for classification for 4.0?

jpartlow: @anodelman This gets back to the question of Scooter and Beaker brought up in the ticket.  Eric had a possible implementation sketched out in one of the comments.

anodelman:
Closing this out until I've finalized the new workflow.

PR #831, opened 2015-06-03
(BKR-239) Remove dependency on puppet-dashboard rake tasks for...

anodelman:
... shallow gravy.

- wrap rake test call in version check, only run if pre 3.99
- use scooter for new frictionless workflow


jpartlow:
:+1:

anodelman:
Failures unrelated to patch.

PR #832, opened 2015-06-03
(BKR-248) support connecting to SUTs with user/password

anodelman:
- check to see if a custom ssh user has been defined, if so then use
  that instead of the default user per-OS type


kevpl:
@anodelman please add a unit test for this.

anodelman:
Unit tests added.

mikewiebe:
@anodelman Thanks for pulling this in!  Appreciate the support here.

anodelman:
It will be included in beaker 2.14.0 - to be released today or tomorrow.

PR #833, opened 2015-06-03
(BKR-304) Prevent redundant copying when doing scp

thallgren:
Before this commit, the method Beaker::Host.do_scp_to would request
copying of both directories and files from the underlying scp
connection. This resulted in a lot of redundant copying since the
scp connection, when handed a directory, would copy it recursively.

This commit ensures that only the files are copied. This is safe
because all remote directories are already created in a preceding
step.


anodelman:
Looks good.  Thanks for the PR!

PR #836, opened 2015-06-03
[BKR-301] add warning when empty fog file is encountered

rick:
When beaker's vmpooler is provided with an empty `~/.fog` file, a failure with a cryptic message has been the result.  This changes that behavior to be more user-friendly (continuing without authentication, posting a friendly warning); as well as improving test coverage in this area, and making other failures more user-friendly as well.

/cc @anodelman @kevpl

### DIFF
--- a/lib/beaker/hypervisor/vmpooler.rb
+++ b/lib/beaker/hypervisor/vmpooler.rb
@@ -28,7 +28,11 @@ module Beaker
       creds = {}
 
       if fog = read_fog_file(dot_fog)
-        creds[:vmpooler_token] = fog[:default][:vmpooler_token]
+        if fog[:default] && fog[:default][:vmpooler_token]
+          creds[:vmpooler_token] = fog[:default][:vmpooler_token]
+        else
+          @logger.warn "Credentials file (#{dot_fog}) is missing a :default section with a :vmpooler_token value; proceeding without authentication"
+        end
       else
         @logger.warn "Credentials file (#{dot_fog}) is empty; proceeding without authentication"
       end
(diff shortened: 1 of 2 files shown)


lib/beaker/hypervisor/vmpooler.rb
```diff
@@ -15,6 +15,8 @@ class Vmpooler < Beaker::Hypervisor
       Errno::ENETUNREACH,
     ]
 
+    attr_reader :options, :logger, :hosts, :credentials
```
anodelman: any reason that you need these when those variables are already stored as @options/@logger/etc?

rick: It's documented in the commit message: the main reason is that I prefer not to reach into an object's instance variables during testing if at all possible, that brings me to adding `:credentials`. 

If I'm going to do that I'd prefer to have them all there (the question "why is A an attribute but not B?" isn't usually best answered by "because we only tested A").

The other reason isn't documented, but, using accessors (`:logger`) instead of instance variables (`@logger`) even inside the class that owns them, ends up being less brittle, reduces coupling, and makes replacing access techniques easier; it also promotes easier testability.  I didn't make the move to changing those internal instance variable usages to calls to the accessors, but it wouldn't be out of hand.

The two arguments I can think of against doing this are: (1) there's something about these attributes (even as readers) that makes it a bad idea to expose them to the outside, and (2) [Liskov](http://en.wikipedia.org/wiki/Liskov_substitution_principle) might object.  The question might be: do we want to make this a uniformly exposed interface on any of the other classes in the family?

anodelman:
retest this please

rick:
retest this please

anodelman:
Might as well get this merged so that it can make it into beaker 2.14.0.

PR #839, opened 2015-06-04
(BKR-311) Allow AWS IP Fallback

vindir:
When bringing up an instance in a VPC there won't always be a public address.  This allows fallback to assign the private ip to host['ip'] for instances without a public address.

When the hypervisor is set to ec2, beaker can't currently connect to an instance that we're provisioning unless we have provisioned first then set a R53 public ip for subsequent runs.  This change would allow users inside a VPC to achieve complete end-to-end beaker runs from either a bastion or across whatever VPN solution they have set up.

Either insight on anything we might be missing or a merge would both be much appreciated!  Thanks :)

### DIFF
--- a/lib/beaker/hypervisor/aws_sdk.rb
+++ b/lib/beaker/hypervisor/aws_sdk.rb
@@ -473,7 +473,7 @@ module Beaker
       @hosts.each do |host|
         @logger.notify("aws-sdk: Populate DNS for #{host.name}")
         instance = host['instance']
-        host['ip'] = instance.ip_address
+        host['ip'] = instance.ip_address ? instance.ip_address : instance.private_ip_address
         host['private_ip'] = instance.private_ip_address
         host['dns_name'] = instance.dns_name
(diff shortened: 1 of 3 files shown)


kevpl:
Hi @vindir,

Thanks for submitting this patch!  I see that you pre-pended the issue to the title of the PR, but for us to merge this, the issue has to also be pre-pended to the git commit message.  Would you please add a quick unit test for the `populate_dns` method, so that we can quickly verify if we break this functionality in the future as well?

Thanks! We look forward to getting this PR through the process and merged soon.
Kevin

anodelman:
ok to test

anodelman:
Failure unrelated to patch.  Just need the commit message updated and this can be merged.

vindir:
Hey all, I went ahead and added tests for the populate_dns method.  It includes a new make_instance method in the helpers that should make it easier to test the other currently unvalidated methods in the AWS hypervisor.

Also rebased everything back up to date with master.

Could you guys test it again and give it a :shipit: if it's all good to go?

kevpl:

vindir:
:clap: Looks like the jenkins build is fixed :+1:

PR #844, opened 2015-06-09
(BKR-53) with_puppet_running_on only restarts before yielding

kevpl:
Before, with_puppet_running_on bounced the puppet service before and after runs.
This was causing us a lot of pain as the last restart proved unnecessary for the
most part, and added 30 seconds to each test run on avg in our CI.

We've decided to improve this by removing the second bounce, since most tests using
puppet should already use this method, so the next test will ensure its wanted state
prior to yielding.

If people still want the old behavior, they can pass in `:restart_when_done` as a boolean
flag in the `conf_opts` argument of this method

### DIFF
--- a/lib/beaker/dsl/helpers/puppet_helpers.rb
+++ b/lib/beaker/dsl/helpers/puppet_helpers.rb
@@ -60,6 +60,14 @@ module Beaker
         # 6. Revert Puppet to the pre-test state
         # 7. Testing artifacts are saved in a folder named for the test
         #
+        # @note Whether Puppet is started or restarted depends on what kind of
+        #   server you're running.  Passenger and puppetserver are restarted before.
+        #   Webrick is started before and stopped after yielding, unless you're using
+        #   service scripts, then it'll behave like passenger & puppetserver.
+        #   Passenger and puppetserver (or webrick using service scripts)
+        #   restart after yielding by default.  You can stop this from happening
+        #   by setting the :restart_when_done flag of the conf_opts argument.
+        #
         # @param [Host] host        One object that act like Host
         #
         # @param [Hash{Symbol=>String}] conf_opts  Represents puppet settings.
@@ -85,6 +93,13 @@ module Beaker
         #                            even if it seems the host has passenger.
         #                            This is needed in FOSS tests to initialize
         #                            SSL.
+        # @option conf_opts [Boolean] :restart_when_done  determines whether a restart
+        #                            should be run after the test has been yielded to.
+        #                            Will stop puppet if false. Default behavior
+        #                            is to restart, but you can override this on the
+        #                            host or with this option.
+        #                            (Note: only works for passenger & puppetserver
+        #                            masters (or webrick using the service scripts))
         # @param [File] testdir      The temporary directory which will hold backup
         #                            configuration, and other test artifacts.
         #
@@ -110,7 +125,10 @@ module Beaker
           raise(ArgumentError, "with_puppet_running_on's conf_opts must be a Hash. You provided a #{conf_opts.class}: '#{conf_opts}'") if !conf_opts.kind_of?(Hash)
           cmdline_args = conf_opts[:__commandline_args__]
           service_args = conf_opts[:__service_args__] || {}
-          conf_opts = conf_opts.reject { |k,v| [:__commandline_args__, :__service_args__].include?(k) }
+          restart_when_done = true
+          restart_when_done = host[:restart_when_done] if host[:restart_when_done]
+          restart_when_done = conf_opts.fetch(:restart_when_done, restart_when_done)
+          conf_opts = conf_opts.reject { |k,v| [:__commandline_args__, :__service_args__, :restart_when_done].include?(k) }
 
           curl_retries = host['master-start-curl-retries'] || options['master-start-curl-retries']
           logger.debug "Setting curl retries to #{curl_retries}"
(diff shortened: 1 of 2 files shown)


lib/beaker/dsl/helpers/puppet_helpers.rb
```diff
@@ -163,7 +176,7 @@ def with_puppet_running_on host, conf_opts, testdir = host.tmpdir(File.basename(
 
               if host.use_service_scripts? && !service_args[:bypass_service_script]
                 restore_puppet_conf_from_backup( host, backup_file )
-                bounce_service( host, host['puppetservice'], curl_retries )
+                bounce_service( host, host['puppetservice'], curl_retries ) if restart_when_done
```
joshcooper: I think we still want to stop the service when `restart_when_done` is false.

kevpl:

spec/beaker/dsl/helpers/puppet_helpers_spec.rb
```diff
@@ -691,6 +751,18 @@ def stub_post_setup
             subject.with_puppet_running_on(host, {:__commandline_args__ => '--with arg'})
             expect(host).to execute_commands_matching(/^puppet master --with arg/).once
           end
+
+          it 'is not effected by the :restart_when_done flag' do
```
waynr: nit: "affected"

spec/beaker/dsl/helpers/puppet_helpers_spec.rb
```diff
@@ -581,9 +581,9 @@ def stub_post_setup
         let(:is_pe) { true }
         let(:service_restart) { true }
 
-        it 'bounces puppet twice' do
+        it 'starts puppet once, stops it twice' do
           subject.with_puppet_running_on(host, {})
-          expect(host).to execute_commands_matching(/puppet resource service #{host['puppetservice']}.*ensure=running/).exactly(2).times
+          expect(host).to execute_commands_matching(/puppet resource service #{host['puppetservice']}.*ensure=running/).once
```
waynr: I believe that changing this unit test means you are breaking API compatibility with previous versions of Beaker--so is this something that should wait for the next major version of Beaker?

spec/beaker/dsl/helpers/puppet_helpers_spec.rb
```diff
@@ -597,9 +597,30 @@ def stub_post_setup
               execution += 1
             end
           end.to change { execution }.by(1)
-          expect(host).to execute_commands_matching(/puppet resource service #{host['puppetservice']}.*ensure=running/).exactly(2).times
```
waynr: My previous comment applies here as well.

lib/beaker/dsl/helpers/puppet_helpers.rb
```diff
@@ -110,7 +123,8 @@ def with_puppet_running_on host, conf_opts, testdir = host.tmpdir(File.basename(
           raise(ArgumentError, "with_puppet_running_on's conf_opts must be a Hash. You provided a #{conf_opts.class}: '#{conf_opts}'") if !conf_opts.kind_of?(Hash)
           cmdline_args = conf_opts[:__commandline_args__]
           service_args = conf_opts[:__service_args__] || {}
-          conf_opts = conf_opts.reject { |k,v| [:__commandline_args__, :__service_args__].include?(k) }
+          restart_when_done = conf_opts.fetch(:restart_when_done, false)
```
waynr: I think the default behavior should be `true` since that is how puppetserver currently works.

spec/beaker/dsl/helpers/puppet_helpers_spec.rb
```diff
@@ -607,10 +628,10 @@ def stub_post_setup
           host.uses_passenger!
         end
 
-        it 'bounces puppet twice' do
+        it 'bounces puppet once' do
           allow( subject ).to receive(:curl_with_retries)
           subject.with_puppet_running_on(host, {})
-          expect(host).to execute_commands_matching(/apache2ctl graceful/).exactly(2).times
+          expect(host).to execute_commands_matching(/apache2ctl graceful/).once
```
waynr: Previous comment applies here as well.

spec/beaker/dsl/helpers/puppet_helpers_spec.rb
```diff
@@ -622,17 +643,36 @@ def stub_post_setup
               execution += 1
             end
           end.to change { execution }.by(1)
-          expect(host).to execute_commands_matching(/apache2ctl graceful/).exactly(2).times
```
waynr: Previous comment applies here as well.

lib/beaker/dsl/helpers/puppet_helpers.rb
waynr: Derp, I meant "since that is how `with_puppet_running_on` currently works"

joshcooper: :+1: to preserving beaker's current behavior by default. That said I would like to not have to change all 59 foss tests that use with_puppet_running_on. Can we default to true, merge whatever value is set on the `host`, and merge whatever value is passed as a parameter. That way the last one takes precedence?

spec/beaker/dsl/helpers/puppet_helpers_spec.rb
anodelman: Since the Beaker DSL method will still correctly apply_manifest_on and we are only changing the internals of the method itself this does not require a major version bump.

waynr:
:+1: Looks good to me once tests pass.

anodelman:
retest this please

lib/beaker/dsl/helpers/puppet_helpers.rb
```diff
@@ -60,6 +60,14 @@ def puppet_group(host)
         # 6. Revert Puppet to the pre-test state
         # 7. Testing artifacts are saved in a folder named for the test
         #
+        # @note Whether Puppet is started or restarted depends on what kind of
+        #   server you're running.  Passenger and puppetserver are restarted before.
+        #   Webrick is started before and stopped after yielding, unless you're using
+        #   service scripts, then it'll behave like passenger & puppetserver.  You
+        #   can make passenger and puppetserver (or webrick using service scripts)
+        #   restart after yielding by setting the :restart_when_done flag of the
```
waynr: This wording makes it sound like the default behavior is for the server to stop when the default is actually to restart.

waynr:
:+1:

PR #846, opened 2015-06-09
(BKR-317) More usable configure_foss_defaults_on

anodelman:
- create configure_defaults_on that provided a type sets the correct
  defaults per-host, will also remove any existing defaults

### DIFF
--- a/lib/beaker/dsl/install_utils/foss_defaults.rb
+++ b/lib/beaker/dsl/install_utils/foss_defaults.rb
@@ -128,7 +128,7 @@ module Beaker
         # @param [Host] host    A single host to act upon
         # @param [String] platform The platform type of this host, one of windows, pswindows, freebsd, mac & unix
         def remove_platform_foss_defaults(host, platform)
-          PE_DEFAULTS[platform].each_pair do |key, val|
+          FOSS_DEFAULTS[platform].each_pair do |key, val|
             host.delete(key)
           end
           host['group'] = nil
--- a/lib/beaker/dsl/install_utils/pe_defaults.rb
+++ b/lib/beaker/dsl/install_utils/pe_defaults.rb
@@ -110,7 +110,7 @@ module Beaker
         # @param [Host] host    A single host to act upon
         # @param [String] platform The platform type of this host, one of windows, freebsd, mac & unix
         def remove_platform_pe_defaults(host, platform)
-          FOSS_DEFAULTS[platform].each_pair do |key, val|
+          PE_DEFAULTS[platform].each_pair do |key, val|
             host.delete(key)
           end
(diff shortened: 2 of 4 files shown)


MikaelSmith:
`install_puppet_on` doesn't configure the type, so I think you can end up in a state where things are broken. If you do `install_puppet_on host, 4.0.0` while `type` is set to `'foss'`, it will install `puppet-agent`, then configure paths for `foss` (old paths). This is what would happen if you had `puppet-agent` installed and configured, then uninstalled it and only did `install_puppet_on`.

anodelman:
install_puppet_on calls a bunch of sub methods (like install_puppet_from_rpm) which in turn call configure_foss_defaults_on - so the type should be correctly set post installation.

MikaelSmith:
[`configure_foss_defaults_on`](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/install_utils/foss_utils.rb#L41) changes behavior based on `type`, but doesn't update it. I don't see anywhere in the call chains for `install_puppet_on` that it would update the node's `type`.

MikaelSmith:
Other than that, appears to work for my use case.

anodelman:
It's a little deep in the chain, but  add_platform_foss_defaults will be called and will set type='foss' (unless the host is already identified as 'aio' and then it will leave it alone).

anodelman:
Failures unrelated to patch.

MikaelSmith:
That is the workflow though...
- upgrade to puppet-agent
- `configure_defaults_on host, 'aio'`, which will call `add_aio_defaults_on` and set `type='aio'`
- uninstall puppet-agent
- `install_puppet_on`, which will install (foss) puppet while type is 'aio'.

I think that's an API problem, but if it can't be fixed then I can do `configure_defaults_on host, 'foss'` before repeating `install_puppet_on`.

anodelman:
It sounds like you need something as part of the "uninstall puppet-agent" step that could remove all current defaults?  Then the install_puppet_on would run clean and assume 'foss' type?

MikaelSmith:
Yeah, that would work.

anodelman:
I'll update this patch to cover that workflow.

MikaelSmith:
If I use `configure_defaults_on host, 'foss'` after the "uninstall puppet-agent" step, everything works. So I'm ok with this API.

I still wonder about being able to do `install_puppet_on` and end up with the wrong configuration - i.e. configure a node with `type='aio'` then do `install_puppet_on host`. You'll end up with puppet 3.8 installed but paths setup for aio. I suppose that's the beaker users problem.

anodelman:
Failure unrelated to patch.

PR #847, opened 2015-06-11
(BKR-322) --fail-mode fast doesn't seem to be failing fast

anodelman:
- was using the default value instead of the setting value

### DIFF
--- a/lib/beaker/test_suite.rb
+++ b/lib/beaker/test_suite.rb
@@ -243,7 +243,7 @@ module Beaker
     #@option options [String] :xml_stylesheet The path to a stylesheet to be applied to the generated XML output
     #@param [Symbol] fail_mode One of :slow, :fast
     #@param [Time] timestamp Beaker execution start time
-    def initialize(name, hosts, options, timestamp, fail_mode = :slow)
+    def initialize(name, hosts, options, timestamp, fail_mode=nil)
       @logger     = options[:logger]
       @test_cases = []
       @test_files = options[name]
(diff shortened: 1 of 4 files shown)


lib/beaker/test_suite.rb
```diff
@@ -251,7 +251,7 @@ def initialize(name, hosts, options, timestamp, fail_mode = :slow)
       @hosts      = hosts
       @run        = false
       @options    = options
-      @fail_mode  = fail_mode || @options[:fail_mode]
+      @fail_mode  = fail_mode || @options[:fail_mode] || :slow
```
rick: :+1:  Good catch.

spec/beaker/test_suite_spec.rb
```diff
@@ -24,6 +24,27 @@ module Beaker
         expect(tfs).to include rb_test
       end
 
+      it 'defaults to :slow fail_mode is not provided through parameter or options' do
```
rick: Maybe a missing `if` in the description, or something like that.

spec/beaker/test_suite_spec.rb
```diff
@@ -24,6 +24,27 @@ module Beaker
...
+        @files = [ rb_test ]
+        ts = Beaker::TestSuite.new('name', 'hosts', options, Time.now)
+        tfm = ts.instance_variable_get(:@fail_mode)
```
rick: This is one of those cases where I prefer to expose an attribute for the options the instance was configured with, and test via that route instead of via `instance_variable_get`, but definitely don't consider this a blocker, or even a change request, really.

rick:
Looks good to me! :tada:

anodelman:
@rick I think that we should revisit when it is appropriate to provide an accessor to class variables, for now these new spec tests behave the same way as the rest of the tests written for this object.

anodelman:
retest this please

anodelman:
retest this please

anodelman:
Failure unrelated to patch.

anodelman:
@kevpl This should be ready to merge.

PR #848, opened 2015-06-12
(BKR-326) Fixes enabling user environments

petems:
Currently this step uses `echo_to_file`, which actually overwrites the file with the one line. This changes it to use the perl regex instead.

### DIFF
--- a/lib/beaker/host_prebuilt_steps.rb
+++ b/lib/beaker/host_prebuilt_steps.rb
@@ -521,7 +521,7 @@ module Beaker
           host.exec(Command.new("stopsrc -g ssh"))
           host.exec(Command.new("startsrc -g ssh"))
         when /freebsd/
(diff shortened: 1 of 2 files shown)


anodelman:
retest this please

lib/beaker/host_prebuilt_steps.rb
```diff
@@ -13,7 +13,7 @@ module HostPrebuiltSteps
     SLEEPWAIT = 5
     TRIES = 5
     UNIX_PACKAGES = ['curl', 'ntpdate']
-    FREEBSD_PACKAGES = ['curl']
+    FREEBSD_PACKAGES = ['curl', 'perl5']
```
petems: Not sure how neccesary this is, as perl is included by default on all bsd builds apparently, but cant hurt (only takes a few secs to check it)

anodelman:
Failure unrelated to patch.

PR #849, opened 2015-06-12
(BKR-326) Fixes FreeBSD install to use ports

petems:


anodelman:
retest this please

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -560,6 +562,27 @@ def install_a_puppet_msi_on(host, opts)
...
 
+        # Installs Puppet and dependencies from FreeBSD ports
+        #
+        # @param [Host] host The host to install packages on
+        # @param [Hash{Symbol=>String}] opts An options hash
+        # @option opts [String] :version The version of Puppet to install (shows warning)
+        #
+        # @return nil
+        # @api private
+        def install_puppet_from_freebsd_ports_on( host, opts )
```
anodelman: Using the _on suffix in the host name means that you should be able to pass a single host or an array of hosts - please use the block_on method to cycle through those.

petems: Does adding the alias method fix that?

anodelman: No, as this is an _on method that doesn't work correctly with an array of hosts.

anodelman:
Last review comment still current.

anodelman:
Will close out next week due to inactivity.

lib/beaker/dsl/install_utils/foss_utils.rb
petems: But none of the other `_on` methods have that ability? This method has the same implementation as `install_puppet_from_deb_on` which gets the host given to it from the `block_on` block on https://github.com/petems/beaker/blob/BKR-326-fix_freebsd_install/lib/beaker/dsl/install_utils/foss_utils.rb#L231

Am I missunderstanding something? I can change the method, but it seems to be the same logic as the rest of the methods.

anodelman:
Closing due to inactivity.

PR #853, opened 2015-06-15
(BKR-319) added puppet-agent install for install_pe for versions over 4.0

kevpl:


kevpl:
failure unrelated to change

lib/beaker/answers/version40.rb
```diff
@@ -34,7 +34,7 @@ def generate_answers
       install_update_server = answer_for(@options, :q_install_update_server, 'y')
 
       the_answers.map do |key, value|
-        the_answers[key][:q_update_server_host] = update_server_host
+        the_answers[key][:q_update_server_host] = update_server_host if the_answers[key]
```
anodelman: In what case is this answer empty?

lib/beaker/dsl/install_utils/pe_utils.rb
```diff
@@ -361,9 +361,30 @@ def do_install hosts, opts = {}
             pre30master = version_is_less(opts[:pe_ver] || master['pe_ver'], '3.0')
           end
 
+          hosts_agent_only = []
+          hosts_not_agent_only = []
+          agent_only_check_needed = !version_is_less(opts.fetch('pe_ver', '3.8'), '3.99')
+          unless agent_only_check_needed
+            hosts.each do |host|
+              if host['pe_ver'] && !version_is_less(host['pe_ver'], '3.99')
```
anodelman: Since you are already cycling through the hosts here it would be possible to construct the arrays of agent-only/not-agent only hosts here.  Also, should probably go into it's own private method to stop do_install from becoming any more unmanageable.

lib/beaker/dsl/install_utils/pe_utils.rb
```diff
@@ -483,6 +503,22 @@ def do_install hosts, opts = {}
...
         end
 
+        # Helper for setting up pe_defaults & setting up the cert on the host
+        # @param [Host] host                            host to setup
+        # @param [Host] master                          the master host, for setting up the relationship
+        # @param [Array<Fixnum>] acceptable_exit_codes  The exit codes that we want to ignore
+        #
+        # @return nil
+        # @api private
+        def setup_defaults_and_config_helper_on(host, master, acceptable_exit_codes=nil)
```
anodelman: This used to run on only osx/eos - is it applicable to windows?  Does acceptable_exit_codes => nil work correctly?  Or does that indicate no acceptable exit codes?

lib/beaker/answers/version40.rb
kevpl: Following the pattern from the [top of the `pe_utils_spec.rb`](https://github.com/puppetlabs/beaker/blob/master/spec/beaker/dsl/install_utils/pe_utils_spec.rb#L21-L28) file, I had to [create 4.0 hosts](https://github.com/kevpl/beaker/blob/bkr319_installpe_puppetagent/spec/beaker/dsl/install_utils/pe_utils_spec.rb#L418-L424) for my testing.  Doing this results in that value being empty.

lib/beaker/dsl/install_utils/pe_utils.rb
kevpl: It now only runs on osx/eos & aio.  I figured that since we're duplicating, better to pull it out.

Yes, `acceptable_exit_codes` works correctly.  In the [`host.exec` line it's used in](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host.rb#L272), it's checked, and a default is used if it's falsy, which it will be when set to `nil`.

The default value when the `acceptable_exit_codes` is falsy is `[0, nil]`, and has been for some time, so I doubt that would be a problem.

kevpl:
@anodelman I've answered your questions, and included the changes for your other comment.  This should be ready for another look (better once testing results are complete).

kevpl:
:point_up: :heart_eyes:

lib/beaker/dsl/install_utils/pe_utils.rb
```diff
@@ -361,9 +361,12 @@ def do_install hosts, opts = {}
             pre30master = version_is_less(opts[:pe_ver] || master['pe_ver'], '3.0')
           end
 
+          agent_only_check_needed, hosts_agent_only, hosts_not_agent_only = \
+            create_hosts_agent_only_specified_arrays_helper(hosts, opts['pe_ver'])
```
anodelman: opts['pe_ver'] could be null if the user is setting the pe_ver per-host instead of globally - then I don't believe that the logic here works correctly.

lib/beaker/dsl/install_utils/pe_utils.rb
```diff
@@ -483,6 +485,63 @@ def do_install hosts, opts = {}
...
+        # @param [Array<Host>]  hosts   hosts to split up into the arrays
+        # @param [String]       pe_ver  overall pe_ver option, will be used if
+        #                               set rather than the hosts pe_ver
+        #
+        # @api private
+        # @return [Boolean, Array<Host>, Array<Host>] whether a version (overall
+        #   overall or on a host) is over 3.99, so agent_only is possible;
+        #   the array of hosts to do an agent_only install on;
+        #   the array of hosts to do our usual install methods on
+        def create_hosts_agent_only_specified_arrays_helper(hosts, pe_ver)
```
anodelman: Nit: every method is a helper, so you probably don't need to call it a helper in the name.  So 'create_hosts_agent_only_specified_arrays'?

lib/beaker/dsl/install_utils/pe_utils.rb
kevpl: I tested this case during development, it does work. Here's a small example from IRB:

    >> def test(val1)
    >> val1 ||= 7
    >> val1
    >> end
    nil
    >> test(nil)
    7

lib/beaker/dsl/install_utils/pe_utils.rb
kevpl: this is an install utils file, not a helpers file. Between this and the `@api private`, I was trying to denote that this shouldn't be called by a user or from the DSL.

Let me know if you still want me to remove that, and I can do that.

anodelman: Yes, I would prefer a somewhat more streamlined name.

lib/beaker/dsl/install_utils/pe_utils.rb
anodelman: I don't doubt the setting of the default value, I just find the logic of this confusing.  You pass in the global, default it to 3.8 and then compare it to 3.99 - and then do a check on individual hosts pe versions.  Maybe this could be streamlined by just generating an array of all possible pe_ver (global and host) and checking to see if the largest is > 3.99?

pe_versions =( [] << opts['pe_ver'] << hosts.map{ |host| host['pe_ver']}).compact.flatten

Otherwise the steps are spread out in a weird way that makes the code hard to follow.

kevpl:

kevpl:

kevpl:

kevpl:

kevpl:
failure unrelated to change

anodelman:
retest this please

kevpl:

kevpl:

kevpl:
failures unrelated to change

PR #854, opened 2015-06-16
(BKR-330) (DONT MERGE) Would code like this work?

petems:
Trying to figure out an easier way of people managing hosts rather than having to add a host each time, and potentially break things.

Would code like this work?


anodelman:
I think that lazy loading of host platforms would be a reasonable path here, much like hypervisors currently work.

petems:
Moving into new branch :+1:

PR #855, opened 2015-06-17
(BKR-339) Remove the target module directory on host during copy_module_to

bstopp:
Without this patch, the `copy_module_to` function will not work correctly to copy updated module files into the target directory on the hosts. Acceptance test spec files can be updated, but not module sources. Additionally the module copy will fail on the second non-provisioning run.

The patch simply executes a delete of the target directory on the host, ensuring it does not exist prior to the mv command.

### DIFF
--- a/lib/beaker/dsl/install_utils/module_utils.rb
+++ b/lib/beaker/dsl/install_utils/module_utils.rb
@@ -131,13 +131,7 @@ module Beaker
               scp_to host, source_path, target_module_dir, {:ignore => ignore_list}
               #rename to the selected module name, if not correct
               cur_path = File.join(target_module_dir, source_name)
-              if (cur_path != target_path)
-                if host.is_powershell?
-                  on host, "move /y #{cur_path} #{target_path}"
-                else
-                  on host, "mv #{cur_path} #{target_path}"
-                end
-              end
+              host.mv cur_path, target_path unless cur_path == target_path
             when 'rsync'
               logger.debug "Using rsync to transfer #{source_path} to #{target_path}"
               rsync_to host, source_path, target_path, {:ignore => ignore_list}
--- a/lib/beaker/host/pswindows/exec.rb
+++ b/lib/beaker/host/pswindows/exec.rb
@@ -21,6 +21,15 @@ module PSWindows::Exec
     execute("del /s /q #{path}")
   end
 
+  # Move the origin to destination. The destination is removed prior to moving.
+  # @param [String] orig The origin path
+  # @param [String] dest the destination path
+  # @param [Boolean] rm Remove the destination prior to move
+  def mv(orig, dest, rm=true)
+    rm_rf dest unless !rm
+    execute("move /y #{orig} #{dest}")
+  end
+
   def path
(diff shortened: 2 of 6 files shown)


lib/beaker/dsl/install_utils/module_utils.rb
```diff
@@ -133,8 +133,10 @@ def copy_module_to(one_or_more_hosts, opts = {})
               cur_path = File.join(target_module_dir, source_name)
               if (cur_path != target_path)
                 if host.is_powershell?
+                  on host, "rm -Recurse -Force #{target_path}"
```
anodelman: You'll want something cmd friendly here along the lines of:

RMDIR [/S] [/Q] [drive:]path

bstopp:
Updated.

anodelman:
ok to test

bstopp:
There any way for me to see the failures?

anodelman:
retest this please

anodelman:
If the test failures are related to the patch I'll post them in a comment - unfortunately our Jenkins is currently internal only.

anodelman:
Failures unrelated to patch.

lib/beaker/dsl/install_utils/module_utils.rb
```diff
@@ -133,8 +133,10 @@ def copy_module_to(one_or_more_hosts, opts = {})
               cur_path = File.join(target_module_dir, source_name)
               if (cur_path != target_path)
                 if host.is_powershell?
+                  on host, "rmdir /S /Q #{target_path}"
```
anodelman: Looking through the beaker host object, you can just do a beaker.rm_rm(target_path) and that will do the right thing per-OS.

lib/beaker/dsl/install_utils/module_utils.rb
```diff
@@ -133,8 +133,10 @@ def copy_module_to(one_or_more_hosts, opts = {})
...
                   on host, "move /y #{cur_path} #{target_path}"
                 else
+                  on host, "rm -Rf #{target_path}"
                   on host, "mv #{cur_path} #{target_path}"
```
anodelman: This should also be put into the host abstraction so that host.mv(cur_path, target_path) works.  Then this code will not need an OS-dependent if statement and will be easier to read/maintain.

bstopp:
Ok, let me take a look. I may have an update pretty quick then.

bstopp:
Ok, I screwed something up with the PR and my branch. I'll need some time to fix whatever it was i did to show all the merges from the upstream.

In the mean time, are the changes to the exec.rb for Unix/PSWindows and copy_module_to what you were wanting?

bstopp:
Ok, this should be correct now. Please let me know if this was direction you wanted the changes to take.

anodelman:
Thanks for your continued work on this!  Just waiting on testing.

bstopp:
Just FYI; i used the updated gem in my local testing and it's working as expected with a box that is *not* provisioned every run.

anodelman:
\o/

anodelman:
Failures unrelated to patch.

PR #856, opened 2015-06-23
(BKR-336) added command-line arg to get test results in timing order

kevpl:
The new argument will allow a second beaker_*.xml file to be
created next to the first one in timing order.


kevpl:
failure unrelated to change

lib/beaker/test_suite.rb
```diff
@@ -310,6 +311,7 @@ def run
       # has not studied the implementation in detail. --daniel 2011-03-14
       @test_suite_results.summarize( Logger.new(log_path("#{name}-summary.txt", @options[:log_dated_dir]), STDOUT) )
       @test_suite_results.write_junit_xml( log_path(@options[:xml_file], @options[:xml_dated_dir]) )
+      @test_suite_results.write_junit_xml( log_path(@options[:xml_time], @options[:xml_dated_dir]), True ) if @options[:xml_time_enabled]
```
waynr: Why not just apply this option to the existing xml file?

kevpl: Because @anodelman let me know that there is interest in preserving runtime order, so we can't just do one or the other, we have to have a way to do both.

waynr: Fair enough

lib/beaker/test_suite.rb
```diff
@@ -172,6 +172,7 @@ def write_junit_xml(xml_file)
             end
             suite.add_child(properties)
 
+            @test_cases.sort! { |x,y| y.runtime <=> x.runtime } if time_sort
```
waynr: I am vaguely concerned about an in-place sort for global attribute that may have other uses in the future that are affected by the order.

Also, it's not clear from the `write_junit_xml` method name or the `time_sort` parameter that it will modify the `test_cases` ordering.

anodelman:
Is it not possible to make the junit formatted output sortable by time?  It would be nice if all this information was displayed on the same page and not by selection.

lib/beaker/test_suite.rb
kevpl: good point.  I was doing this to make diff size smaller, but I'll change it soon.

waynr:
@anodelman it depends on what you are using to view the junit output. In Jenkins there is limited sorting capability, but viewing in Firefox v31 there is no sorting at all.

kevpl:
I doubt that it's not possible, it's just not efficient for our time, as opposed to the solution in this PR.  It takes a file which is on avg 3MB on our Jenkins, and doubles it, but only if you want that information and indicate it by setting the flag.

When you say by selection, it sounds like you can only have one or the other.  You can have both at the same time.  Maybe I'm not understanding that objection?

anodelman:
Could we at least update the junit template to interlink between the two files.

lib/beaker/test_suite.rb
```diff
@@ -172,7 +172,9 @@ def write_junit_xml(xml_file)
             end
             suite.add_child(properties)
 
-            @test_cases.each do |test|
+            test_cases_to_report = @test_cases
+            test_cases_to_report = @test_cases.sort { |x,y| y.runtime <=> x.runtime } if time_sort
```
waynr: Should this be in-place sorting `test_cases_to_report`?

lib/beaker/test_suite.rb
```diff
@@ -172,7 +172,9 @@ def write_junit_xml(xml_file)
             end
             suite.add_child(properties)
 
-            @test_cases.each do |test|
+            test_cases_to_report = @test_cases
```
waynr: It's not clear to me if this is creating a reference to `@test_cases` or actually creating a copy of that data structure.

Could you add a spec test to validate that the order of test_cases is not affected by the `time_sort` option?

waynr: Or just validate this in an existing test case.

lib/beaker/test_suite.rb
anodelman: Should be 'true', not 'True'.

anodelman:
I believe that we are just waiting here on the interlink between test-results-sorted and test-results-run-order.

kevpl:
interlink between files added, as well as the test for validating order consistency

lib/beaker/test_suite.rb
kevpl: I've added a test for this

lib/beaker/test_suite.rb
kevpl: @waynr why would that be better?  The way that `test_cases_to_report` is currently setup, that would in-place sort `@test_cases` as well, which is why we've done this, to make sure we don't sort that member variable.

The logic for the 2 above lines is that if we're not sorting based on time, then we just use a reference to `@test_cases`, because everything past these lines is read-only, so there's no worry about munging that up.  If time sorting is required, then we want to make a copy, and have that copy be sorted, which is what this line is doing.

Let me know if any of this is unclear, or if you have any questions about it.

Thanks,
Kevin

waynr: Oh, I was thinking that `@test_cases.sort` was an in-place sort but it actually returns a new sorted array. (Basically I was confusing Ruby's `Array.sort` with `Array.sort!`)

waynr:
:+1: Looks good to me!

anodelman:
Tested fine locally.

PR #858, opened 2015-06-23
(BKR-347) Openstack: Add support for cinder attached volumes

spjmurray:
Implement the storage array feature introduced for vagrant virtual box
to OpenStack users.  Unlike virtual box the bus type is outside of our
control.  If you need SCSI support this must be attached to the image
in glance.  How to do this is in the method documentation

### DIFF
--- a/lib/beaker/hypervisor/openstack.rb
+++ b/lib/beaker/hypervisor/openstack.rb
@@ -59,9 +59,9 @@ module Beaker
       @network_client ||= Fog::Network.new(networkoptionhash)
 
       if not @network_client
-
         raise "Unable to create OpenStack Network instance (api_key: #{@options[:openstack_api_key]}, username: #{@options[:openstack_username]}, auth_url: #{@options[:openstack_auth_url]}, tenant: #{@options[:openstack_tenant]})"
-                                                                                                                                      end
+      end
+
     end
 
     #Provided a flavor name return the OpenStack id for that flavor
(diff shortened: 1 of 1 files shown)


lib/beaker/hypervisor/openstack.rb
```diff
@@ -59,9 +59,23 @@ def initialize(openstack_hosts, options)
...
+      volumeoptionhash[:provider]           = :openstack
+      volumeoptionhash[:openstack_api_key]  = @options[:openstack_api_key]
+      volumeoptionhash[:openstack_username] = @options[:openstack_username]
+      volumeoptionhash[:openstack_auth_url] = @options[:openstack_auth_url]
+      volumeoptionhash[:openstack_tenant]   = @options[:openstack_tenant]
+      volumeoptionhash[:openstack_region]   = @options[:openstack_region] if @options[:openstack_region]
+
+      @volume_client ||= Fog::Volume.new(volumeoptionhash)
+
+      if not @volume_client
```
seanhandley:
```ruby
unless @volume_client
```

lib/beaker/hypervisor/openstack.rb
```diff
@@ -88,6 +102,46 @@ def network n
       @network_client.networks.find { |x| x.name == n } || raise("Couldn't find network: #{n}")
     end
 
+    # Guest volume support
+    # - Creates an array of volumes and attaches them to the current host
+    # - The host bus type is determined by the image type, so by default
+    #   devices appear as /dev/vdb, /dev/vdc etc.  Setting the glance
+    #   properties hw_disk_bus=scsi, hw_scsi_model=virtio-scsi will present
+    #   them as /dev/sdb, /dev/sdc (or 2:0:0:1, 2:0:0:2 in SCSI addresses)
```
anodelman: Please add appropriate yard doc @params here for each method argument.

lib/beaker/hypervisor/openstack.rb
```diff
@@ -88,6 +102,46 @@ def network n
...
+
+          # Fog needs a device name to attach as, so invent one.  The guest
+          # doesn't pay any attention to this
+          device = "/dev/vd#{('b'.ord + index).chr}"
+          vm.attach_volume(vol.id, device)
+        end
+      end
+    end
+
+    def cleanup_storage vm
```
anodelman: Please yard doc this.

spjmurray:
Thanks Alice.  Done, let me know if it needs any more finessing, or I got the types wrong... I'm not exactly expert in Ruby :)

lib/beaker/hypervisor/openstack.rb
anodelman: You are forcing the creation of the volume_client, even though there may be no host['volumes'] to create.

Probably want to use the pattern:
```
def volume_client
  @volume_client ||= Fog::Volume.new(volumeoptionhash)
end
```

spjmurray:
Updated the volume client to be lazily created.  Passes my acceptance test of spinning up a Ceph cluster with 3 OSDs and a journal volume :cocktail:

anodelman:
ok to test

PR #860, opened 2015-06-24
(BKR-323) Allow install_puppet_agent_on to support "pe only" platforms

anodelman:
- support both pm.puppetlabs.com and builds.puppetlabs.com for
  downloading puppet-agent
- use install_puppet_agent_on for yum|apt.puppetlabs.com
- use install_puppet_agent_dev_repo_on for builds.puppetlabs.com
- use install_puppet_agent_pe_promoted_repo_on for pm.puppetlabs.com

example usage:
  install_puppet_agent_on(host, { :version => '1.1.0', :default_action => 'gem_install'})
  install_puppet_agent_dev_repo_on(host, { :sha => 'd3377feaeac173aada3a2c2cedd141eb610960a7', :version => '1.1.1.225.gd3377fe'  })
  install_puppet_agent_pe_promoted_repo_on(host, { :sha => '1.1.0.227', :version => '1.1.0.227.g1d8334c', :pe_ver => '4.0.0-rc1'})


lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -20,8 +20,12 @@ module FOSSUtils
         include PuppetUtils
 
         DEFAULT_DOWNLOAD_URLS = {
```
kevpl: since these are defaults for FOSS, shouldn't they go in the `foss_defaults.rb`?

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -720,20 +728,21 @@ def install_puppet_from_gem_on( host, opts )
...
+          opts = DEFAULT_DOWNLOAD_URLS.merge(opts)
 
           case variant
-          when /^(fedora|el|centos)$/
+          when /^(sles|fedora|el|centos)$/
             variant = (($1 == 'centos') ? 'el' : $1)
 
             rpm = "puppetlabs-release%s-%s-%s.noarch.rpm" % [repo_name, variant, version]
-            remote = URI.join( options[:release_yum_repo_url], rpm )
+            remote = URI.join( opts[:release_yum_repo_url], rpm )
```
kevpl: yay! :+1:

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -923,26 +934,30 @@ def install_packages_from_local_dev_repo( host, package_name )
...
         #                       (default: '/root')
+        # @option opts [String] :puppet_collection Defaults to 'PC1'
+        # @option opts [String] :download_url Base url to look for build at
         #
         # @note on windows, the +:ruby_arch+ host parameter can determine in addition
         # to other settings whether the 32 or 64bit install is used
         #
         # @return nil
-        def install_puppetagent_dev_repo( host, opts )
+        def install_puppet_agent_repo( host, opts )
```
kevpl: shouldn't this be an `_on` method?

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -972,8 +987,68 @@ def install_puppetagent_dev_repo( host, opts )
...
+        #                       artifact will be pushed to on the external machine
+        #                       (default: '/root')
+        # @option opts [String] :puppet_collection Defaults to 'PC1'
+        # @option opts [String] :share_builds_url Base URL to pull artifacts from
+        #
+        # @note on windows, the +:ruby_arch+ host parameter can determine in addition
+        # to other settings whether the 32 or 64bit install is used
+        #
+        # @return nil
+        def install_puppet_agent_share_repo( host, opts )
```
kevpl: shouldn't this be an `_on` method? [same as above]

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -972,8 +987,68 @@ def install_puppetagent_dev_repo( host, opts )
...
+        #                       artifact will be pushed to on the external machine
+        #                       (default: '/root')
+        # @option opts [String] :puppet_collection Defaults to 'PC1'
+        # @option opts [String] :dev_builds_url Base URL to pull artifacts from
+        #
+        # @note on windows, the +:ruby_arch+ host parameter can determine in addition
+        # to other settings whether the 32 or 64bit install is used
+        #
+        # @return nil
+        def install_puppet_agent_dev_repo( host, opts )
```
kevpl: shouldn't this be an `_on` method? [same as above]

justinstoller:
@joshcooper, @haus (whoever else owns the test suites for PE only agents) this is the PR for trying to get in front of the work that will need to be done for SG updates. I would like you folks to review the PR as both devs and consumers of the dsl.

kevpl:
:+1:

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -455,26 +462,27 @@ def install_puppet_from_deb_on( host, opts )
...
+          block_on hosts do |host|
+            version = opts[:version]
 
-          if version && !version_is_less(version, '4.0.0')
-            if opts[:puppet_agent_version].nil?
-              raise "You must specify the version of puppet agent you " +
-                    "want to install if you want to install Puppet 4.0 " +
-                    "or greater on Windows"
-            end
+            if version && !version_is_less(version, '4.0.0')
```
haus: This check may break when we move agent versioning forward to 4.y.z, as mentioned on the dev list.

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -923,57 +950,134 @@ def install_packages_from_local_dev_repo( host, package_name )
...
+        # @example
+        #   install_puppet_agent_share_repo_on(host, { :sha => '1.1.0.227', :version => '1.1.0.227.g1d8334c', :pe_ver => '4.0.0-rc1'})
+        #
+        # @return nil
+        def install_puppet_agent_share_repo_on( hosts, opts )
+          block_on hosts do |host|
+            pe_ver = host[:pe_ver] || opts[:pe_ver] || '4.0'
+            variant, version, arch, codename = host['platform'].to_array
+            opts = FOSS_DEFAULT_DOWNLOAD_URLS.merge(opts)
+            opts[:download_url] = "#{opts[:share_builds_url]}/puppet-agent/#{pe_ver}/#{opts[:sha]}/repos/"
```
haus: As long as sha can also be a version this should be fine. For example, 1.2.0 agent just shipped, so `opts[:sha]` would be 1.2.0 for that build.

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -637,111 +656,116 @@ def install_puppet_agent_from_dmg_on(host, opts)
...
+          block_on hosts do |host|
+            variant, version, arch, codename = host['platform'].to_array
+            repo_name = repo.nil? ? '' : '-' + repo
+            opts = FOSS_DEFAULT_DOWNLOAD_URLS.merge(opts)
 
-          case variant
-          when /^(fedora|el|centos)$/
-            variant = (($1 == 'centos') ? 'el' : $1)
+            case variant
+            when /^(sles|fedora|el|centos)$/
```
haus: sles don't currently have release packages or repos at all on yum.pl.com.

haus:
This looks reasonable to me.

lib/beaker/dsl/install_utils/foss_utils.rb
kevpl: this is just used as a part of the URL to fetch the install on L994, so as long as that works in the URL to the location of the install, this should be fine.  Is that right, @anodelman ?

anodelman: Yes, this should work for either sha or version.

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -509,63 +517,72 @@ def compute_puppet_msi_name(host, opts)
...
+        def install_puppet_agent_from_msi_on(hosts, opts)
+          block_on hosts do |host|
 
-          # If we don't specify a version install the latest MSI for puppet-agent
-          if opts['version']
-            host['dist'] = "puppet-agent-#{opts[:version]}-#{arch}"
-          else
-            host['dist'] = "puppet-agent-#{arch}-latest"
-          end
+            is_config_32 = true == (host['ruby_arch'] == 'x86') || host['install_32'] || opts['install_32']
```
justinstoller: why the `true`?

justinstoller:
I'm not clear on why the repo or the methods are called "share". Could we change those to something that provides more of a clue that these methods are for install PE agents?

anodelman:
It's a 'share' because it has some publicly download-able items and some internal only.  So it's a shared resource.

I'm willing to go with another name if it makes more sense.

lib/beaker/dsl/install_utils/foss_utils.rb
anodelman: Consistency with the rest of these checks which we sprinkle through the code... and should really be gathered into a method.

justinstoller:
Cool

PR #867, opened 2015-06-29
(BKR-367) create beaker ec2 smoketest

anodelman:
- add acceptance/tests/hypervisor/communication.rb, checks to ensure
  that hosts can ping each other
- enable ping for aws ec2 instances created by beaker
- create host.ping method to ping from host to a second host, return
  'true' when ping is successful, 'false' otherwise

### DIFF
--- a/acceptance/tests/hypervisor/communication.rb
+++ b/acceptance/tests/hypervisor/communication.rb
@@ -1,0 +1,7 @@
+# hosts should be able to talk to each other by name
+step "hosts can ping each other"
+hosts.each do |one|
+  hosts.each do |two|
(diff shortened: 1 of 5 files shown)


kevpl:
@anodelman does our current acceptance test pswindows & regular windows hosts?  

I'm assuming creating a Jenkins base acceptance test for ec2 will be a follow-up to merging this, is that the case?

anodelman:
We do not have any pswindows testing - that is blocked on template creation.

I have an ec2 acceptance tests setup in jenkins.  It's green and ready to go and will be enabled once this lands.

PR #871, opened 2015-06-30
(BKR-370) beaker doesn't use answers set as env vars

anodelman:
- not correctly using 'answer_for' to determine if there is a provided
  value for all questions
- fixed up some variable naming for clarity


anodelman:
retest this please

PR #872, opened 2015-07-01
(BKR-374) apt/yum.puppetlabs use 'pc1' not 'PC1'

anodelman:


justinstoller:
Sweet, sweet consistency!

PR #873, opened 2015-07-02
(BKR-376) better interface to install puppet-agent through install_pe

anodelman:
- standardize on parameter names puppet_agent_version, puppet_agent_sha
    - if no puppet_agent_sha is provided default to puppet_agent_version
    - support new env vars
      * BEAKER_PUPPET_AGENT_VERSION
      * BEAKER_PUPPET_AGENT_SHA
      * BEAKER_PUPPET_COLLECTION
    
    - examples (puppet-agent installation):
      install_puppet_agent_on(host, { :version => '1.1.0', :default_action => 'gem_install'})
      install_puppet_agent_dev_repo_on(host, { :puppet_agent_sha => 'd3377feaeac173aada3a2c2cedd141eb610960a7', :puppet_agent_version => '1.1.1.225.    377fe'  })
      install_puppet_agent_pe_promoted_repo_on(host, { :puppet_agent_sha => '1.1.0.227', :puppet_agent_version => '1.1.0.227.g1d8334c', :pe_ver => '    .0-rc1'})
    
    - using install_pe/install_pe_on (for versions of PE > 4.0)
    set the puppet-agent version through an env var:
      export BEAKER_PUPPET_AGENT_VERSION=1.2.0
    can then just call:
      install_pe or install_pe_on(hosts, options)
    
    set the puppet-agent version through the options hash
      options[:puppet_agent_version] = '1.2.0'
      install_pe


lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -1069,14 +1080,15 @@ def install_puppet_agent_pe_promoted_repo_on( hosts, opts )
         # to other settings whether the 32 or 64bit install is used
         #
         # @example
-        #   install_puppet_agent_dev_repo_on(host, { :sha => 'd3377feaeac173aada3a2c2cedd141eb610960a7', :version => '1.1.1.225.gd3377fe'  })
+        #   install_puppet_agent_dev_repo_on(host, { :sha => 'd3377feaeac173aada3a2c2cedd141eb610960a7', :puppet_agent_version => '1.1.1.225.gd3377fe'  })
```
justinstoller: This also needs to be updated for `sha`.
ie
```
install_puppet_agent_dev_repo_on( host, sha: 'd337...', version: '1.2.1.16.d337')
```
becomes
```
install_puppet_agent_dev_repo_on( host, puppet_agent_sha: 'd337...', puppet_agent_version: '1.2.1.16.d337')
```

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -1069,14 +1080,15 @@ def install_puppet_agent_pe_promoted_repo_on( hosts, opts )
...
         #
         # @return nil
         def install_puppet_agent_dev_repo_on( hosts, opts )
+          opts[:puppet_agent_version] ||= opts[:version] #backward compatability
           block_on hosts do |host|
             variant, version, arch, codename = host['platform'].to_array
             opts = FOSS_DEFAULT_DOWNLOAD_URLS.merge(opts)
-            opts[:download_url] = "#{opts[:dev_builds_url]}/puppet-agent/#{opts[:sha]}/repos/"
+            opts[:download_url] = "#{opts[:dev_builds_url]}/puppet-agent/#{ opts[:puppet_agent_sha] || opts[:puppet_agent_repo] }/repos/"
```
justinstoller: `opts[:puppet_agent_repos]`? I don't see that value described in the docs. Do you mean `opts[:puppet_agent_version]`?

justinstoller:
@anodelman your description doesn't include a sample usage of what changed in the `install_pe_on` method, which from the title (and ticket) I assume this PR is addressing. Could you explain how this patch actually makes the PE interface easier?

justinstoller:
AFAICT, changing line 418 in pe_utils.rb to this:
```ruby
install_puppet_agent_pe_promoted_repo_on(host, { :version => opts[:puppet_agent_version],
                                                 :sha => opts[:puppet_agent_sha],
                                                 :pe_ver => opts[:pe_ver],
                                                 :puppet_collection => opts[:puppet_collection] })
```

Would have the same outward facing interface change to `install_pe_on`. I'm not sure why changing the interface to `install_pe_on` necessitates changing the interface to the other methods. ie. Why should the options hash to `install_puppet_agent_on` be the same as the options hash to `install_pe_on`?

anodelman:
I don't think that the parameters to the install_puppet_agent* methods were well chosen in the first place - so I'm trying to make the code more readable and understandable as a ride along to this.  I think that it is poor form to have some methods understand pe_ver and some understand version and some use version to mean both agent and puppet version and some to use both pe version and agent version (and on and on).  I'm aiming for clarity at what 'version' in each case means.

justinstoller:
I disagree with you (I think that `version` in a method called `install_puppet_agent` naturally is the version of puppet-agent and that coupling the options hash for every install helper is fragile). But we always think the code we write is more understandable than it probably is and I'm happy to defer to a larger consensus. Do you have a preference, @kevpl ?

kevpl:
while in general I agree with @justinstoller that if a method name makes a parameter name obvious, then it shouldn't be duplicated in the parameter name, in this situation, I agree with @anodelman.  My reasons:
- these files are changing rapidly since we're trying to keep up/stay ahead of development & testing
- with the modularization effort, things are only going to get more confusing until we fully separate these utils out into their own constructs

Due to these, I think that the less state implied by context, the better.  I know that the short-term solution becomes the long-term solution, and I think over-specifying is an alright thing to live longer than intended.

kevpl:
My concern with these changes is that I think we're making people pass in an explicit variable when there is a reasonable default for one of these values: `latest`.  I don't believe we have code to put latest into the PM URL as described by @haus in [BKR-341](https://tickets.puppetlabs.com/browse/BKR-341).  I'm not able to find it, anyways, so if we have it, please point that out to me.

To be honest, I didn't remember latest until @jpartlow brought it up:

    [11:02 AM] Joshua Partlow: If 'latest' is implicit then I think we're okay;
    I don't think it's tenable for the test to need to determine the latest 
    puppet-agent version associated with the current PE version.

anodelman:
Unfortunately, 'latest' doesn't work without additional info:
```
http://agent-downloads.delivery.puppetlabs.net/4.0/puppet-agent-latest/repos/el/7/PC1/x86_64/puppet-agent-1.2.1-1.el7.x86_64.rpm
```

You can see that it still have the version number in the rpm name - something that I can't discover.

kevpl:
since this is the case, I'm :+1: in general on this.
Let me know when you've gotten stake holders to a consensus, and if it's still this, I'll merge it.

anodelman:
Let's hold on this till the questions about the 'latest' file get resolved.

anodelman:
Actually, let's merge this and follow up in https://tickets.puppetlabs.com/browse/BKR-377.

justinstoller:
I doubt anyone's waiting on me, but FWIW, I'm a :+1: after further discussion.

PR #876, opened 2015-07-06
(BKR-383) puppet-agent dmg installation doesn't work for anything...

anodelman:
... other than 'latest'

- fix up the url construction for downloading a release mac puppet-agent
- tested locally
- there currently is not puppet-agent-latest.dmg file, so that won't
  work until that is posted

### DIFF
--- a/lib/beaker/dsl/install_utils/foss_utils.rb
+++ b/lib/beaker/dsl/install_utils/foss_utils.rb
@@ -210,7 +210,7 @@ module Beaker
         # @option opts [String] :version Version of puppet to download
         # @option opts [String] :puppet_agent_version Version of puppet agent to download
         # @option opts [String] :mac_download_url Url to download msi pattern of %url%/puppet-%version%.msi
-        # @option opts [String] :win_download_url Url to download dmg  pattern of %url%/(puppet|hiera|facter)-%version%.msi
+        # @option opts [String] :win_download_url Url to download dmg pattern of %url%/(puppet|hiera|facter)-%version%.msi
         #
         # @return nil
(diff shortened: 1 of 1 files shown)


sschneid:
:+1:

doug-rosser:
:+1:

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -632,25 +634,36 @@ def install_puppet_from_dmg_on( hosts, opts )
...
+            if agent_version == 'latest'
+              dmg_name = "puppet-agent-#{agent_version}.dmg"
+              on host, "curl -O #{opts[:mac_download_url]}/#{dmg_name}"
+            else
+              dmg_name = "puppet-agent-#{agent_version}-osx-#{version}-x86_64.dmg"
+              on host, "curl -O #{opts[:mac_download_url]}/#{opts[:puppet_collection]}/#{dmg_name}"
+            end
 
-            host.install_package("puppet-agent-#{version}")
+            host.install_package(pkg_name)
```
waynr: @joshcooper brought this up in another PR, but what happens when there are multiple dmgs on the SUT? Would it be reasonable to use `dmg_name` instead?

https://github.com/puppetlabs/beaker/pull/877#discussion-diff-33994167

anodelman: dmg_name actually won't work because the installer name and the package name that it installs are different - that was the first approach that I took.

anodelman:
Update to make the installed package name have the version number in it to stop accidental dmg installation.

waynr:
:+1:

anodelman:
Failure unrelated to patch.

PR #882, opened 2015-07-07
(BKR-388) Beaker::Host::Mac#touch returns non-existent path

anodelman:
- ensure that you can run the host.touch command successfully on mac
- ride along: touch also broken on windows, fixed and operational
- added acceptance test to ensure that this works against all platforms

### DIFF
--- a/lib/beaker/host/mac.rb
+++ b/lib/beaker/host/mac.rb
@@ -5,10 +5,11 @@ end
 module Mac
     class Host < Unix::Host
 
-    [ 'user', 'group', 'pkg' ].each do |lib|
+    [ 'exec', 'user', 'group', 'pkg' ].each do |lib|
       require "beaker/host/mac/#{lib}"
     end
 
+    include Mac::Exec
     include Mac::User
     include Mac::Group
     include Mac::Pkg
--- a/lib/beaker/host/mac/exec.rb
+++ b/lib/beaker/host/mac/exec.rb
@@ -1,0 +1,8 @@
+module Mac::Exec
+  include Beaker::CommandFactory
+
+  def touch(file, abs=true)
+    (abs ? '/usr/bin/touch' : 'touch') + " #{file}"
+  end
+
+end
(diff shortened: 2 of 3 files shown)


lib/beaker/host/windows/exec.rb
```diff
@@ -5,7 +5,7 @@ def reboot
     exec(Beaker::Command.new('shutdown /r /t 0 /d p:4:1 /c "Beaker::Host reboot command issued"'), :expect_connection_failure => true)
   end
 
-  ABS_CMD = 'c:\\\\windows\\\\system32\\\\cmd.exe'
+  ABS_CMD = "/cygdrive/c/WINDOWS/system32/cmd.exe"
```
joshcooper: This part surprises me, cmd.exe is very sensitive to forward slashes in the command line, as it dumbly thinks they are command arguments. It maybe that cygwin coverts that to backslashes before it invokes cmd.exe, but in any case, it seems risky, and unrelated to the mac issue. Maybe separate this out until we can verify windows acceptance is not affected? /cc @Iristyle

anodelman: The previous command didn't execute on cygwin windows at all, 
```
$ c:\\WINDOWS\\system32\\cmd.exe /c echo.
-bash: c:\WINDOWS\system32\cmd.exe: command not found
```

I'm unsure if this code path is in any use in our codebase, because it does not seem functional to me.

joshcooper: I think it's more that in the acceptance test, you could call the `Host#echo` method, and get a string back that could then be used in a manifest, eg https://github.com/puppetlabs/puppet/blob/master/acceptance/tests/language/resource_refs_with_nested_arrays.rb#L7

anodelman: Ah, that would change my view on it. :)  Let me pull it out of this PR.

colinPL:
Is this missing specs?

kevpl:
agreed, this is missing spec.  It's small enough that I'd be willing to let it slide, however.

@anodelman, do you think this is worth adding specs, or merging as is?

anodelman:
I can be a good citizen and add a spec test.

kevpl:
:+1:

PR #883, opened 2015-07-08
(BKR-390) Scooter pipeline is using pe-httpd instead of pe-puppetserver

anodelman:
- only use aio defaults on agents that are not master, database or
  dashboard (was previously using aio defaults/paths on all hosts
  identified as being >= 4.0)

### DIFF
--- a/lib/beaker/dsl/install_utils/foss_utils.rb
+++ b/lib/beaker/dsl/install_utils/foss_utils.rb
@@ -35,8 +35,9 @@ module Beaker
         #                            or a role (String or Symbol) that identifies one or more hosts.
         def configure_foss_defaults_on( hosts )
           block_on hosts do |host|
-            if (host[:version] && (not version_is_less(host[:version], '4.0'))) or host['type'] && host['type'] =~ /aio/
-              # add foss defaults to host
+            if (not_controller(host) && host[:version] && (not version_is_less(host[:version], '4.0'))) \
+              or (host['type'] && host['type'] =~ /aio/)
+              # add aio defaults to host
               add_aio_defaults_on(host)
             else
               add_foss_defaults_on(host)
--- a/lib/beaker/dsl/roles.rb
+++ b/lib/beaker/dsl/roles.rb
@@ -89,6 +89,35 @@ module Beaker
         find_host_with_role :default
       end
 
+      # Determine if host is not a controller, does not have roles 'master',
+      # 'dashboard' or 'database'.
+      #
+      # @return [Boolean]      True if agent-only, false otherwise
+      #
+      # @example Basic usage
+      #     if not_controller(host)
+      #       puts "this host isn't in charge!"
+      #     end
+      #
+      def not_controller(host)
+        controllers = ['dashboard', 'database', 'master', 'console']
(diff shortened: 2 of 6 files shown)


spec/beaker/dsl/install_utils/foss_utils_spec.rb
```diff
@@ -75,11 +75,19 @@ def logger
       subject.configure_foss_defaults_on( hosts )
     end
 
-    it 'uses aio paths for hosts of version >= 4.0' do
+    it 'uses aio paths for hosts of version >= 4.0, except fro master/database/dashboard' do
```
doug-rosser: Typo?

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -35,8 +35,9 @@ module FOSSUtils
         #                            or a role (String or Symbol) that identifies one or more hosts.
         def configure_foss_defaults_on( hosts )
           block_on hosts do |host|
-            if (host[:version] && (not version_is_less(host[:version], '4.0'))) or host['type'] && host['type'] =~ /aio/
-              # add foss defaults to host
+            if (agent_only(host) && host[:version] && (not version_is_less(host[:version], '4.0'))) \
+              or (host['type'] && host['type'] =~ /aio/)
```
sschneid: Is there a reason to use a regex instead of matching a value?  Could this bite us with future `host['type']`s?

anodelman:
retest this please

lib/beaker/dsl/install_utils/foss_utils.rb
anodelman: I'm doing a little future proofing for the potential creation of pe-aio and foss-aio (aio agent for pe master, aio agent for foss master).

doug-rosser:
:+1:

lib/beaker/dsl/roles.rb
```diff
@@ -89,6 +89,22 @@ def default
...
+      #
+      # @example Basic usage
+      #     if agent_only(host)
+      #       puts "this host is only an agent!"
+      #     end
+      #
+      def agent_only(host)
+        not_agent = ['dashboard', 'database', 'master', 'console']
+        roles = host['roles'].select { |v| not_agent.include?(v) }
+        roles.length == 0
```
kevpl: @anodelman I'm sure the way you're using this method is fine, but it's actually confusing to a user of the method.

For example, if a host has the roles `['agent', 'frictionless']` (as all agents that should be installed with frictionless should be unless I'm missing something), it will report as agent only.  This would probably be fine for your use case, but as soon as I saw this method, I thought, "oh hey, I should be able to use this for the [pe_util's install method](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/install_utils/pe_utils.rb#L508-L519)!"  Unfortunately it will report that we're agent only in that case, which we aren't.

I would suggest that we either say it doesn't have "higher" roles (hard to define), or we actually make this agent only as in the code example that I referenced.

anodelman:
retest this please

anodelman:
Green on scooter/classifier testing pipeline.

PR #884, opened 2015-07-10
(maint) Allow PE puppet-agent installs on custom roles

justinstoller:
Previously we installed puppet-agent on a host if there was the agent
role and no other role assigned to it. This prevents assigning a
puppet-agent node the "default" role to be used as the default test
target, or allow any other custom roles.

To allow assigning the "default" or custom roles to a puppet-agent node
we list the non-puppet-agnet roles we know how to generate installation answers
for and use that list to determine what nodes are puppet-agent or not.

### DIFF
--- a/lib/beaker/dsl/install_utils/pe_utils.rb
+++ b/lib/beaker/dsl/install_utils/pe_utils.rb
@@ -509,8 +509,9 @@ module Beaker
         def create_agent_specified_arrays(hosts)
           hosts_agent_only = []
           hosts_not_agent_only = []
+          non_agent_only_roles = %w(master database dashboard console frictionless)
           hosts.each do |host|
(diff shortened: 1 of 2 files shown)


justinstoller:
This, I think, is the solution to what we were talking about yesterday, @kevpl. Also, this is blocking module testing, @kevpl  and @anodelman.

anodelman:
@justinstoller the concept of a `not_controller` node was just introduced in a separate PR (https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/roles.rb#L102) that is probably appropriate to use here.

justinstoller:
@anodelman we also want to exclude installing puppet-agent on PE when it is a `frictionless` role that `not_controller` doesn't include.

justinstoller:
Passed locally on Debian 7 and Windows 2012r2!

PR #886, opened 2015-07-10
(BKR-393) fixed ubuntu PM URLs, didn't match package versions

kevpl:


lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -1114,6 +1114,7 @@ def install_puppet_agent_pe_promoted_repo_on( hosts, opts )
               if arch == 'x86_64'
                 arch = 'amd64'
               end
+              version = version[0,2] + '.' + version[2,2] if variant =~ /ubuntu/
```
kevpl: @anodelman do we already have a way of solving this issue?  I feel like it's something we've probably seen before, but I'm not recalling if we have a better solution for it..

anodelman: Should maybe use the same solution that we have for osx?

https://github.com/puppetlabs/beaker/blob/master/lib/beaker/platform.rb#L26

Same thing, need to ensure that 1010 -> 10.10.

anodelman: No wait, that conflicts with what our templates consider a valid version number.

anodelman:
retest this please

anodelman:
retest this please

kevpl:

anodelman:
retest this please

anodelman:
retest this please

anodelman:
retest this please

lib/beaker/dsl/install_utils/foss_utils.rb
anodelman: The issue here is that 'ubuntu-14.04-amd64' is a valid platform (and would be the platform generated by genconfig) - and then this code would not be necessary.

I'm poking around to see if there is a more elegant solution to this.

kevpl: how about just changing the if statement check to this?

    if variant =~ /ubuntu/ && version =~ /\.{0}/

kevpl:
@anodelman, you and I talked about your objection to this PR, which was that we should standardize on a version of the ubuntu to use (either something like "1404" or "14.04").

Looking at the platform code for this review, I'm seeing that we are standardizing on a version to use, it's the version without the period ([reference](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/platform.rb#L81)).

Since that's the case, that makes this code just handling a weirdness of the PM server, which makes this location the best place for it.

kevpl:
failure unrelated to change

anodelman:
Sweet, sweet consistency.

PR #887, opened 2015-07-13
(BKR-394) added support for 3.8/AIO mixed installations

kevpl:


lib/beaker/dsl/install_utils/pe_utils.rb
```diff
@@ -512,7 +516,11 @@ def create_agent_specified_arrays(hosts)
           non_agent_only_roles = %w(master database dashboard console frictionless)
           hosts.each do |host|
             if host['roles'].none? {|role| non_agent_only_roles.include?(role) }
-              hosts_agent_only << host
+              if host[:pe_ver] && version_is_less(host[:pe_ver], '4.0')
```
colinPL: I see this check twice - here and again in `roles.rb`. Can this be moved to a method? `is_aio_capable?` or something?

lib/beaker/dsl/roles.rb
```diff
@@ -115,7 +115,11 @@ def not_controller(host)
       #       puts "this host is ONLY an agent!"
       #     end
       def agent_only(host)
-        host['roles'].length == 1 && host['roles'].include?('agent')
+        if host[:pe_ver] && version_is_less(host[:pe_ver], '4.0')
+          false
```
doug-rosser: the test "agent_only" becomes a bit ambiguous here, because there's a small chance someone might consider a "legacy agent" as an agent. This is admittedly a corner case!

kevpl: it is an agent, it's just not "agent_only", which is a special kind of agent.

that does seem confusing, however.  Can we think of a better name for it?  That would be a much bigger diff, but I'm open to it if something more specific strikes.  `aio_agent?`, perhaps?

doug-rosser:
:+1:  I've already seen @johnduarte report success manually testing installs

lib/beaker/dsl/roles.rb
anodelman: Yeah, I don't like mixing this logic into agent_only, which is only for indicating if a host has the single role 'agent'.

A separate method is in order for what you are trying to determine.

kevpl:
I manually stopped that jenkins run.  a new one has already kicked off from my latest push

kevpl:

anodelman:
@kevpl you ready to have this merged?  Did you diagnose the ride along issue?

kevpl:
@anodelman once it goes through this time, then I'll be ready for merge.

The issue before was from an old assumption.  In AIO, the host property `privatebindir` was introduced because the dev teams' were splitting up where the vendored binaries and the rest go.  When we set that up, the pe version was set to be the same as the new AIO version, since it had never existed before.  It should have been set to the old `puppetbindir` path, so that you could always use it to find ruby and the vendored bins, however.  The path change fixes that.

anodelman:
retest this please

lib/beaker/dsl/install_utils/pe_defaults.rb
```diff
@@ -32,7 +32,7 @@ module PEDefaults
             'puppetbin'        => '/opt/puppet/bin/puppet',
             'puppetbindir'     => '/opt/puppet/bin',
             'puppetsbindir'    => '/opt/puppet/sbin',
-            'privatebindir'    => '/opt/puppetlabs/puppet/bin',
+            'privatebindir'    => '/opt/puppet/bin',
```
johnduarte: This winds up flipping the desired values. Looks like I gave you a bum steer @kevpl

kevpl: after talking with John, we realized that this is the correct path, but that other logic was causing the wrong hosts to be set with their paths, so we've updated that instead of this.

anodelman:
Green on classifier tests.

anodelman:
retest this please

PR #888, opened 2015-07-13
(BKR-397) Use yum for installing pe promoted puppet-agent

justinstoller:
Prior to this we used rpm, which means puppet-agent won't resolve its dependency on dmidecode for EL systems.

To ensure that it correctly resolves that dependency we need to use yum.


justinstoller:
Running locally, brb!

justinstoller:
Okay, I had originally forgotten the `-y` but now this is working locally!

PR #890, opened 2015-07-14
(BKR-403) Use rpm on sles for pe puppet-agent

justinstoller:
In BKR-397 we updated rpm using platforms to use yum instead of rpm in
the install_puppet_agent_pe_promoted_repo, however it wasn't noticed
that SLES was also an rpm using platform in the match that doesn't also
include yum.

This patch moves SLES platforms into their own branch of logic that uses
the old rpm installation method while retaining the yum based
installation for other platforms.


justinstoller:
Huzzah!

PR #891, opened 2015-07-14
(BKR-402) Regression in Beaker 2.18.0 AIO testing for PuppetDB usage

anodelman:
- provide a sane default for puppetservice for aio defaults, but allow
  override with a user setting

### DIFF
--- a/lib/beaker/dsl/install_utils/pe_utils.rb
+++ b/lib/beaker/dsl/install_utils/pe_utils.rb
@@ -28,6 +28,8 @@ module Beaker
           block_on hosts do |host|
             if host[:pe_ver] && aio_version?(host) or (host['type'] && host['type'] =~ /aio/)
               add_aio_defaults_on(host)
+              # provide a sane default here for puppetservice
+              host['puppetservice'] ||= 'pe-puppetserver'
             else
               add_pe_defaults_on(host)
             end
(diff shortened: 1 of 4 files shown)


anodelman:
Fixes platform testing, breaks classifier testing.

lib/beaker/dsl/install_utils/pe_utils.rb
```diff
@@ -28,6 +28,8 @@ def configure_pe_defaults_on( hosts )
           block_on hosts do |host|
             if host[:pe_ver] && aio_version?(host) or (host['type'] && host['type'] =~ /aio/)
               add_aio_defaults_on(host)
+              # provide a sane default here for puppetservice
+              host['puppetservice'] ||= 'pe-puppetserver'
```
kevpl: @anodelman why don't this line (and it's duplicate in `foss_utils.rb`) just get added to the aio defaults themselves (or to `add_aio_defaults_on`) so we eliminate the duplicates?

kevpl:
@anodelman why do we need to add the `puppetservice` default differently for aio than we do for pe?

Are we supposed to wait to consider this for merge until it works for both platform & classifier testing?  Are we supposed to be able to get these results anywhere automatically, or is this just from your reporting of local testing?

anodelman:
We don't currently have aio platform or classifier testing in our smoketests, so I'm running them locally.

AIO does not include a puppetservice default at all, because we don't know if we are installing in a pe or foss environment, so we can't determine which puppetservice is correct.

kevpl:
ah, ok. I missed the small difference between those lines. Thanks for clearing that stuff up.  I'm :+1: on these changes, so let me know when this passes your local testing, and once it's through CI, we'll get it merged.

anodelman:
Passed local platform & classifier testing.

PR #892, opened 2015-07-15
(maint) Moves various VB options out of `disk_path` logic

petems:
Looks like these got moved into the `if` logic for `disk_path` accidentally around v2.8. This means that you can no longer use these options without adding disk_path logic in later versions of Beaker. This moves them back into their own section.


anodelman:
Please update commit message with (MAINT) tag.

anodelman:
retest this please

anodelman:
Failure unrelated to patch.

PR #897, opened 2015-07-18
(BKR-411) Set host.ip for Fusion hypervisor

johnduarte:
This commit updates the Fusion hypervisor to set the host[:ip]
value based on the IP address obtained from querying the VM data
using the fission gem.

This removes the need for the user to have setup dns resolution to
the `host.name` prior to running beaker.

### DIFF
--- a/lib/beaker/hypervisor/fusion.rb
+++ b/lib/beaker/hypervisor/fusion.rb
@@ -54,6 +54,15 @@ module Beaker
         end
         time = Time.now - start
         @logger.notify "Spent %.2f seconds resuming VM" % time
+
+        @logger.notify "Obtaining IP information for #{host.name}"
+        ip = vm.network_info.data[vm.network_info.data.keys[0]]['ip_address']
+        if ip.empty?
+          @logger.notify "Unable to obtain IP information for #{host.name}. Continuing."
+        else
+          @logger.notify "Found IP address #{ip} for #{host.name}. Setting host.ip value"
+          host[:ip] = ip
+        end
       end
       end #revert_fusion
 
--- a/lib/beaker/hypervisor/fusion.rb
+++ b/lib/beaker/hypervisor/fusion.rb
@@ -1,3 +1,4 @@
+require 'beaker/ssh_connection'
 module Beaker
   class Fusion < Beaker::Hypervisor
 
@@ -55,20 +56,64 @@ module Beaker
         time = Time.now - start
         @logger.notify "Spent %.2f seconds resuming VM" % time
 
-        @logger.notify "Obtaining IP information for #{host.name}"
-        ip = vm.network_info.data[vm.network_info.data.keys[0]]['ip_address']
-        if ip.empty?
-          @logger.notify "Unable to obtain IP information for #{host.name}. Continuing."
-        else
-          @logger.notify "Found IP address #{ip} for #{host.name}. Setting host.ip value"
-          host[:ip] = ip
+        begin
+          try_ssh_connection(host.name, host['user'], host['ssh'])
+        rescue *Beaker::SshConnection::RETRYABLE_EXCEPTIONS => e
+          @logger.notify "Obtaining IP information for #{host.name}"
+          ip = vm.network_info.data[vm.network_info.data.keys[0]]['ip_address']
+          if not ip
+            raise "Unable to connect to vm via hostname #{host.name}, and ip address is unavailable via vmname #{host[:vmname]}."
+          else
+            @logger.notify "Found IP address #{ip} for #{host.name}. Setting host.ip value"
+            begin
+              try_ssh_connection(ip, host['user'], host['ssh'])
+            rescue *Beaker::SshConnection::RETRYABLE_EXCEPTIONS => e
+              raise "Unable to connect to vm via IP address #{ip} for #{host.name}"
+            end
+            host[:ip] = ip
+          end
+          set_hostnames @hosts, @options
+
         end
       end
+      hack_etc_hosts @hosts, @options
+
       end #revert_fusion
 
       def cleanup
         @logger.notify "No cleanup for fusion boxes"
       end
 
+      # Set the hostname of all instances to be the hostname defined in the
+      # beaker configuration.
+      #
+      # @param [Host, Array<Host>] hosts An array of hosts to act upon
+      # @param [Hash{Symbol=>String}] opts Options to alter execution.
+      #
+      # @return [void]
+      # @api private
+      def set_hostnames hosts, opts
+        hosts.each do |host|
+          if host['platform'] =~ /el-7/
+            # on el-7 hosts, the hostname command doesn't "stick" randomly
+            host.exec(Command.new("hostnamectl set-hostname #{host.name}"))
+          else
+            host.exec(Command.new("hostname #{host.name}"))
+          end
+        end
+      end
+
+      # Attempt SSH connection to given host
+      #
+      # @param host [String] the hostname or IP address to connect to
+      # @param user [String] the username to use for the connection
+      # @param ssh [Hash] SSH parameters to use for the connection
+      #
+      # @return [void]
+      # @api private
+      def try_ssh_connection(host, user, ssh)
+        @logger.notify "Attempting SSH connection to #{host}"
+        Net::SSH.start(host,user,ssh)
+      end
   end
 end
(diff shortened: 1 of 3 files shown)


johnduarte:
@anodelman @kevpl  this is a quick hack to add this functionality. Please take a look and let me know how it should be amended to meet your code quality standards.

johnduarte:

anodelman:
retest this please

johnduarte:
This PR should also address setting up the internal DNS between hosts when using the IP address of Fusion nodes/

anodelman:
I'd like to test this locally before merging.

anodelman:
Looks good locally.  I would like a spec test to ensure that it correctly handles the case when an ip is not available.

johnduarte:
@anodelman spec tests have been applied along with additional logic to add update the hosts files on the nodes if the host.ip is set.

anodelman:

lib/beaker/hypervisor/fusion.rb
```diff
@@ -61,5 +81,48 @@ def cleanup
...
+            domain = get_domain_name(neighbor)
+            ip = neighbor['ip']
+            etc_hosts += "#{ip}\t#{name} #{name}.#{domain} #{neighbor['dns_name']}\n"
+          end
+          set_etc_hosts(host, etc_hosts)
+        end
+      end
+
+      def try_ssh_connection(host)
+        Net::SSH.start(host.name,host['user'],host['ssh'])
```
anodelman: Need to pass logger through here.

lib/beaker/hypervisor/fusion.rb
```diff
@@ -61,5 +81,48 @@ def cleanup
...
+            name = neighbor.name
+            domain = get_domain_name(neighbor)
+            ip = neighbor['ip']
+            etc_hosts += "#{ip}\t#{name} #{name}.#{domain} #{neighbor['dns_name']}\n"
+          end
+          set_etc_hosts(host, etc_hosts)
+        end
+      end
+
+      def try_ssh_connection(host)
```
anodelman: Needs yard docs.

lib/beaker/hypervisor/fusion.rb
```diff
@@ -54,6 +55,25 @@ def provision
...
+          try_ssh_connection(host)
+        rescue *Beaker::SshConnection::RETRYABLE_EXCEPTIONS => e
+          @logger.notify "Obtaining IP information for #{host.name}"
+          ip = vm.network_info.data[vm.network_info.data.keys[0]]['ip_address']
+          if not ip
+            raise "Unable to connect to vm via hostname #{host.name}, and ip address is unavailable via vmname #{host[:vmname]}."
+          else
+            @logger.notify "Found IP address #{ip} for #{host.name}. Setting host.ip value"
+            host[:ip] = ip
+          end
```
anodelman: The assumption here is that the ip will work, considering that we are testing the ssh connection by name we should probably do the same for ip and then raise as invalid before doing additional work on the host.

lib/beaker/hypervisor/fusion.rb
```diff
@@ -54,6 +55,25 @@ def provision
...
+
+          # Set the hostname for each box
+          set_hostnames()
```
anodelman: Could this be generally useful?  Should it be the default case and not just the host-by-ip case?

lib/beaker/hypervisor/fusion.rb
```diff
@@ -54,6 +55,25 @@ def provision
...
+
+          # Configure /etc/hosts on each host
+          configure_hosts()
```
anodelman: Could this be generally useful? Should it be the default case and not just the host-by-ip case?

lib/beaker/hypervisor/fusion.rb
```diff
@@ -61,5 +81,48 @@ def cleanup
...
+            host.exec(Command.new("hostname #{host.name}"))
+          end
+        end
+      end
+
+      # Configure /etc/hosts for each node
+      #
+      # @return [void]
+      # @api private
+      def configure_hosts
```
anodelman: I believe that this logic is already contained in host_prebuild_steps::hack_etc_hosts.

johnduarte: FYI: I copy pasta'd this from `lib/beaker/hypervisor/aws_sdk.rb`, so we may want to refactor it out of that location as well.

lib/beaker/hypervisor/fusion.rb
johnduarte: +1

lib/beaker/hypervisor/fusion.rb
johnduarte: +1. Thanks @anodelman !!

spec/beaker/hypervisor/fusion_spec.rb
```diff
@@ -36,6 +36,45 @@ module Beaker
...
+      end
+
+      it 'then host fails init if ip cannot be introspected' do
+        allow_any_instance_of( MockFissionVM ).to receive( :network_info ).and_return( Response.new(0,'',{'eth0' => {'ip_address' => nil}}))
+        expect{ fusion.provision }.to raise_error(/ip address is unavailable/)
+      end
+
+      it 'then sets host dns_name to instance.dns_name' do
+        @hosts.each do |host|
+          expect(fusion).to receive(:set_etc_hosts).with(host, kind_of(String)).exactly(@hosts.count).times
```
johnduarte: @anodelman I am not sure about this line. I had to set the receive count to the number of hosts to get it to pass, but my expectation would be that this method would be passed only the host in question.

anodelman:

kevpl:
last comment still current

anodelman:
@johnduarte If you don't have time to look at this can we reassign the ticket and close out this PR?

er0ck:
@anodelman FYI: john is on PTO for a bit.

lib/beaker/hypervisor/fusion.rb
johnduarte: I don't think we want to crush the hostnames on the SUTs unless we cannot get to them by hostname.

johnduarte:
retest this please

kevpl:
:+1:, looks good to me, but since @anodelman had the majority of the concerns, I'll wait for her to have a pass at this to make sure all of her concerns are addressed.  She's on PTO for a few days, so this should happen mid-to-end-of-week

rick:
Thanks! Yeah, I'm not bottlenecked on this, I've worked around it in the other PR.

lib/beaker/hypervisor/fusion.rb
```diff
@@ -54,12 +55,65 @@ def provision
...
+          else
+            @logger.notify "Found IP address #{ip} for #{host.name}. Setting host.ip value"
+            begin
+              try_ssh_connection(ip, host['user'], host['ssh'])
+            rescue *Beaker::SshConnection::RETRYABLE_EXCEPTIONS => e
+              raise "Unable to connect to vm via IP address #{ip} for #{host.name}"
+            end
+            host[:ip] = ip
+          end
+          set_hostnames @hosts, @options
```
anodelman: This should be done outside of the @hosts.each loop.

lib/beaker/hypervisor/fusion.rb
```diff
@@ -54,12 +55,65 @@ def provision
         end
         time = Time.now - start
         @logger.notify "Spent %.2f seconds resuming VM" % time
+
+        begin
+          try_ssh_connection(host.name, host['user'], host['ssh'])
```
anodelman: I'm not a big fan of moving ssh connection knowledge outside of the ssh connection module.  Might be more reasonable to update the ssh_connection.connect to take a parameter to reduce the number of re-tries and then use it instead.  The, instead of this whole block, you could just run a random command on the host (like: ```on host, 'echo'```) to establish a connection.  If the connection fails to establish catch the runtime error and raise from there, otherwise you can then check the ip address.

rick: Is there a way to use the command to gather the desired information -- instead of issuing an I'm-calling-this-command-for-its-connection-testing-side-effects-only command?

kevpl:
last comment still current

kevpl:
last comment still current. @johnduarte, this PR will be closed next week if no action is taken.

PR #898, opened 2015-07-20
(BKR-415) re-targeted puppet_pkg acceptance from 3.7.5 -> 3.8.1

kevpl:


anodelman:
This passed the part of the testing that it affects.  Please update the jenkins test name to reflect the new version of puppet pkg under test.

PR #900, opened 2015-07-23
(maint) Always wait for Windows agent installs

ferventcoder:
When installing Windows agents with cygwin or any other means, always
prefer `start /w` over just calling msiexec.

This adds to what was first introduced in 9c32cac.

Also adds using `type` over `cat` in pure Windows
 - If the host is not a cygwin environment, prefer `type` to produce the
   contents of the log file over `cat`. `type` is built-in, where `cat` is
   only available if someone has put *nix tools on the path, whether
   through a git install (MinGW tools), installing GnuWin utils or some
   other method.


anodelman:
Looks good.  Test failure unrelated to patch.

PR #902, opened 2015-07-24
(BKR-412) clarify 'types' in beaker

anodelman:
- having made 'aio' a type meant that we were losing pe/foss specific
  information (like the puppet service)
- make 'aio' a role that a host can have, meaning that the aio defaults
  will be installed overtop of any foss/pe defaults
- assume hosts with no pe/foss puppet version provided are aio

### DIFF
--- a/lib/beaker/dsl/install_utils/puppet_utils.rb
+++ b/lib/beaker/dsl/install_utils/puppet_utils.rb
@@ -65,6 +65,46 @@ module Beaker
           end
         end
 
+        # Configure the provided hosts to be of their host[:type], it host[type] == nil do nothing
+        def configure_type_defaults_on( hosts )
+          block_on hosts do |host|
+            has_defaults = false
+            if host[:type]
+              host_type = host[:type]
+              # clean up the naming conventions here (some teams use foss-package, git-whatever, we need
+              # to correctly handle that
+              # don't worry about aio, that happens in the aio_version? check
+              host_type = case host_type
+                          when /(\A|-)(git)|(foss)(\Z|-)/
+                            'foss'
+                          when /(\A|-)pe(\Z|-)/
+                            'pe'
+                          else
+                            nil
+                          end
+              if host_type
+                add_method = "add_#{host_type}_defaults_on"
+                if self.respond_to?(add_method, host)
+                  self.send(add_method, host)
+                else
+                  raise "cannot add defaults of type #{host_type} for host #{host.name} (#{add_method} not present)"
+                end
+                has_defaults = true
+              end
+            end
+            if aio_version?(host)
+              add_aio_defaults_on(host)
+              has_defaults = true
+            end
+            # add pathing env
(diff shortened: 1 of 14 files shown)


lib/beaker/dsl/install_utils/puppet_utils.rb
```diff
@@ -65,6 +65,34 @@ def configure_defaults_on( hosts, type )
           end
         end
 
+        # Configure the provided hosts to be of their host[:type], it host[type] == nil do nothing
+        def configure_type_defaults_on( hosts )
```
waynr: What about projects that are already setting `type => 'aio'` in their Beaker options?

spec/beaker/dsl/install_utils/foss_utils_spec.rb
```diff
@@ -29,66 +30,62 @@ def logger
...
       end
+      expect(subject).to receive(:add_foss_defaults_on).exactly(hosts.length).times
       expect(subject).to receive(:add_aio_defaults_on).exactly(hosts.length).times
       expect(subject).to receive(:add_puppet_paths_on).exactly(hosts.length).times
 
       subject.configure_foss_defaults_on( hosts )
     end
 
-    it 'uses foss paths for hosts of type foss' do
+    it 'uses no paths for hosts with no type' do
```
waynr: I am interested in what happens when `:type` is set as `aio` because I know there are projects currently using this setting and I want to be sure something sane happens.

lib/beaker/dsl/install_utils/puppet_utils.rb
anodelman: These will still end up calling 'aio_defaults_on' through 'add_method = "add_#{host_type}_defaults_on"'.

spec/beaker/dsl/install_utils/foss_utils_spec.rb
anodelman: I added some additional spec tests to show the new behavior and that this code is backwards compatible.

rick:
Definitely not intending to stop the :bus: here, as I think it pertains to more of the modularity work we've been talking about doing, ... I'm wondering long-term if "AIO" is really a strong concept in the domain.

When we start moving PE-related helpers out to a gem, it's probably a good time to see if we can get everything that making AIO into a `type` or `role` gives us can be had by just having versions of installable software that define collections of paths (for example).  I have a hunch that "AIO" might evaporate.

anodelman:
Yeah, if we only had to set defaults based upon the software installed by beaker we would be set since we can do version checks easily.  The tricky bit is that there are outside installation libraries that depend heavily on the ability to run with a 'type'.  So, the aim here is to maintain backwards compatibility until we can get everyone on the same :bus:.

anodelman:
retest this please

anodelman:
As green as it gets, these days.

PR #905, opened 2015-07-27
(BKR-330) Refactor for host creation logic

petems:
Takes some of the logic from Hypervisor, this will allow us to use custom hosts in the future.

Not sure how to write tests for this, might need a hand.

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -35,23 +35,25 @@ module Beaker
     end
 
     def self.create name, host_hash, options
-      case host_hash['platform']
+      host_class = case host_hash['platform']
       when /windows/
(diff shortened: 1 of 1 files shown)


anodelman:
retest this please

anodelman:
@petems I am putting together an answer file library that has a method for determining the correct object to create on the fly, I think that we will be using the recipe used there for this sort of thing in the future.  I'll link once I get it into the puppet namespace.

anodelman:
@petems You can see the pattern for how we discover additional modules here:

https://github.com/puppetlabs/beaker-answers/blob/master/lib/beaker-answers/answers.rb#L57

Closing this out.

PR #907, opened 2015-07-28
(BKR-427) Support for Docker Container Names and Container Reuse

hamidnazari:
This allows you to specify a name for your docker container if you so desire. If the container doesn't exist, it gets created as usual. If it does exist, it doesn't get created and will be reused. You'll also be able to refer to your containers using names you specify, rather than using container IDs or randomly generated names.

I usually create a master container, and an agent container to do my testing. Spinning up a master container every single test became unnecessarily time consuming and repetitive. I needed my master container to run only for the first time and remain there, while my agents would get created and destroyed regularly. 

Sample usage:
```YAML
HOSTS:
  ubuntu-12-10:
    platform: ubuntu-12.10-x64
    image: ubuntu:12.10
    hypervisor: docker
    docker_container_name: ubuntu-container
CONFIG: 
  type: foss
```

### DIFF
--- a/lib/beaker/hypervisor/docker.rb
+++ b/lib/beaker/hypervisor/docker.rb
@@ -64,11 +64,28 @@ module Beaker
           image_name = image.id
         end
 
-        @logger.debug("Creating container from image #{image_name}")
-        container = ::Docker::Container.create({
+        container_opts = {
           'Image' => image_name,
           'Hostname' => host.name,
-        })
+        }
+
+        unless host['docker_container_name'].nil?
+          @logger.debug("Looking for an existing container called #{host['docker_container_name']}")
+          existing_container = ::Docker::Container.all.select do |container| container.info['Names'].include? "/#{host['docker_container_name']}" end
+
+          # Prepare to use the existing container or else create it
+          if existing_container.any?
+            container = existing_container[0]
+          else
+            container_opts['name'] = host['docker_container_name']
+          end
+        end
+
+        # If the specified container exists, then use it rather creating a new one
+        if container.nil?
+          @logger.debug("Creating container from image #{image_name}")
+          container = ::Docker::Container.create(container_opts)
+        end
 
         @logger.debug("Starting container #{container.id}")
         container.start({"PublishAllPorts" => true, "Privileged" => true})
--- a/spec/beaker/hypervisor/docker_spec.rb
+++ b/spec/beaker/hypervisor/docker_spec.rb
@@ -39,6 +39,9 @@ module Beaker
       container = double('Docker::Container')
       allow( container ).to receive(:id)
       allow( container ).to receive(:start)
+      allow( container ).to receive(:info).and_return(
+        *(0..2).map { |index| { 'Names' => ["/spec-container-#{index}"] } }
+      )
       allow( container ).to receive(:json).and_return({
         'NetworkSettings' => {
           'IPAddress' => '192.0.2.1',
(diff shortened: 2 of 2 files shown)


anodelman:
Please associated your commit with a beaker ticket, ie (BKR-xxx).

anodelman:
@electrical of interest to you, i think.

hamidnazari:
Back to you Alice. @anodelman

electrical:
Hi,

I'm trying to think up the use case for this with regards of docker containers.
In most cases you'll have a long living image but a short living container since you don't want old tests / data pollute your new tests.
What i can imagine is that you supply an image name specifically for the master container for example.

With regards to the random name, since these are all short lived containers, nobody cares about the random names.

anodelman:
Looks like your commit history is a little corrupt, has reference to a beaker gem release.  Needs a rebase.

kevpl:
last comment still current

hamidnazari:
Better now? @kevpl @anodelman

hamidnazari:
@electrical you're right. Generally you'd want to have short living containers and long living images. A use case in a simple case like this is simply when you want your containers to be named. So if a test fails, you know a command like `docker exec -it testcontainer bash` will always refer to the test container, not a randomly generated docker name. And if you're happy with the random name, then just don't give your containers a name.

Another example, which was my main motive behind this PR, is that I need a long living Puppet master container. This container is essential to running the tests and is stable and don't need to be tested itself. So I don't want to create/destroy/recreate it every time I run a test due to the time it consumes. Is there a better way of doing this?

electrical:
@hamidnazari I understand your use case.
For connecting to the container, there is always the way to connect via ssh to the docker instance.
The connection string is printed out at startup.
This way it will also work if you would use something like Docker Swarm.
I personally would use a separate container that is not managed via Beaker for a Puppet Master since you want to use it for multiple tests?

kevpl:
@hamidnazari yeah, the commits don't look polluted by merge commits anymore.  The commit messages will also have to be updated with the ticket ID just like you've updated the PR title.  You can do this with an interactive rebase.

hamidnazari:
@electrical I thought about that too. I still interact with the Puppet Master container in `spec_helpter_acceptance.rb` to do various things such as load up the latest code, revoke previous test agent's cert, etc.

hamidnazari:
@kevpl I'd already done that. Not quite what happened, but it got messed up with the second rebase! Regardless, it should be good now.

electrical:
@hamidnazari Ah i see. Since you do load up the latest test code you could most likely live with a PuppetMaster Image and start it at test time so you don't have to keep it running.
Advantage of that is that you don't have to worry about old certs or stale code since it will be loaded fresh at test time.
And you won't have to worry to much about startup time if you build the image correctly.

electrical:
That said, As long as the current workflow is not impacted by default I'm okay with the change.

kevpl:

hamidnazari:
Thanks @electrical. That's very close to what I'm doing. I have a base Puppet Master image, with close to no content in it. Beaker brings it up and let's it live (or for the duration of the tests, if I set it to). Then it's just reused from then on. Maybe I should point out that this set up is more about BDD rather than pure testing. That's why the agent's come and go, but the master has a somewhat longer life span.

hamidnazari:
@kevpl is it `ruby-1.8.7-p374` failing? Probably not related to my changes.
https://travis-ci.org/hamidnazari/beaker

anodelman:
retest this please

anodelman:
Failure unrelated to patch.

anodelman:
@electrical just waiting on your :+1:

lib/beaker/hypervisor/docker.rb
```diff
@@ -64,11 +64,25 @@ def provision
...
-        @logger.debug("Creating container from image #{image_name}")
-        container = ::Docker::Container.create({
+        container = []
+        container_opts = {
           'Image' => image_name,
           'Hostname' => host.name,
-        })
+        }
+
+        unless host['docker_container_name'].nil?
```
electrical: I think the logic here could be a little bit better.

- Not using the 'container' variable since its re-used later, rename into something else.
- Also the container = [] should be renamed. all kinds of weird assignments are happening here.
- container.any? should only be done if we actually set the docker_container_name ( otherwise follow normal workflow )

electrical:
Left 1 comment about the code flow.

hamidnazari:
Sorry I was on a trip for a few days. I'll get back to you @electrical.

hamidnazari:
@electrical I see your point. Better now? I'm not too fussed about this, if you think there's a better way of doing it.

anodelman:
@electrical any final comments here?

Please cleanup the errant merge commits in your commit chain.

electrical:
Looks good to me.

anodelman:
Cool, just looking for some commit cleanup and this is good to go.

kevpl:
last comment still current

hamidnazari:
Hey @kevpl, what sort of cleanup are you after?

kevpl:
from above:

    Please cleanup the errant merge commits in your commit chain.

We don't usually have merge commits in PRs, we rebase as needed to get upstream changes.

hamidnazari:
Not a problem. I'll take of it later tonight.

hamidnazari:
Back to you @kevpl.

kevpl:

PR #910, opened 2015-07-28
(BKR-426) long-term fix to end '/' issue in fetch_http_file

kevpl:


kevpl:
failure unrelated to change

PR #911, opened 2015-07-30
(BKR-422) debian7 fails to reboot during smoketest

anodelman:
- to fix current smoketest redness change how we connect to beaker boxes
- do three separate connection attempts (with retries)
  * first with ip
  * then the virtual hostname (vmhostname)
  * then the name of the box provided by the user (hostname)
- Only when all three fail do we fail out
- decrease the Net::SSH timeout so that we can move through the various
  attempts in a reasonable amount of time
- add a sleep post windows reboot, getting reds here because windows
  takes so long to come back up

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -214,8 +214,9 @@ module Beaker
     end
 
     #Return the ip address of this host
+    #Always pull fresh, because this can sometimes change
     def ip
-      self[:ip] ||= get_ip
+      self['ip'] = get_ip
     end
 
     #@return [Boolean] true if x86_64, false otherwise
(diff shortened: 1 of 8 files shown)


anodelman:
Green tests!

lib/beaker/host.rb
```diff
@@ -224,13 +225,29 @@ def is_x86_64?
     end
 
     def connection
-      @connection ||= SshConnection.connect( reachable_name,
+      # create new connection object if necessary
+      @connection ||= SshConnection.connect( { :ip => self['ip'], :vmhostname => self['vmhostname'], :hostname => @name },
```
colinPL: Should this call `ip` method instead of the host param (`self['ip']`)? Or will `self['ip']` trigger the above `ip` method anyway.

lib/beaker/host.rb
```diff
@@ -224,13 +225,29 @@ def is_x86_64?
...
                                              self['user'],
                                              self['ssh'], { :logger => @logger } )
+      # update connection information
+      if self['ip'] && (@connection.ip != self['ip'])
```
colinPL: Any reason not to just set `@connection.ip` to `self['ip']` if the latter has a value? Is there any reason to compare the two to determine if the same or not?

lib/beaker/host.rb
```diff
@@ -224,13 +225,29 @@ def is_x86_64?
...
+      end
+      if @name && (@connection.hostname != @name)
+        @connection.hostname = @name
+      end
+      @connection
     end
 
     def close
       @connection.close if @connection
+      # update connection information
```
colinPL: Does this do anything if 3 lines later `@connection` is set to `nil`?

anodelman:
Failed to reboot on windows.  I'll take another look.

lib/beaker/host.rb
anodelman: The first ensures that the connection is closed, the second throws away the old object.

lib/beaker/host.rb
anodelman: It was to make writing the spec tests easier - since you ask for the connection object every time that you do a host.exec it goes through this code path a lot and will end up resetting the value again and again.  Probably not the prettiest, but you also don't know if the ip has become available or if it is a new ip.   ip used to be considered constant but now we allow for change during runtime (at least, we've had situations where it has lost its ip lease and gotten a new one).

lib/beaker/host.rb
anodelman: Ah, that's tricky.  If you call self.ip it will attempt to get the ip through a host.exec - which then attempts to get a connection object to the host which then attempts to determine the ip and so on until stack too deep.

So, we only use the ip if it is known, either set by the provisioner or determined by the user after a connection has been established.

PR #918, opened 2015-08-07
(BKR-439) Add DNF for fedora-22

johnduarte:
This commit adds a case condition for fedora-22 to use the 'dnf'
command where the 'yum' command would previously have been executed.

### DIFF
--- a/lib/beaker/dsl/install_utils/foss_utils.rb
+++ b/lib/beaker/dsl/install_utils/foss_utils.rb
@@ -1231,6 +1231,9 @@ module Beaker
             scp_to host, File.join(copy_dir_local, download_file), onhost_copy_base
 
             case variant
+            when /^(fedora-22)$/
+              on host, "tar -zxvf #{onhost_copied_download} -C #{onhost_copy_base}"
+              on host, "dnf --nogpgcheck localinstall -y #{onhost_copied_file}"
             when /^(fedora|el|centos)$/
               on host, "tar -zxvf #{onhost_copied_download} -C #{onhost_copy_base}"
               on host, "yum --nogpgcheck localinstall -y #{onhost_copied_file}"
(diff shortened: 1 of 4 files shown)


lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -1231,6 +1231,9 @@ def install_puppet_agent_pe_promoted_repo_on( hosts, opts )
             scp_to host, File.join(copy_dir_local, download_file), onhost_copy_base
 
             case variant
+            when /^(fedora-22)$/
+              on host, "tar -zxvf #{onhost_copied_download} -C #{onhost_copy_base}"
+              on host, "dhf --nogpgcheck localinstall -y #{onhost_copied_file}"
```
er0ck: typo in 'dnf' here?

er0ck:
no tests?

johnduarte:
The underlying `rpm` call should be the same and is covered by

https://github.com/puppetlabs/beaker/blob/master/spec/beaker/host/unix/pkg_spec.rb#L41

electrical:
A test should be added so we can confirm the right command us used for fedora 22.

kevpl:
last comment (from electrical) still current

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -1231,6 +1231,9 @@ def install_puppet_agent_pe_promoted_repo_on( hosts, opts )
             scp_to host, File.join(copy_dir_local, download_file), onhost_copy_base
 
             case variant
+            when /^(fedora-22)$/
+              on host, "tar -zxvf #{onhost_copied_download} -C #{onhost_copy_base}"
```
er0ck: all these tars for non-windows should/could probably be pulled out of the case statement

johnduarte:
Spec test added to confirm use of `dnf` on `fedora-22` and `yum` on `fedora-20`

kevpl:
:100:

PR #924, opened 2015-08-17
(BKR-447) Ubuntu PE puppet-agent urls not being generated correctly

anodelman:
- incorrectly adding additional '.' to ubuntu version strings that were
  already correct
- added spec test to ensure correct behavior
- passed local testing


anodelman:
Failure unrelated to patch.

PR #925, opened 2015-08-17
(BKR-448) Check for connection object during host.close

objectverbobject:
The close method for hosts assumed the connection object existed when
trying to set the hostname; this fix checks to ensure the connection
object exists prior to calling any methods on it.


anodelman:
Failure unrelated to patch.

PR #928, opened 2015-08-19
(maint) Update Ubuntu Code Names

colinPL:
Add Ubuntu 14.04 to debian code name mapping.


lib/beaker/host/unix/pkg.rb
```diff
@@ -179,6 +179,8 @@ def upgrade_package(name, cmdline_args = '', opts = {})
     'ubuntu-10.04-i386'  => 'lucid',
     'ubuntu-12.04-amd64' => 'precise',
     'ubuntu-12.04-i386'  => 'precise',
+    'ubuntu-14.04-amd64' => 'trusty',
+    'ubuntu-14.04-i386'  => 'trusty',
```
anodelman: This same info, in a slightly different format, is in platform.rb - maybe we should fix that :)

colinPL: The one in `platform.rb` is more up-to-date and complete.

colinPL:
There is a better solution than what is in this PR. Closing as this method is invalid.

PR #929, opened 2015-08-19
(MAINT) Fix acceptance puppet_git load path

rick:
Prior to this, running the acceptance tests locally without specifying a specific load path would result in a require failure if attempting to test against puppet installed via git.

/cc @er0ck

### DIFF
--- a/acceptance/pre_suite/puppet_git/install.rb
+++ b/acceptance/pre_suite/puppet_git/install.rb
@@ -1,4 +1,5 @@
 begin
+  $LOAD_PATH << File.expand_path(File.join(File.dirname(__FILE__), '..', '..', 'lib'))
   require 'beaker/acceptance/install_utils'
   extend Beaker::Acceptance::InstallUtils
 end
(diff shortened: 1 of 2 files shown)


rick:
For what it's worth, running locally I consistently get a failure that seems unrelated to the load path problem: see https://gist.github.com/rick/0c0bc2ec5c855cc5c5b3

er0ck:
hmmm... so i'm already hitting the limits of this solution.  eg: tests/foss_utils/clone_git_repo_on.rb would also need this patch. this leads me back to.. an options file.

rick:
Yeah, I don't have a problem with getting the load path right for any tests that need it. In situations where there's something like a `test_helper.rb` or a bundler dep, or a Rakefile that are always run prior to tests, etc., lifting up is fine, but I'm used to seeing load path work happening in multiple files otherwise.

anodelman:
Unrelated to this patch, but I would prefer if foss_utils lived in a slightly different directory structure:
```
acceptance/tests/dsl/foss_utils
```

er0ck:
good idea, alice.  hell, why not just jumble a whole slew of unrelated fixes in here :-)

anodelman:
Fine, fine.  I can file a ticket for that.  But I don't like you much for making me do that.

er0ck:
@anodelman i didnt' mean to be facetious.. i'm about to push to this branch ...
but shouldn't that be dsl/install_utils/foss_utils/ ?

rick:
Tests are passing for me locally on this branch now.

rick:
:bow:

PR #931, opened 2015-08-20
(BRK-452) Use Platform To Get Code Name

colinPL:
The `unix/pkg.rb` had its own mapping of platform to debian code name. However, this is available through the Platform class. This commit removes the duplicate data in `pkg.rb` and uses `self['platform'].codename` in its place.


kevpl:
:+1:

anodelman:
Just waiting on test results.

anodelman:
Failure unrelated to patch.

PR #934, opened 2015-08-27
(BKR-459) update osx package name and directory to correspond...

anodelman:
... to RE changes
    
    - updated install_puppet_agent_dev_repo_on to handle new directory
      structure and name change
    - backwards compatible, handles both old and new
    - fixed up a small bug about overriding passed in options


haus:
@anodelman this looks reasonable to me.

PR #930, opened 2015-08-20
[BKR-358] [DO NOT MERGE] add acceptance tests for `dsl::helper::host_helpers`

rick:
Adding acceptance tests for Beaker's `dsl::helper::host_helpers` (see: https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/helpers/host_helpers.rb)

**remaining work**

 - [x] ensure that all characterization tests (`CURRENTLY` in the step descriptions) have documented what is the expected behavior instead
 - [x] get PR review :eyes: 
 - [x] `hosts.first` -> `default`
 - [x] `TEST_SCP_ERROR_ON_CLOSE` -> `BEAKER_ TEST_SCP_ERROR_ON_CLOSE`
 - [x] can we fixture-file `testfile.sh` (and friends) ?
 - [x] fix `rsync` issues on CentOS (it doesn't come installed by default)
 - [x] use `confine` syntax instead of `if` blocks
 - [x] get solaris green
 - [x] is there more post-test cleanup that should be done?
   - we are using tempdirs on all remote and local hosts, not sure what else can be done (especially with no support for teardown blocks)
 - [x] update the configuration used in beaker acceptance tests to use the criteria discussed [in this comment](https://github.com/puppetlabs/beaker/pull/930#issuecomment-136538874)
 - [x] **LAND #944** otherwise `confine_block` doesn't even work
 - [x] split test files out for readability
 - [x] ensure tests for jenkins platforms are :green_heart::
   - [x] :green_heart: solaris11
   - [x] :green_heart: ubuntu1404
   - [x] :green_heart: fedora22
   - [x] :green_heart: osx109
   - [x] :green_heart: ubuntu1504
   - [x] :yellow_heart: debian7
     - have seen intermittent failures: [log](http://jenkins-beaker.delivery.puppetlabs.net/job/qe_beaker_intn-sys_beaker-acceptance-base-vpool/798/agent=debian7/console), [log](http://jenkins-beaker.delivery.puppetlabs.net/job/qe_beaker_intn-sys_beaker-acceptance-base-vpool/799/agent=debian7/console)
      - these appear to be SSH connection problems: `rsync returned #<Rsync::Result:0x00000003a40be8 @raw="ssh: connect to host 10.32.122.240 port 22: Connection timed out\r\nrsync: connection unexpectedly closed (0 bytes received so far) [sender]\nrsync error: unexplained error (code 255) at io.c(605) [sender=3.0.9]\n", @exitcode=255>`
[...]

PR #937, opened 2015-08-28
(MAINT) constrain `add_system32_hosts_entry`

rick:
Prior to this, the host helper method `add_system32_hosts_entry` was available to any windows platform, including cygwin-enabled windows installs. This method only actually works on non-cygwin windows platforms. This discrepancy was uncovered during work for BKR-358, where we are adding acceptance tests for the host helper methods.

/cc @puppetlabs/beaker

### DIFF
--- a/lib/beaker/dsl/helpers/host_helpers.rb
+++ b/lib/beaker/dsl/helpers/host_helpers.rb
@@ -12,7 +12,7 @@ module Beaker
         #     (or range) of integer exit codes that should be considered
         #     acceptable.  An error will be thrown if the exit code does not
         #     match one of the values in this list.
-        #   @option opts [Boolean] :accept_all_exit_codes (false) Consider all 
+        #   @option opts [Boolean] :accept_all_exit_codes (false) Consider all
         #     exit codes as passing.
         #   @option opts [Boolean] :dry_run (false) Do not actually execute any
         #     commands on the SUT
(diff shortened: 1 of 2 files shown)


kevpl:
@rick what's the reason that we're removing the tests instead of updating them to have the correct actions based on the new test?

lib/beaker/dsl/helpers/host_helpers.rb
```diff
@@ -331,7 +331,7 @@ def upgrade_package host, package_name
         #
         # @return nil
         def add_system32_hosts_entry(host, opts = {})
```
joshcooper: A meta question, why not use puppet to manage the hosts file, e.g. `puppet resource host ...`?

rick:
@kevpl the tests aren't actually exercising the system under tests -- if they were, they would be failing. We're replacing the tests, just not inside this PR (with the tests that uncovered the fact that these tests aren't actually testing the code in question).  See: https://github.com/puppetlabs/beaker/pull/930/files#diff-664d0804b338cc06a8f5765e31e0eebeR760

lib/beaker/dsl/helpers/host_helpers.rb
rick: @joshcooper raises a good point.  

There are a few meta-questions here, including "why is there this random powershell-specific method in the hosts helpers?".  My understanding is that @anodelman is in the midst of more cleanup and consolidation of windows-related methods.

As the code exists at the moment, things inside the hosts helper are presuming that puppet is not necessarily installed.  In the bigger picture, it may well be useful to lift some methods up and out to a place where they could assume that puppet is installed.

This PR is just concerned with wrapping acceptance tests (actually, a number of these are specifically [characterization tests](http://c2.com/cgi/wiki?CharacterizationTest)) around core code that doesn't really have coverage.  After we have tests in place we should feel more empowered to think about consistency and how functionality is carried out.

rick: Correction -- s/This PR/that other PR/ (#930)

PR #939, opened 2015-09-02
(BKR-486) Supply vmpooler auth token on provision/destroy

sschneid:
Supply vmpooler auth token (if defined in `.fog`) on provision/destroy.

### DIFF
--- a/lib/beaker/hypervisor/vcloud_pooled.rb
+++ b/lib/beaker/hypervisor/vcloud_pooled.rb
@@ -19,10 +19,35 @@ module Beaker
       @options = options
       @logger = options[:logger]
       @hosts = vcloud_hosts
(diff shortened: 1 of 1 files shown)


sschneid:
The code in-question passes the `Beaker::VcloudPooled` spec:

````
Beaker::VcloudPooled
  #get_template_url
    works returns the valid url when passed valid pooling_api and template name
    adds a missing scheme to a given URL
    raises an error on an invalid pooling api url
    raises an error on an invalide template name
  #provision
    provisions hosts from the pool
  #cleanup
    cleans up hosts in the pool

Finished in 0.15882 seconds
6 examples, 0 failures
````

Acceptance tests are likely never going to pass on this, though...

anodelman:
Sounds good to me.

PR #940, opened 2015-09-03
(BKR-440) host_preserved.yml file should disable the pre_suite

anodelman:
- instead of just removing the pre_suite/post_suite/tests from
the options, explicitly set it to be []
- this means that you don't end up accidentally running a pre-suite
  defined in a secondary options file

### DIFF
--- a/spec/beaker/cli_spec.rb
+++ b/spec/beaker/cli_spec.rb
@@ -191,6 +191,39 @@ module Beaker
         end
       end
 
+      describe "#preserve_hosts_file" do
+        it 'removes the pre-suite/post-suite/tests and sets to []' do
+          hosts =  make_hosts
+          options = cli.instance_variable_get(:@options)
+          options[:log_dated_dir] = Dir.mktmpdir
+          File.open("sample.cfg", "w+") do |file|
+            file.write("HOSTS:\n")
+            hosts.each do |host|
+              file.write("  #{host.name}:\n")
+              file.write("    roles:\n")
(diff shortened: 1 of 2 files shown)


anodelman:
Failure unrelated to patch.

er0ck:
:+1:

waynr:
@anodelman it seems like there should probably be a test validating this behavior to prevent regressions and accidental behavior incompatibilities between major versions of Beaker. Actually, this PR does introduce a backwards incompatibility (anyone who relied on setting pre_suite in an options file when using preserved hosts will not be able to do so anymore...i know that seems unlikely, just saying.)

anodelman:
Added spec coverage.

spec/beaker/cli_spec.rb
```diff
@@ -191,6 +191,39 @@ module Beaker
...
+          options[:post_suite] = ['post1']
+          options[:tests] = ['test1', 'test2']
+
+          cli.instance_variable_set(:@options, options)
+          cli.instance_variable_set(:@hosts, hosts)
+
+          preserved_file = cli.preserve_hosts_file
+          hosts_yaml = YAML.load_file(preserved_file)
+          expect(hosts_yaml['CONFIG'][:tests]).to be == []
+          expect(hosts_yaml['CONFIG'][:pre_suite]).to be == []
```
er0ck: nice.

waynr:
Failure unrelated to change, :+1: from me!

PR #942, opened 2015-09-03
(BKR-497) skip configuration for f5 hosts

kevpl:


LuvCurves:
Not doing the job. set_env from lib/beaker/host_prebuilt_steps.rb is still being called.

kevpl:
I abandoned that test since it was out of date due to the new changes.  There's already another one in the queue for these new changes, so the next update from Jenkins should be accurate

LuvCurves:
Working now.  Partially my fault.  Forgot platform now has to be f5-11.6.0-25mbps-good instead of el-f5-11.6.0-25mbps-good like it used to be in the old hypervisor.

Ship it.

PR #944, opened 2015-09-09
(BKR-505) rescue `SkipTest` in `confine_block`

rick:
Prior to this, use of `confine_block` would result in all following code in the test case being skipped in situations where the host list had no matches for the confine_block.  The entire purpose of `confine_block` is to limit the scope of the confine to the passed block.

Also, `ensure` and `rescue` at the method level do not require an additional wrapping `begin..end` block.

This was tested locally against the acceptance test suite still in development (and currently partially blocked by this bug) at https://github.com/puppetlabs/beaker/pull/930

/cc @puppetlabs/beaker @justinstoller

### DIFF
--- a/acceptance/tests/base/dsl/structure_test.rb
+++ b/acceptance/tests/base/dsl/structure_test.rb
@@ -1,0 +1,63 @@
+test_name "dsl::structure" do
+  step "#confine_block runs specified block on matching hosts" do
+    begin
+      @in_confine = 0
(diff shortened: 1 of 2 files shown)


rick:
I have added acceptance tests for `#confine_block` here which also expose the bug and verify the fix.

:eyes:

anodelman:
Failure unrelated to patch.

PR #953, opened 2015-09-15
(BKR-523) Beaker prints error message on exit when using docker containers

GeoffWilliams:
Capture the strange error message from docker (and all other internal docker errors) and display the message as a logged warning instead of a stack trace.  This will preven the message from causing a CI test failure as the exit status is now zero.

Doesn't address the root-cause of this warning message but handles it in a more graceful way


GeoffWilliams:
## Error message
screen shot 2015-09-15 at 10 49 23

PR #954, opened 2015-09-15
(BKR-528) would like to be able to run full systemd in docker

GeoffWilliams:
Hardcode ENV and VOLUME to allow systemd to work in docker.


electrical:
Any specific reasons for these values? I've been running full systemd with docker for quite a while and never needed these.

GeoffWilliams:
Didn't seem to work without them, but perhaps I overlooked something.  I'll retest when I get a moment

GeoffWilliams:
OK I seemed to be able to run systemd without this despite what the above links said - closing

PR #955, opened 2015-09-15
(BKR-532) added beaker-hiera library usage

kevpl:


anodelman:

PR #958, opened 2015-09-16
(BKR-536) skip ssh installation for images that already provide it

GeoffWilliams:
Add check to see if a host tells us it already has ssh installed, if so, don't attempt to install it again.  To skip installation, add the key  to your node definition, eg:

```
HOSTS:
  pe-puppet.localdomain:
    roles:
      - "agent"
      - "master"
    platform: "el-7-x86_64"
    image: "pe2015-2-0_aio-master"
    ssh_installed: true
```

### DIFF
--- a/lib/beaker/hypervisor/docker.rb
+++ b/lib/beaker/hypervisor/docker.rb
@@ -160,62 +160,71 @@ module Beaker
         FROM #{host['image']}
       EOF
 
-      # additional options to specify to the sshd
-      # may vary by platform
-      sshd_options = ''
+      if not host.has_key?("ssh_installed") then
+        # additional options to specify to the sshd
+        # may vary by platform
+        sshd_options = ''
 
-      # add platform-specific actions
-      service_name = "sshd"
-      case host['platform']
-      when /ubuntu/, /debian/
-        service_name = "ssh"
-        dockerfile += <<-EOF
-          RUN apt-get update
-          RUN apt-get install -y openssh-server openssh-client #{Beaker::HostPrebuiltSteps::DEBIAN_PACKAGES.join(' ')}
-        EOF
-        when  /cumulus/
+        # add platform-specific actions
+        service_name = "sshd"
+        case host['platform']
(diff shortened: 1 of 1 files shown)


electrical:
I'm -1 on this. We should either do auto detection which is not possible , or just not care about it.
it brings no extra overhead anyway.

GeoffWilliams:
@electrical how would we autodetect?  I need something along these lines as I've got a docker image that isn't bootable with stock beaker. What do you think about my suggest to have static Dockerfiles (hipchat channel)?  That would solve this for me.  As for not caring - how can we bypass the above failure? thanks

GeoffWilliams:
```
{"stream":"Step 3 : RUN ssh-keygen -t rsa -f /etc/ssh/ssh_host_rsa_key\n"}

{"stream":" ---\u003e Running in 6c122e50aff0\n"}

{"stream":"Generating public/private rsa key pair.\n/etc/ssh/ssh_host_rsa_key already exists.\nOverwrite (y/n)? "}

{"errorDetail":{"code":1,"message":"The command '/bin/sh -c ssh-keygen -t rsa -f /etc/ssh/ssh_host_rsa_key' returned a non-zero code: 1"},"error":"The command '/bin/sh -c ssh-keygen -t rsa -f /etc/ssh/ssh_host_rsa_key' returned a non-zero code: 1"}
```

electrical:
Your problem is not as much that ssh is installed but that we always do the keygen part.
There are 2 ways we can solve this;

1. The main image you use, remove the generated keys out of it and let beaker generate it. This ensures its random at all times.

2. Do a test if the file exist and only if it doesn't run ssh-keygen.
For example: `test -f /etc/ssh/ssh_host_rsa_key|| ssh-keygen -t rsa -f /etc/ssh/ssh_host_rsa_key`
This should only run the ssh-keygen part if the file doesn't exist.

GeoffWilliams:
@electrical I had a think about this and came up with a different approach - closed in favour of https://github.com/puppetlabs/beaker/pull/961

PR #959, opened 2015-09-16
(BKR-535) Regression: confine_block does not skip tests...

anodelman:
...with beaker 2.24.0

- allow users to include skip_test in block parameter for confine_block
- added acceptance test to ensure correct behavior

### DIFF
--- a/acceptance/tests/base/dsl/structure_test.rb
+++ b/acceptance/tests/base/dsl/structure_test.rb
@@ -60,4 +60,20 @@ test_name "dsl::structure" do
       fail "#confine_block raised unexpected SkipTest exception: #{e}"
     end
   end
+
+  step "#confine_block allows blocks to raise skip_test" do
+    previous_hosts = hosts.dup
+
+    begin
+      @in_confine = 0
+      confine_block :to, :platform => default["platform"] do
+        @in_confine +=1
+        skip_test "this block raises a skip"
+      end
+    rescue Beaker::DSL::Outcomes::SkipTest => e
+      assert_match /this block raises a skip/, e.message, "#confine_block raised an unexpected skip_test"
+      assert_equal 1, @in_confine, "#confine_block did not execute supplied block"
+      assert_equal hosts.dup, hosts, "#confine_block did not preserve the hosts array"
+    end
+  end
 end
--- a/lib/beaker/dsl/structure.rb
+++ b/lib/beaker/dsl/structure.rb
@@ -214,7 +214,13 @@ module Beaker
 
         yield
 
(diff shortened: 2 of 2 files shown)


rick:
:+1:

I'm also totally down with being /cc'd on PRs, and even via:

/cc @puppetlabs/beaker

joshcooper:
@anodelman @rick this isn't a blocker for us and since PE doesn't use `confine_block` I'm thinking we fix  this during the next scheduled beaker release (so no firedrill), and I'll submit a puppet PR to fix the acceptance tests. Make sense? /cc @kylog @johnduarte

anodelman:
Sounds reasonable - we're scheduled to do a release next Wednesday.

rick:
@joshcooper I'm in tune with that :musical_note:  

@anode wondering if, in light of this not being a :fire: drill whether we can look at refactoring this a bit to make things a little cleaner/clearer for posterity.  I haven't sat with it to try a candidate refactoring, but would be game to see if there's a way to clean it a bit.

johnduarte:
:+1: for no :fire: and rolling into regular release. Thanks @joshcooper

anodelman:
@rick The code could be cleaned up, but it would involve updating any test using confine or confine_block to use options  - otherwise we end up with two arguments with default values and everything goes sideways.

rick:
Mostly thinking of a behavior-preserving refactoring (which, I guess is the
definition), it'll either be something I can spike out quickly, or I will
pretty quickly not find anything.  I guess I'm just asking not to merge for
a day and see if I can find something.

On Wed, Sep 16, 2015 at 12:47 PM, alice nodelman <notifications@github.com>
wrote:

> @rick <https://github.com/rick> The code could be cleaned up, but it
> would involve updating any test using confine or confine_block to use
> options - otherwise we end up with two arguments with default values and
> everything goes sideways.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/puppetlabs/beaker/pull/959#issuecomment-140817019>.
>



-- 

*PuppetConf 2015 <http://2015.puppetconf.com/> is coming to Portland,
Oregon! Join us October 5-9.*
*Register now to take advantage of the Final Countdown discount
<https://www.eventbrite.com/e/puppetconf-2015-october-5-9-register-now-its-the-final-countdown-tickets-13115894995?discount=FinalCountdown>
*
*—**save $149!*

anodelman:
Yup, more eyes (and brains) are welcome!  We can pause on this for now.  :hourglass_flowing_sand:

joshcooper:
Fixed puppet in https://github.com/puppetlabs/puppet/pull/4255

PR #961, opened 2015-09-18
(BKR-539) simplified Dockerfile support

GeoffWilliams:
Allow user to override the automatically generated Dockerfile, externalise the generation of Dockerfiles to ERB templates, add an osfamily method to platform.rb

### DIFF
--- a/lib/beaker/hypervisor/docker.rb
+++ b/lib/beaker/hypervisor/docker.rb
@@ -151,90 +151,98 @@ module Beaker
     end
 
     def dockerfile_for(host)
+      if host['dockerfile'] then
+        @logger.debug("attempting to load user Dockerfile from #{host['dockerfile']}")
+        if File.exist?(host['dockerfile']) then
+          dockerfile = File.read(host['dockerfile'])
+        else
+          raise "requested Dockerfile #{host['dockerfile']} does not exist"
+        end 
+      else 
+        raise("Docker image undefined!") if (host['image']||= nil).to_s.empty?
 
-      # Warn if image is not define, empty or nil
-      @logger.error("Docker image undefined!") if (host['image']||= nil).to_s.empty?
-
-      # specify base image
-      dockerfile = <<-EOF
-        FROM #{host['image']}
-      EOF
-
-      # additional options to specify to the sshd
-      # may vary by platform
-      sshd_options = ''
-
-      # add platform-specific actions
-      service_name = "sshd"
-      case host['platform']
-      when /ubuntu/, /debian/
-        service_name = "ssh"
-        dockerfile += <<-EOF
-          RUN apt-get update
-          RUN apt-get install -y openssh-server openssh-client #{Beaker::HostPrebuiltSteps::DEBIAN_PACKAGES.join(' ')}
+        # specify base image
+        dockerfile = <<-EOF
+          FROM #{host['image']}
         EOF
-        when  /cumulus/
+
+        # additional options to specify to the sshd
+        # may vary by platform
+        sshd_options = ''
+
+        # add platform-specific actions
+        service_name = "sshd"
+        case host['platform']
+        when /ubuntu/, /debian/
+          service_name = "ssh"
           dockerfile += <<-EOF
-          RUN apt-get update
-          RUN apt-get install -y openssh-server openssh-client #{Beaker::HostPrebuiltSteps::CUMULUS_PACKAGES.join(' ')}
-        EOF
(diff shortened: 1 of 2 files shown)


GeoffWilliams:
tests seem to be aborted not failing - jenkins problem?

kevpl:
@electrical, look good to you?

electrical:
I'll have a look at this today.

electrical:
I don't really like the loads of code duplication for the docker files.
There must be a better solution for whatever you are trying to achieve.

GeoffWilliams:
@electrical  intention was that users and programmers would have a well defined bunch of Dockerfile templates to look at when debugging or adding support for new platforms.  This would also allow for diverging code in the future.  

The reality is that this isn't helpful - at least not at the moment so have refactored the repeated bottom half of the Dockerfile to be in a single template that gets used everywhere - hows this?

GeoffWilliams:
Test failure seems unreleated?

electrical:
I'm still trying to understand this change and don't see it.
From https://github.com/puppetlabs/beaker/pull/958 you had issues with the generation of the ssh keys that caused issues, wouldn't it make more sense to solve that specific issue?
In most cases a docker image shouldn't have the keys pre-genned anyway since every container would have the same ssh key set ( which makes no sense )

GeoffWilliams:
@electrical, thanks for looking at this and good point about the keys.  For a throw-away test system its not something I'd considered.

Removing the keys from my image would have solved my immediate need but while writing my Beaker tests I've been finding it difficult to determine and control the final Dockerfile that gets generated so I wanted to see if I could improve flexibility while also simplifying the Ruby code somewhat.  The aims of this PR are:
* Allow users to supply their own complete and final Dockerfile via the dockerfile key.  This is great for edge cases/experiments that need bits added or removed from the file to work.  Its also a nice way of supporting additional platforms without having to hassle the developers
* Allow existing Beaker/Docker tests to work unaltered
* Remove the inline strings from the docker.rb file and turn them into simple-to-read templates.  This makes it easier to see how the Dockerfile is generated and reduces the amount of ruby code that needs to be maintained.

A side effect of these changes are that my problem gets solved.  I did consider the embedding shell script in the existing code but I wanted to avoid the deep-nesting of bash, Dockerfile, Beaker and Ruby as its difficult to debug when things go wrong.

Hopefully this clears up the intention of this PR a bit?

electrical:
I don't think some core things should be easy editable since beaker relies on them.
And if not done correctly other failures might happen.
That's the whole reason its all been coded inside the hypervisor and exposed all configurable parts via the node definition.
Although it might be easier to supply your own docker parts it should not be needed and we should avoid code duplication as much as possible, even if its in templates.

because of this i'm still not 100% happy with the PR, but perhaps other can chime in their idea's.

GeoffWilliams:
@electrical - I found the ability to swap in a complete Dockerfile was useful to me for experimentation if nothing else.  If users swap in a Dockerfile, then its reasonable for them to take responsibility for any breakages that occur when doing so and we can ask them to replicate failure using the stock auto-generated Dockerfile before opening tickets if necessary.

In the end, I found that it was easier to prepare the image I wanted outside of Beaker and then reference the finished image.

I cleaned up the templates so that the repeated section at the end is no longer duplicated.  The code that's left is pretty much the contents of the case statement externalised to ERB files.  In my mind this was a nice code improvement but then I'm not the one who's going to be maintaining this.

If you have specific improvements to this PR in mind, please let me know and I'll try and put something together, otherwise I think we can look at closing this PR and the accompanying ticket if its not a useful addition.

- Thanks

electrical:
@GeoffWilliams I personally rather see a feature then to completely replace the docker file instead of extracting the current things in the templates which i still don't like tbh.
What you could do is supply a Dockerfile in the node config and use that, otherwise the default things.
What do you think?

GeoffWilliams:
so keep the override and lose the templates?  That could be a good compromise, I'll try and put something together when I get a moment

kevpl:
@GeoffWilliams please update the commit messages with the ticket reference.

Thanks,
Kevin

GeoffWilliams:
@kevpl updated PR with squashed commit and fixed message

electrical:
Looks good to me. Not sure what the failing test is since i don't have access to it.

GeoffWilliams:
@electrical the failing test seems unrelated.  It looks like the command `dnf -y  install ntpdate` is failing due to some repositories being unavailable/connectivity problems - thanks

PR #962, opened 2015-09-18
(BKR-542) remove hosts object from log about execution against empty …

pondohva:
…host list

### DIFF
--- a/lib/beaker/shared/host_manager.rb
+++ b/lib/beaker/shared/host_manager.rb
@@ -98,7 +98,7 @@ module Beaker
             # should warn here
             # check if logger is defined in this context
             if ( cur_logger = (logger || @logger ) )
-              cur_logger.warn "Attempting to execute against an empty array of hosts (#{hosts}, filtered to #{block_hosts}), no execution will occur"
+              cur_logger.info "Attempting to execute against an empty array of hosts (#{hosts}, filtered to #{block_hosts}), no execution will occur"
             end
           end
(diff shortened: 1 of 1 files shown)


kevpl:
@pondohva, thanks for submitting this patch!

while I agree that printing `block_hosts` in the empty array case doesn't look useful, I like having the hosts array printed to get a reminder of what was cut.  Why cut `hosts` as well?

pondohva:
@kevpl I was too lazy to do that. Ok, I'll look into hosts dump

kevpl:
@pondohva I was actually referring to not cutting as much, no additional work needed.  So that string would become something like:

    against an empty array of hosts (filtered from #{hosts})

anodelman:
I think that we can keep the hosts dump, but change the message to an info?

pondohva:
Well, finally fix it, sorry for delay. BTW, I commented to jira with loooong logger string

kevpl:
@pondohva ok, you definitely get different output than we do.  Rather than creating the array implicitly, you could make it explicit with something like this:

    hosts.map { |h| h.name }

If you do decide to continue fixing this, please make sure to add a spec test as well.

Thanks,
Kevin

pondohva:
@kevpl sure, I'll try to fix this as soon as I can. Could you merge this PR with severity decreasing, please? I'd like to see it in next beaker release :)

kevpl:

PR #963, opened 2015-09-21
(BKR-545) Use p5p for Solaris 11 puppet-agent pkgs

johnduarte:
This commit updates install_utils/foss_utils to use `p5p` as
the expected package suffix for puppet-agent when the platform
is Solaris 11. Solaris 10 will continue to use `pkg.gz` as its
expected suffix.

### DIFF
--- a/lib/beaker/dsl/install_utils/foss_utils.rb
+++ b/lib/beaker/dsl/install_utils/foss_utils.rb
@@ -1153,14 +1153,36 @@ module Beaker
               if arch == 'x86_64'
                 arch = 'i386'
               end
+              release_path << "solaris/#{version}/#{opts[:puppet_collection]}"
+              solaris_revision_conjunction = '-'
+              revision = '1'
               if version == '10'
                 # Solaris 10 uses / as the root user directory. Solaris 11 uses /root.
                 onhost_copy_base = '/'
+                solaris_release_version = ''
+                pkg_suffix = 'pkg.gz'
+                solaris_name_conjunction = '-'
+                component_version = opts[:puppet_agent_version]
+              elsif version == '11'
+                # Ref:
+                # http://www.oracle.com/technetwork/articles/servers-storage-admin/ips-package-versioning-2232906.html
+                #
+                # Example to show package name components:
+                #   Full package name: puppet-agent@1.2.5.38.6813,5.11-1.sparc.p5p
+                #   Schema: <component-name><solaris_name_conjunction><component_version><solaris_release_version><solaris_revision_conjunction><revision>.<arch>.<pkg_suffix>
+                solaris_release_version = ',5.11' # injecting comma to prevent from adding another var
+                pkg_suffix = 'p5p'
+                solaris_name_conjunction = '@'
+                component_version = opts[:puppet_agent_version]
+                component_version.gsub!(/[a-zA-Z]/, '')
+                component_version.gsub!(/(^-)|(-$)/, '')
+                # Here we strip leading 0 from version components but leave
+                # singular 0 on their own.
+                component_version = component_version.split('.').map(&:to_i).join('.')
               end
-              release_path << "solaris/#{version}/#{opts[:puppet_collection]}"
-              release_file = "puppet-agent-#{opts[:puppet_agent_version]}-1.#{arch}.pkg.gz"
+              release_file = "puppet-agent#{solaris_name_conjunction}#{component_version}#{solaris_release_version}#{solaris_revision_conjunction}#{revision}.#{arch}.#{pkg_suffix}"
               if not link_exists?("#{release_path}/#{release_file}")
-                release_file = "puppet-agent-#{opts[:puppet_agent_version]}.#{arch}.pkg.gz"
+                release_file = "puppet-agent#{solaris_name_conjunction}#{component_version}#{solaris_release_version}.#{arch}.#{pkg_suffix}"
               end
             else
               raise "No repository installation step for #{variant} yet..."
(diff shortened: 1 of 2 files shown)


haus:
I'm not sure if this is covered elsewhere, but versioning for solaris 11 is super funky. Our p5p repos look like this:

```
puppet-agent@1.0.1.786.477,5.11-1.i386.p5p
puppet-agent@1.2.5.38.6813,5.11-1.sparc.p5p
puppet-agent@1.2.5.38.6813,5.11-1.i386.p5p
puppet-agent@1.2.5.38.6813,5.11-1.sparc.p5p
puppet-agent@1.2.5.38.6813,5.11-1.i386.p5p
puppet-agent@1.0.1.808.58495,5.11-1.sparc.p5p
puppet-agent@1.0.1.808.58495,5.11-1.i386.p5p
puppet-agent@1.0.1.807.183550,5.11-1.i386.p5p
```

johnduarte:
@anodelman do not merge this, it looks like @haus has assigned me some more work here. :)

johnduarte:
@haus feel free to take a look and see if this makes sense for the IPS style of package naming.

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -1153,15 +1153,33 @@ def install_puppet_agent_dev_repo_on( hosts, opts )
...
+                # Here we strip leading 0 from version components but leave
+                # singular 0 on their own.
+                component_version = component_version.split('.').map(&:to_i).join('.')
               end
-              release_path << "solaris/#{version}/#{opts[:puppet_collection]}"
-              release_file = "puppet-agent-#{opts[:puppet_agent_version]}-1.#{arch}.pkg.gz"
-              if not link_exists?("#{release_path}/#{release_file}")
-                release_file = "puppet-agent-#{opts[:puppet_agent_version]}.#{arch}.pkg.gz"
-              end
+                release_file = "puppet-agent#{solaris_name_conjunction}#{component_version}#{solaris_release_version}#{solaris_revision_conjunction}#{revision}.#{arch}.#{pkg_suffix}"
```
haus: I think the tabbing got a little funky here.

haus:
@johnduarte i think the tabbing is a little off on the new section but the logic looks good to me. Might be worth adding a test to make sure that non-final versions behave as expected as well, but I have no idea if that's in scope for this PR or not.

kevpl:

kevpl:
@johnduarte is everything here ready to be looked at and tested, or is there still work to be done on this?

johnduarte:
@kevpl this is ready to be reviewed now, thanks.

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -1153,14 +1153,32 @@ def install_puppet_agent_dev_repo_on( hosts, opts )
...
+                component_version = opts[:puppet_agent_version]
+              elsif version == '11'
+                # Ref:
+                # http://www.oracle.com/technetwork/articles/servers-storage-admin/ips-package-versioning-2232906.html
+                solaris_release_version = ',5.11' # injecting comma to prevent from adding another var
+                pkg_suffix = 'p5p'
+                solaris_name_conjunction = '@'
+                component_version = opts[:puppet_agent_version]
+                component_version.gsub!(/[a-zA-Z]/, '')
+                component_version.gsub!(/(^-)|(-$)/, '')
```
anodelman: I'm a little confused as to what this is doing, since puppet-agent versions can be of the form '1.1.1.225.gd33
77fe' - after processing that turns into a version of nil.  Can you give some examples here of expected input/output?

haus: I think gsub! is probably the wrong thing and it should be `version = version.gsub(...)` instead.

haus: @anodelman some examples of how IPS version munging works can be found here: https://github.com/puppetlabs/vanagon/blob/master/spec/lib/vanagon/platform/solaris_11_spec.rb#L11-L23

johnduarte: Thanks @haus . I will add these to the spec tests.

cc @anodelman

anodelman: okay, so these are specific solaris 11 style puppet-agent version strings?

haus: @anodelman yes. solaris 11 versioning only allows numbers and . in the version string. no leading or trailing - is allowed, and leading 0 in a version field are also disallowed.

johnduarte:
@anodelman thanks to the help of @colinPL this should be fairly ready. @colinPL has suggested additional documentation that I will add shortly

lib/beaker/dsl/install_utils/foss_utils.rb
haus: @johnduarte did you see the previous comments about using `component_version = component_version.gsub` instead of `component_version.gsub!`? `gsub!` returns nil when no substitutions were made, which probably isn't what you want here.

johnduarte:
@anodelman , @colinPL , @haus I think this is buttoned up. Please provide any additional feedback.

haus:
@johnduarte did you see the previous comments about using `component_version = component_version.gsub` instead of `component_version.gsub!`? `gsub!` returns nil when no substitutions were made[1], which probably isn't what you want here.

[1] - http://apidock.com/ruby/v1_9_3_392/String/gsub%21

johnduarte:
@haus Sorry, I did not address your concern earlier. The gsub! makes the substitution in place and we are not using the return value for any assignment, so it is behaving as expected.

haus:
@johnduarte ah yes. don't mind me. i can't read irb correctly today (or last week)

anodelman:
@johnduarte Is this ready to go?

johnduarte:
@anodelman :+1:

PR #969, opened 2015-09-26
(BKR-554) Install AIX packages via install repo

johnduarte:
This commit adds support for installing AIX 'packages' via the
install repo logic. AIX does not support repository management
and can only install 'RPM' files directly. Since this logic
is contained in the install repo logic in Beaker, we can use
this to install a package that has been mirrored to the standard
repo location.

This assumes that the package is mirrored to the repo location.
In other words, that the following files are the same.
  * http://builds.delivery.puppetlabs.net/puppet-agent/1214e51d63b84a82df0c55cab99abc2a3f90a597/artifacts/aix/7.1/PC1/ppc/puppet-agent-1.2.5.49.g1214e51-1.aix7.1.ppc.rpm
  * http://builds.delivery.puppetlabs.net/puppet-agent/1214e51d63b84a82df0c55cab99abc2a3f90a597/repos/aix/7.1/PC1/ppc/puppet-agent-1.2.5.49.g1214e51-1.aix7.1.ppc.rpm

It also assumes that the desired package is not defined for AIX
in the acceptance pre-suite for the project.

### DIFF
--- a/lib/beaker/dsl/install_utils/foss_utils.rb
+++ b/lib/beaker/dsl/install_utils/foss_utils.rb
@@ -1111,6 +1111,10 @@ module Beaker
               variant = ((variant == 'centos') ? 'el' : variant)
               release_path << "#{variant}/#{version}/#{opts[:puppet_collection]}/#{arch}"
               release_file = "puppet-agent-#{opts[:puppet_agent_version]}-1.#{variant}#{version}.#{arch}.rpm"
+            when /^(aix)$/
+              if arch == 'power' then arch = 'ppc' end
+              release_path << "#{variant}/#{version}/#{opts[:puppet_collection]}/#{arch}"
+              release_file = "puppet-agent-#{opts[:puppet_agent_version]}-1.#{variant}#{version}.#{arch}.rpm"
             when /^(debian|ubuntu|cumulus)$/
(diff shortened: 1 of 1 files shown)


lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -1173,6 +1177,25 @@ def install_puppet_agent_dev_repo_on( hosts, opts )
             case variant
             when /^(fedora|el|centos|sles)$/
               on host, "rpm -ivh #{onhost_copied_file}"
+            when /^(aix)$/
+              # NOTE: AIX does not support repo management. This block assumes
+              # that the desired rpm has been mirrored to the 'repos' location.
+              #
+              # NOTE: tar is a dependency for puppet packages on AIX. So,
+              # we install it prior to the 'repo' file.
+              tar_pkg_path = "ftp://ftp.software.ibm.com/aix/freeSoftware/aixtoolbox/RPMS/ppc/tar"
```
anodelman: Installing tar should probably be something that we do during the configuration stage of beaker - but since we are under a time crunch I'll file a bug to move this around and we can go forward with the current patch.

anodelman:
Test failures unrelated to patch.

PR #970, opened 2015-09-28
(BKR-545) Dup solaris puppet_agent_version

johnduarte:
This commit dups the `opts[:puppet_agent_version]` in the foss_utils to
prevent a `RuntimeError: can't modify frozen String` error


anodelman:
:+1:

PR #972, opened 2015-09-30
(BKR-380) Fix '3389' (host port) is declared multiple times

azhurbilo:
(BKR-380) beaker vagrant windows box support issues

Without "auto_correct" we got error if source Vagrantfile of box already contains this port forwarding.

```
Error:
vm:
* Forwarded port '3389' (host port) is declared multiple times
with the protocol 'tcp'.
```

I think we need add this change as most part of Windows boxes in Vagrant Atlas already have this port forwarding:
- https://github.com/joefitzgerald/packer-windows/blob/master/vagrantfile-windows_2012_r2.template
- https://github.com/boxcutter/windows/blob/master/tpl/vagrantfile-eval-win2012r2-standard.tpl


anodelman:
ok to test

anodelman:
Failure unrelated to patch.

PR #977, opened 2015-10-06
(QENG-2604) Add Host Tags

colinPL:
This commit adds the ability to specify arbitrary tags and have them applied to AWS or vmpooler instances. The configuration parameter is called "host_tags" to avoid confusion with test tagging. Host tags can be specified at the host or CONFIG level.

The presets have been modified to pre-populate "host_tags" with: project, department, and created_by.

Example:
```
HOSTS:
  'agent1':
    roles:
      - agent
    host_tags:
      created_by: 'me'
      some_key: 'yes it is'
CONFIG:
    host_tags:
      project: 'beaker'
```

### DIFF
--- a/lib/beaker/options/presets.rb
+++ b/lib/beaker/options/presets.rb
@@ -121,11 +121,7 @@ module Beaker
           :project                => 'Beaker',
           :department             => 'unknown',
           :created_by             => ENV['USER'] || ENV['USERNAME'] || 'unknown',
-          :host_tags              => {
-                                      :project    => 'Beaker',
-                                      :department => 'unknown',
-                                      :created_by => ENV['USER'] || ENV['USERNAME'] || 'unknown'
-          },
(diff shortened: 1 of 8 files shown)


lib/beaker/options/presets.rb
```diff
@@ -121,6 +121,11 @@ def presets
           :project                => 'Beaker',
           :department             => 'unknown',
           :created_by             => ENV['USER'] || ENV['USERNAME'] || 'unknown',
+          :host_tags              => {
```
anodelman: I believe that you are going to have to do some work here to get the env var support for these values to still work correctly, otherwise they will get stored as :department instead of host_tags[:department].

kevpl:

kevpl:
@colinPL sorry for the late notice, but "additional spec testing required"

PR #980, opened 2015-10-07
(maint) Fix Solaris 11 package name logic

johnduarte:
This commit fixes a bug in the Solaris 11 package naming logic
that dropped the SHA suffix entirely from the final package name.
The SHA should be incorporated into the package name.

An additional spec test has been added for the expected package
name in this circumstance.


kevpl:
@johnduarte we're not going to be merging `(maint)` PRs going forward, so please keep that in mind for future beaker work

PR #981, opened 2015-10-09
(BKR-576) scp_to does not handle dotfiles

GeoffWilliams:
Use find instead of shell glob to obtain a listing of the FILES within a directory structure.  We must exclude empty directories to prevent attempting to SCP them (this will also cause test failures).  This code now supports dotfiles and files within hidden dotfile directories, eg `.git`

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -2,6 +2,7 @@ require 'socket'
 require 'timeout'
 require 'benchmark'
 require 'rsync'
+require 'find'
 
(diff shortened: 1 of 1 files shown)


GeoffWilliams:
Test failure seems unrelated

kevpl:
@GeoffWilliams please include some spec testing for this

kevpl:
last comment still current

kevpl:
last comment still current.  If this PR has not been updated in a week, it will be closed from inactivity.

PR #978, opened 2015-10-06
(BKR-571) apply_manifest_on always returns an array

kevpl:

PR #982, opened 2015-10-12
(BKR-571) updated documentation on return types

kevpl:
tried in #978 to update to a simple array of results in `apply_manifest_on`, but that breaks compatibility with current uses (check there for more details).

Since that's the case, I've updated the docs appropriately for current usage, & added a note for update during Beaker 3.0 changes later.


sschneid:
:+1:

waynr:
:+1:

PR #983, opened 2015-10-13
(BKR-579) rpm calls to respect package_proxy opt

alexharv074:
Without this patch applied, the call to rpm in Beaker::DSL::InstallUtils ignores the package_proxy option.
    
We add two new methods, extract_rpm_proxy_options and install_package_with_rpm.
    
The extract_rpm_proxy_options method is needed as the rpm command expects command line options --httpproxy and --httpport.
    
The install_package_with_rpm method is needed in order to install rpms from a remote URL address.  The existing install_package method uses yum and yum can't install an rpm from a remote URL.
    
Spec tests are provided for the extract_rpm_proxy_options method.  A redundant spec test is deleted.

### DIFF
--- a/lib/beaker/host/unix/pkg.rb
+++ b/lib/beaker/host/unix/pkg.rb
@@ -120,6 +120,24 @@ module Unix::Pkg
     end
   end
 
+  # Install a package using RPM
+  #
+  # @param [String] name          The name of the package to install.  It
+  #                               may be a filename or a URL.
+  # @param [String] cmdline_args  Additional command line arguments for
+  #                               the package manager.
+  # @option opts [String] :package_proxy  A proxy of form http://host:port
+  # 
+  # @return nil
+  # @api public
+  def install_package_with_rpm(name, cmdline_args = '', opts = {})
+    proxy = ''
+    if name =~ /^http/ and opts[:package_proxy]
+      proxy = extract_rpm_proxy_options(opts[:package_proxy])
+    end
+    execute("rpm #{cmdline_args} -ivh #{name} #{proxy}")
+  end
+
   def uninstall_package(name, cmdline_args = '', opts = {})
     case self['platform']
       when /sles-/
@@ -249,6 +267,24 @@ module Unix::Pkg
     else
       result = exec(Beaker::Command.new("arch | grep x86_64"), :accept_all_exit_codes => true)
       result.exit_code == 0
+    end
+  end
+
+  # Extract RPM command's proxy options from URL
+  #
+  # @param [String] url  A URL of form http://host:port
+  # @return [String]     httpproxy and httport options for rpm
+  #
+  # @raise [StandardError] When encountering a string that 
+  #                        cannot be parsed
+  # @api private
+  def extract_rpm_proxy_options(url)
+    begin
+      host, port = url.match(/https?:\/\/(.*):(\d*)/)[1,2]
+      raise if host.empty? or port.empty?
+      "--httpproxy #{host} --httpport #{port}"
+    rescue
+      raise "Cannot extract host and port from '#{url}'"
     end
   end
 
--- a/spec/beaker/dsl/install_utils/foss_utils_spec.rb
+++ b/spec/beaker/dsl/install_utils/foss_utils_spec.rb
@@ -396,7 +396,7 @@ describe ClassMixedWithDSLInstallUtils do
     context 'on el-6' do
       let(:platform) { Beaker::Platform.new('el-6-i386') }
       it 'installs' do
-        expect(subject).to receive(:on).with(hosts[0], /puppetlabs-release-el-6\.noarch\.rpm/)
+        expect(hosts[0]).to receive(:install_package_with_rpm).with(/puppetlabs-release-el-6\.noarch\.rpm/, '--replacepkgs', {:package_proxy=>false})
         expect(hosts[0]).to receive(:install_package).with('puppet')
         subject.install_puppet
       end
@@ -414,7 +414,7 @@ describe ClassMixedWithDSLInstallUtils do
     context 'on el-5' do
       let(:platform) { Beaker::Platform.new('el-5-i386') }
       it 'installs' do
-        expect(subject).to receive(:on).with(hosts[0], /puppetlabs-release-el-5\.noarch\.rpm/)
+        expect(hosts[0]).to receive(:install_package_with_rpm).with(/puppetlabs-release-el-5\.noarch\.rpm/, '--replacepkgs', {:package_proxy=>false})
         expect(hosts[0]).to receive(:install_package).with('puppet')
         subject.install_puppet
       end
@@ -422,7 +422,7 @@ describe ClassMixedWithDSLInstallUtils do
     context 'on fedora' do
       let(:platform) { Beaker::Platform.new('fedora-18-x86_84') }
       it 'installs' do
-        expect(subject).to receive(:on).with(hosts[0], /puppetlabs-release-fedora-18\.noarch\.rpm/)
+        expect(hosts[0]).to receive(:install_package_with_rpm).with(/puppetlabs-release-fedora-18\.noarch\.rpm/, '--replacepkgs', {:package_proxy=>false})
         expect(hosts[0]).to receive(:install_package).with('puppet')
         subject.install_puppet
       end
@@ -580,16 +580,6 @@ describe ClassMixedWithDSLInstallUtils do
         expect(subject).to receive(:on).with( host, /wget .*/ ).ordered
         expect(subject).to receive(:on).with( host, /dpkg .*/ ).ordered
         expect(subject).to receive(:on).with( host, "apt-get update" ).ordered
-        subject.install_puppetlabs_release_repo host
-      end
-
-    end
-
-    describe "When host is a redhat-like platform" do
-      let( :platform ) { Beaker::Platform.new('el-7-i386') }
-
-      it "installs an rpm" do
-        expect(subject).to receive(:on).with( host, /^(rpm --replacepkgs -ivh).*/ ).ordered
         subject.install_puppetlabs_release_repo host
       end
 
--- a/spec/beaker/host/unix/pkg_spec.rb
+++ b/spec/beaker/host/unix/pkg_spec.rb
@@ -145,6 +145,48 @@ module Beaker
 
     end
 
+    context "install_package_with_rpm" do
+
+      it "accepts a package as a single argument" do
+        @opts = {'platform' => 'el-is-me'}
+        pkg = 'redhat_package'
+        expect( Beaker::Command ).to receive(:new).with("rpm  -ivh #{pkg} ", [], {:prepend_cmds=>nil, :cmdexe=>false}).and_return('')
+        expect( instance ).to receive(:exec).with('', {}).and_return(generate_result("hello", {:exit_code => 0}))
+        expect( instance.install_package_with_rpm(pkg) ).to be == "hello"
+      end
+
+      it "accepts a package and additional options" do
+        @opts = {'platform' => 'el-is-me'}
+        pkg = 'redhat_package'
+        cmdline_args = '--foo'
+        expect( Beaker::Command ).to receive(:new).with("rpm #{cmdline_args} -ivh #{pkg} ", [], {:prepend_cmds=>nil, :cmdexe=>false}).and_return('')
+        expect( instance ).to receive(:exec).with('', {}).and_return(generate_result("hello", {:exit_code => 0}))
+        expect( instance.install_package_with_rpm(pkg, cmdline_args) ).to be == "hello"
+      end
+
+    end
+
+    context 'extract_rpm_proxy_options' do
+
+      [ 'http://myproxy.com:3128/',
+        'https://myproxy.com:3128/',
+        'https://myproxy.com:3128',
+        'http://myproxy.com:3128',
+      ].each do |url|
+        it "correctly extracts rpm proxy options for #{url}" do
+          expect( instance.extract_rpm_proxy_options(url) ).to be == '--httpproxy myproxy.com --httpport 3128'
+        end
+      end
+
+      url = 'http:/myproxy.com:3128'
+      it "fails to extract rpm proxy options for #{url}" do
+        expect{
+          instance.extract_rpm_proxy_options(url)
+        }.to raise_error(RuntimeError, /Cannot extract host and port/)
+      end
+
+    end
+
   end
 end
 
--- a/lib/beaker/dsl/install_utils/foss_utils.rb
+++ b/lib/beaker/dsl/install_utils/foss_utils.rb
@@ -861,11 +861,10 @@ module Beaker
             case variant
             when /^(fedora|el|centos)$/
               variant = (($1 == 'centos') ? 'el' : $1)
+              remote = "%s/puppetlabs-release%s-%s-%s.noarch.rpm" % [opts[:release_yum_repo_url],
+		                                                       repo_name, variant, version]
 
-              rpm = "puppetlabs-release%s-%s-%s.noarch.rpm" % [repo_name, variant, version]
-              remote = URI.join( opts[:release_yum_repo_url], rpm )
-
-              on host, "rpm --replacepkgs -ivh #{remote}"
+	      host.install_package_with_rpm(remote, '--replacepkgs', {:package_proxy => opts[:package_proxy]})
 
             when /^(debian|ubuntu|cumulus)$/
               deb = "puppetlabs-release%s-%s.deb" % [repo_name, codename]


alexharv074:
@anodelman Any chance someone can look at this?

kevpl:
Hi @alexharv074. Since these changes are specifically for handling the details of RPM, I think the appropriate place for them would be in [`unix/pkg.rb`](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host/unix/pkg.rb).

Let me know if you have any questions or comments about this.

Thanks,
KI

alexharv074:
Hi @kevpl.  I think I understand, although involves a bit of refactoring.

There are quite a number of direct calls to rpm (not to mention dpkg etc) in Beaker::DSL::InstallUtils and Beaker::HostPrebuiltSteps:

    alexs-mbp:beaker-2.23.0 alexharvey$ grep -r 'rpm ' lib/
    lib//beaker/dsl/install_utils/foss_utils.rb:        # Installs Puppet and dependencies using rpm on provided host(s).
    lib//beaker/dsl/install_utils/foss_utils.rb:              rpm = "puppetlabs-release%s-%s-%s.noarch.rpm" % [repo_name, variant, version]
    lib//beaker/dsl/install_utils/foss_utils.rb:              remote = URI.join( opts[:release_yum_repo_url], rpm )
    lib//beaker/dsl/install_utils/foss_utils.rb:              on host, "rpm --replacepkgs -ivh #{remote} #{proxy}"
    lib//beaker/dsl/install_utils/foss_utils.rb:            find_command  = 'rpm -ivh'
    lib//beaker/dsl/install_utils/foss_utils.rb:              on host, "rpm -ivh #{onhost_copied_file}"
    lib//beaker/dsl/install_utils/foss_utils.rb:              # that the desired rpm has been mirrored to the 'repos' location.
    lib//beaker/dsl/install_utils/foss_utils.rb:              on host, "rpm -ivh #{onhost_copied_tar_file}"
    lib//beaker/dsl/install_utils/foss_utils.rb:              on host, "rpm -ivh #{onhost_copied_file}"
    lib//beaker/dsl/install_utils/foss_utils.rb:              on host, "rpm -ihv #{onhost_copied_file}"
    lib//beaker/host/unix/pkg.rb:        result = execute("rpm -q #{name}", opts) { |result| result }
    lib//beaker/host_prebuilt_steps.rb:    # @option opts [Boolean] :debug If true, print verbose rpm information when installing EPEL
    lib//beaker/host_prebuilt_steps.rb:          result = host.exec(Command.new('rpm -qa | grep epel-release'), :acceptable_exit_codes => [0,1])
    lib//beaker/host_prebuilt_steps.rb:              host.exec(Command.new("rpm -i#{debug_opt} #{url}/#{arch}/e/#{pkg}"))
    lib//beaker/host_prebuilt_steps.rb:              host.exec(Command.new("rpm -i#{debug_opt} #{url}/#{arch}/#{pkg}"))

And then dpkg:

    alexs-mbp:beaker-2.23.0 alexharvey$ grep -r 'dpkg ' lib/
    lib//beaker/dsl/install_utils/foss_utils.rb:              on host, "dpkg -i --force-all /tmp/puppet.deb"
    lib//beaker/dsl/install_utils/foss_utils.rb:            find_command  = 'dpkg -i'
    lib//beaker/dsl/install_utils/foss_utils.rb:              on host, "dpkg -i --force-all #{onhost_copied_file}"
    lib//beaker/dsl/install_utils/foss_utils.rb:              on host, "dpkg -i --force-all #{onhost_copied_file}"
    lib//beaker/host/unix/pkg.rb:        result = execute("dpkg -s #{name}", opts) { |result| result }

Now, ideally, we could just have a platform independent method 'install_package' but of course, we already do - but it uses yum on Redhat-based platforms.

The problem of rpm vs yum is itself an ugly legacy issue with the Redhat platform.  

What if I provide a new method in unix/pkg.rb:

    def rpm_install_package(package, additional_args = '')
      ...
    end

If the package looks like a URL it will attempt to install remotely and automatically respect opts[:package_proxy].  An option like --replacepkgs could be passed in optionally in additional_args.

How does this sound?

kevpl:
@alexharv074, agreed, that stuff is spread everywhere.  It's an old pattern that was just copy-pasted to get things done quickly, I'm pretty sure.  It's definitely something I'd like to get away from going forward.

I think this is a good idea. My two caveats would be to make sure that people can pass in similar options to `install_package` like `cmdline_args` (the more general, the better), and it would be nice for review purposes if the added method & spec functionality were separate commits from any refactors
 that you do for this.

alexharv074:
@kevpl I am stuck on these spec tests.  

    Failed examples:
    
    rspec ./spec/beaker/dsl/install_utils/foss_utils_spec.rb:398 # ClassMixedWithDSLInstallUtils#install_puppet on el-6 installs
    rspec ./spec/beaker/dsl/install_utils/foss_utils_spec.rb:416 # ClassMixedWithDSLInstallUtils#install_puppet on el-5 installs
    rspec ./spec/beaker/dsl/install_utils/foss_utils_spec.rb:424 # ClassMixedWithDSLInstallUtils#install_puppet on fedora installs
    rspec ./spec/beaker/dsl/install_utils/foss_utils_spec.rb:591 # ClassMixedWithDSLInstallUtils#install_puppetlabs_release_repo When host is a redhat-like platform installs an rpm

I can't get my head around what these tests are trying to do; thus struggling to fix them.  Any help appreciated.

alexharv074:
@kevpl I confirmed that my code works by the way; and I intend to rebase of course.

kevpl:
@alexharv074, to go through the test on line [398](https://github.com/puppetlabs/beaker/blob/master/spec/beaker/dsl/install_utils/foss_utils_spec.rb#L398) to give you some idea of the tests:

- [line 399](https://github.com/puppetlabs/beaker/blob/master/spec/beaker/dsl/install_utils/foss_utils_spec.rb#L399) really tests that [`install_puppetlabs_release_repo`](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/install_utils/foss_utils.rb#L443) sets the [right package format](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/install_utils/foss_utils.rb#L865)
- [line 400](https://github.com/puppetlabs/beaker/blob/master/spec/beaker/dsl/install_utils/foss_utils_spec.rb#L400) tests that [`install_puppet_from_rpm_on`](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/install_utils/foss_utils.rb#L454) calls `install_package` correctly

The other examples are all similar.  Let me know if you still have questions about this.

alexharv074:
@kevpl It made more sense after sleeping. :)  Ok, let me know your thoughts on the updated PR.  Since I haven't really done any refactoring, it's all in one commit.  The issue with extensive refactoring to use my new method is that I found most of the calls to RPM are actually on the AIX platform and I can't test that here, so felt it better to leave it alone.

lib/beaker/host/unix/pkg.rb
```diff
@@ -120,6 +120,34 @@ def install_package(name, cmdline_args = '', version = nil, opts = {})
     end
   end
 
+  # Extract RPM command's proxy options from URL
+  #
+  # @param [Url] url A URL of form http://host:port
+  #
+  def extract_rpm_proxy_options(url)
```
kevpl: I know that you do guard against `opts[:package_proxy]` being undefined in `install_package_with_rpm`, but it would be even safer if you guarded against anyone passing nil in by accident.

If you consider this dealt with in the catching of the `url.match` failure, then I'm down with that choice.  If that's the case, it might be better to put quotes around the `#{url}` part of the exception, so that it's easier to see it's nil rather than wondering if there might be a newline in there or something.

lib/beaker/host/unix/pkg.rb
```diff
@@ -120,6 +120,34 @@ def install_package(name, cmdline_args = '', version = nil, opts = {})
...
+    end
+  end
+
+  # Install a package using RPM
+  #
+  # @param [String] name          The name of the package to install.  It may be a
+  #                               filename or a URL.
+  # @param [String] cmdline_args  Additional command line arguments for
+  #                               the package manager.
+  def install_package_with_rpm(name, cmdline_args = '', opts = {})
```
kevpl: please include a doc section for the opts parameters that are taken by this method.  You can see an example of this [here](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/install_utils/foss_utils.rb#L243-L247).

Also, please include a `@return` doc for both methods.

spec/beaker/dsl/install_utils/foss_utils_spec.rb
```diff
@@ -396,7 +396,7 @@ def logger
     context 'on el-6' do
       let(:platform) { Beaker::Platform.new('el-6-i386') }
       it 'installs' do
-        expect(subject).to receive(:on).with(hosts[0], /puppetlabs-release-el-6\.noarch\.rpm/)
+        expect(hosts[0]).to receive(:install_package_with_rpm).with('http://yum.puppetlabs.com/puppetlabs-release-el-6.noarch.rpm', '--replacepkgs', {:package_proxy=>false})
```
kevpl: since we're not actually testing the entire URL that's being called, but just the form of the package name itself, it would be better to leave the regex as it was, rather than being so specific about it.  Let me know if this isn't the case for some reason, and we'll continue to talk about this.

spec/beaker/dsl/install_utils/foss_utils_spec.rb
```diff
@@ -414,7 +414,7 @@ def logger
     context 'on el-5' do
       let(:platform) { Beaker::Platform.new('el-5-i386') }
       it 'installs' do
-        expect(subject).to receive(:on).with(hosts[0], /puppetlabs-release-el-5\.noarch\.rpm/)
+        expect(hosts[0]).to receive(:install_package_with_rpm).with('http://yum.puppetlabs.com/puppetlabs-release-el-5.noarch.rpm', '--replacepkgs', {:package_proxy=>false})
```
kevpl: if you're not testing the `:package_proxy` option, then you shouldn't need to provide it at all to this method.  It would give more confidence in this functionality if you didn't actually include that option, even if setting it to be false should create the same behavior.

lib/beaker/host/unix/pkg.rb
alexharv074: Yes, I do consider that dealt with.

As guarding against passing in nil, I consider this a private method; so no one should call it other than the install_package_with_rpm method so I think it is a better balance to keep the code clean and not add logic to handle input that can never be passed in?

Perhaps I should explicitly mark this method private and add @api private as I see has been done in windows/pkg.rb.

Happy to add quotes around url.

lib/beaker/host/unix/pkg.rb
alexharv074: No problem.

spec/beaker/dsl/install_utils/foss_utils_spec.rb
alexharv074: Whether expecting a match on part of the RPM filename or the entire URL, I feel there must be a way of designing these tests where we don't need to make assumptions about URLs or filenames.  However, rspec is a bit of a black art to me so I don't know how to improve upon it.  Happy to make this change as you suggest.

spec/beaker/dsl/install_utils/foss_utils_spec.rb
alexharv074: But I have this line in install_puppetlabs_release_repo_on

    host.install_package_with_rpm(remote, '--replacepkgs', {:package_proxy => opts[:package_proxy]})

So the expectation is that the method under test is going to pass {:package_proxy => opts[:package_proxy]} and opts[:package_proxy] will be false.

Perhaps I should be more explicit by setting it to false before the test?

lib/beaker/host/unix/pkg.rb
kevpl: yeah, quotes and `@api private` would be great!  Thanks!

spec/beaker/dsl/install_utils/foss_utils_spec.rb
kevpl: `install_package_with_rpm` will be used by more people than just that one usage, especially as we go through refactoring, getting other parts of the system to use this method to eliminate much of our rpm usage duplication.

I think there is an implication in an `opts` hash parameter that the options are _optional_, so if this is not, I would suggest either making it a real argument of the `install_package_with_rpm` method, or making sure that you strongly point out that it will be expected in usage every time in the doc.

alexharv074: Sorry, one of us is misunderstanding something; it could be me!

My spec test is saying: *I expect that when the `install_puppet` method is called, an options hash `{:package_proxy=>false}` will be passed to the `install_package_with_rpm` method*.  I am not saying that the `install_package_with_rpm` requires an options hash to be passed (it doesn't).  I am testing the behaviour of `install_puppetlabs_release_repo_on` here; not the `install_package_with_rpm` method.

Now you could question the value of this test; I also question the value of this test and some of the other tests in this class; but I have only updated a test that I inherited after I refactored the code.

Note that I have not provided a test of the `install_package_with_rpm` method at all.  The reason is there were no tests for the [install_package](https://github.com/puppetlabs/beaker/blob/master/spec/beaker/host/unix/pkg_spec.rb#L128) method for me to copy. :)

If you feel I really must add tests for my method (it doesn't do much other than call `execute` which is tested elsewhere) let me know.

alexharv074: Ok, you convinced me to do the right thing and add some tests for `install_package_with_rpm`. :)

lib/beaker/host/unix/pkg.rb
alexharv074: Interestingly, it looks like it's not really possible to use rspec to test private methods  https://www.ruby-forum.com/topic/197346

I am a bit surprised; I think you should be able to test private methods.

Anyhow the workaround I have chosen is to mark the method as private in the documentation only, although you may have a better suggestion.

lib/beaker/host/unix/pkg.rb
alexharv074: Interestingly, it looks like it's not really possible to use rspec to test private methods https://www.ruby-forum.com/topic/197346

I am a bit surprised; I think you should be able to test private methods.

Anyhow the workaround I have chosen is to mark the method as private in the documentation only, although you may have a better suggestion.

spec/beaker/host/unix/pkg_spec.rb
```diff
@@ -145,6 +145,48 @@ def exec
...
+        expect( instance.install_package_with_rpm(pkg) ).to be == "hello"
+      end
+
+      it "accepts a package and additional options" do
+        @opts = {'platform' => 'el-is-me'}
+        pkg = 'redhat_package'
+        cmdline_args = 'foo'
+        expect( Beaker::Command ).to receive(:new).with("rpm #{cmdline_args} -ivh #{pkg} ", [], {:prepend_cmds=>nil, :cmdexe=>false}).and_return('')
+        expect( instance ).to receive(:exec).with('', {}).and_return(generate_result("hello", {:exit_code => 0}))
+        expect( instance.install_package_with_rpm(pkg, cmdline_args) ).to be == "hello"
```
alexharv074: @kevpl This test shows that `install_package_with_rpm` can be called without an options hash.

spec/beaker/host/unix/pkg_spec.rb
```diff
@@ -145,6 +145,48 @@ def exec
...
+
+      it "accepts a package and additional options" do
+        @opts = {'platform' => 'el-is-me'}
+        pkg = 'redhat_package'
+        cmdline_args = '--foo'
+        expect( Beaker::Command ).to receive(:new).with("rpm #{cmdline_args} -ivh #{pkg} ", [], {:prepend_cmds=>nil, :cmdexe=>false}).and_return('')
+        expect( instance ).to receive(:exec).with('', {}).and_return(generate_result("hello", {:exit_code => 0}))
+        expect( instance.install_package_with_rpm(pkg, cmdline_args) ).to be == "hello"
+      end
+
```
alexharv074: @kevpl Fyi these tests show that `install_package_with_rpm` can be called without the options hash.

kevpl:
:+1:

kevpl:

electrical:
Initially i was a bit confused by the extra PR's for the proxy parts, but now i understand its for the rpm commands directly.
I only used Yum so far and with the proxy part i already write the yum config for enabling proxy :-)

alexharv074:
I am not able to follow the link to Jenkins.  All the tests pass when I run them; are there really tests failing?

alexharv074:
@kevpl Any chance you could have a look into what's wrong here?

kevpl:
My apologies, it could very well be the case that these are unrelated failures to your change. :crying_cat_face: 

Unfortunately, I can't see the run due to history limits.  I'll kick off another one, & look at this later today for you (I'm also looking at increasing our limits, so that we don't have to have this delay in the future).

I'll let you know what I find.

kevpl:

kevpl:
transient failures not related to your changes.  Will retest in a bit

kevpl:

kevpl:
failures unrelated to change

alexharv074:
@kevpl How long before this will be released in a new gem?

kevpl:
@alexharv074 the next scheduled Beaker release is for Nov 4th, so it should be then or earlier.

alexharv074:
Thanks.

PR #984, opened 2015-10-13
(BKR-582) Colorization adjustments / CI Build / Allow overriding log colors

ferventcoder:
When beaker detects it is being run by Jenkins, it should adjust output
colors so that they are better handled in console output. This is done
by detecting BUILD_NUMBER environment variable, a variable that is
known to be present on Jenkins build slaves.

Adjust colorization to colors that come out better during runs. The
colors that should come up to the front are shown better when running
in the console. Additionally make it easier to see logs coming back
from the SUT with an explicit color instead of NORMAL (which is the
same setting as the current GREY).

Allow overriding log colors by passing in log_colors in options.
If a user has overridden the default colors, don't override the
colors in CI.

Also 
* Warn on skipped tests - debug is not the right location for those messages
* When beginning a new test file, use info instead of notify to separate it from other notify messages and make it easier to find the file that may have had testing errors.


ferventcoder:
This is with CI detected:

image


This is normal console colors:

image

lib/beaker/logger.rb
```diff
@@ -297,14 +305,18 @@ def success *args
     # @param args[Array<String>] Strings to be reported
     def notify *args
       return unless is_notify?
-      optionally_color BRIGHT_WHITE, *args
+      if @ci_build
```
Iristyle: This seems like an odd way to implement this.

If we intend to be able to change colors for a `@ci_build`, maybe we should just maintain a couple of hashes with the colors, and switch the active hash based on whether we're operating in CI.


```ruby
default_colors = { :success => MAGENTA, :notify => BLUE }
ci_colors =  { :success => MAGENTA, :notify => NORMAL }

colors = @ci_build ? ci_colors : default_colors
```

Then make the method calls look like:

```ruby
def success *args
  optionally_color colors[:success], *args		
end
```

ferventcoder:
Looks like it needs a rebuild - followed it down to http://jenkins-beaker.delivery.puppetlabs.net/job/qe_beaker_intn-sys_beaker-acceptance-puppet-vpool-pe/722/agent=centos7,master=centos7/console

lib/beaker/logger.rb
ferventcoder: @Iristyle I wasn't trying to fundamentally adjust the way it is now. When in Rome.

ferventcoder: Although that looks like a pretty good add to this.

ferventcoder: Not odd, more keeping it naive or primitive.

ferventcoder: I liked that. I changed it a bit on the way up though. Because we only change two of the colors for CI output, I wanted that to stand out

lib/beaker/logger.rb
```diff
@@ -83,6 +83,25 @@ def initialize(*args)
...
+        :error => RED,
+        :warn => BRIGHT_RED,
+        :success => MAGENTA,
+        :notify => BLUE,
+        :info => GREEN,
+        :debug => WHITE,
+        :trace => BRIGHT_YELLOW,
+        :perf => BRIGHT_MAGENTA,
+        :host => YELLOW
+      }
```
ferventcoder: @kevpl Keep in mind these are not the same colors that were there before. It's going to be harder to see that after adjusting for @Iristyle's earlier comments.

ferventcoder: Take a look at the images above in the PR to see what the output looks like now.

lib/beaker/logger.rb
```diff
@@ -83,6 +83,25 @@ def initialize(*args)
 
       @destinations = []
 
+      @color_levels = options[:color_levels] || {
```
ferventcoder: @Iristyle @kevpl Even better, now you can pass the color levels in the options.

ferventcoder:
### *NOTE:* This is a completely archaic set of colors, it's only meant to show you can adjust to your needs.


Even better, the ability to override with your own levels:

Options cfg file:
~~~yaml
{
    :color_levels => {
        :error => "\e[00;30m",
        :warn => "\e[00;30m",
        :success => "\e[00;30m",
        :notify => "\e[00;30m",
        :info => "\e[00;30m",
        :debug => "\e[00;31m",
        :trace => "\e[00;30m",
        :perf => "\e[00;30m",
        :host => "\e[00;34m"
    }
}
~~~

Output above now looks like:

image

lib/beaker/logger.rb
Iristyle: I like the idea here, though you probably want to use a merge, not a wholesale replacement, so that users can partially override.

lib/beaker/logger.rb
```diff
@@ -83,6 +83,25 @@ def initialize(*args)
...
+        :trace => BRIGHT_YELLOW,
+        :perf => BRIGHT_MAGENTA,
+        :host => YELLOW
+      }
+
+      # Jenkins exposed variable - should be present on the slave directing
+      # the beaker run
+      ci_build = ENV['BUILD_NUMBER'] != nil
+
+      @color_levels[:notify] = NORMAL if ci_build
```
Iristyle: The only problem with it setup like this is that it ignores user overrides...

lib/beaker/logger.rb
ferventcoder: Good point.

lib/beaker/logger.rb
ferventcoder: I thought about this. I think right now it's definitely what is wanted. Perhaps we add an additional option not to allow CI to override colors?

lib/beaker/logger.rb
ferventcoder:
~~~yaml
    :log_colors => {
        :debug => "\e[00;30m"
    }
~~~

image

lib/beaker/logger.rb
ferventcoder: I updated this to add an option to disallow ci - `log_colors_disable_ci_override`

lib/beaker/logger.rb
ferventcoder: Repushed.

lib/beaker/logger.rb
```diff
@@ -83,6 +83,32 @@ def initialize(*args)
...
+        :success => MAGENTA,
+        :notify => BLUE,
+        :info => GREEN,
+        :debug => WHITE,
+        :trace => BRIGHT_YELLOW,
+        :perf => BRIGHT_MAGENTA,
+        :host => YELLOW
+      }
+
+      @log_colors.merge!(options[:log_colors]) if options[:log_colors]
```
er0ck: merge.  good

er0ck:
huge :+1: on this.  :shipit:

lib/beaker/logger.rb
Iristyle: I'm not sure I understand the need for 2 options for this - I think you should trust that if a user has overridden the color values, they know what they're doing.  Such changes could be part of local settings like `Gemfile.local` that aren't checked into source control.

FWIW, having Beaker ignore user specified values in an effort to "do what it thinks is best" has been a major source of headaches in the past (see `:type`).

er0ck: damn me for missing this.  i totally agree with @Iristyle here.

ferventcoder: The main purpose of this ticket is for the console color override for CI. The rest is gravy. However, I think this is probably best.

ferventcoder: Okay, fixed up.

ferventcoder: I think this is probably much further than it's original intention, which is override colors for CI so they are easy to find when reviewing logs on Jenkins. Let's not scope creep this one much further. :+1:

kevpl:

er0ck:
i can't wait to solarize my beaker output.  :sunrise: 
i guess i need to file a ticket about a global, local options file in ~/.beaker ;-)
yay, options from yet another place!

ferventcoder:
Passed and then failed.

ferventcoder:
>I guess i need to file a ticket about a global, local options file in ~/.beaker ;-)

Sounds very cool.

kevpl:
@ferventcoder looks good, but it will require some spec testing, so that we don't mess this functionality up in the future

ferventcoder:
@kevpl Let me add that :)

ferventcoder:
@kevpl This should cover all of the scenarios:

image

Pushing up in a moment to see how Jenkins actually does with it.

ferventcoder:
@kevpl it's ready to have jenkins kick off again

kevpl:
It's running spec in a Jenkins job, so the section about defaults is being overridden and that test fails:

    Failures:
    
      1) Beaker::Logger new log_colors should have the default log_colors
         Failure/Error: expect(logger.log_colors).to be == {
           expected: == {:error=>"\e[00;31m", :warn=>"\e[01;31m", :success=>"\e[00;35m",     :notify=>"\e[00;34m", :info=>"\e[00;32m", :debug=>"\e[00;37m", :trace=>"\e[01;33m", :perf=>"\e[01;35m", :host=>"\e[00;33m"}
                got:    {:error=>"\e[00;31m", :warn=>"\e[01;31m", :success=>"\e[00;35m", :notify=>"\e[00;00m", :info=>"\e[00;00m", :debug=>"\e[00;37m", :trace=>"\e[01;33m", :perf=>"\e[01;35m", :host=>"\e[00;33m"}
         # ./spec/beaker/logger_spec.rb:132:in `block (4 levels) in <module:Beaker>'

ferventcoder:
@kevpl I thought I didn't do quite enough there. :)

ferventcoder:
We'll wait for it to finish again. This should fix that. :)

ferventcoder:
`\o/`

er0ck:
bit late now, but i found this:
https://github.com/watson/is-ci/blob/master/index.js#L9
TLDR: javascript, but:
```
module.exports = !!(
  #trimmed
  process.env.JENKINS_URL
)
```

translate to ruby:
```
is_ci = ENV['JENKINS_URL']
```
maybe?

ferventcoder:
That could also work. If we find that BUILD_NUMBER is not sufficient. It also maps to TeamCity (which we are not using but other folks might).

lib/beaker/logger.rb
```diff
@@ -83,6 +83,35 @@ def initialize(*args)
...
+      # option a user can pass to be explicit about still allowing
+      # the override.
+      unless options[:log_colors]
+        # Jenkins exposed variable - should be present on the slave directing
+        # the beaker run
+        ci_build = ENV['BUILD_NUMBER'] != nil
+
+        @log_colors[:notify] = NORMAL if ci_build
+        @log_colors[:info] = NORMAL if ci_build
+      end
```
Iristyle: I think we could have condensed lines 107 - 112 into:

```ruby
@log_colors.merge!({:notify => NORMAL, :info => NORMAL}) if ENV['BUILD_NUMBER']
```

ferventcoder: Probably, but you mean lines 109-112. The comments still apply and should be left in there.

kevpl:
yeah, there are a number of things that Jenkins passes to Beaker, any of them could have done the job here. I'm mainly concerned with if we've ever done that before, or if we go to do it again, we make sure to abstract it out into a standard helper, so that we don't do a bunch of varied checks to get the same info.

ferventcoder:
@kevpl agreed. See https://github.com/auchenberg/volkswagen for fun and the https://www.npmjs.com/package/is-ci library. This may be what @er0ck was mentioning earlier in a comment.

It's super simple in that it just uses https://github.com/watson/is-ci/blob/cc89aa9c9b1ac7946c69486f014100a3b484c922/index.js#L3-L16

~~~js
 // Generic environment variables
  process.env.CI ||                      // Travis CI, CircleCI, ...
  process.env.CONTINUOUS_INTEGRATION ||  // Travis CI, ...
  process.env.BUILD_NUMBER ||            // Jenkins, TeamCity, ...

  // Sever specific environment variables
  process.env.JENKINS_URL ||             // Jenkins
  process.env.bamboo_planKey ||          // Bamboo (by Atlassian)
  process.env.TF_BUILD ||                // Team Foundation Server (by Microsoft)
  process.env.TEAMCITY_VERSION ||        // TeamCity (by JetBrains)
  process.env.BUILDKITE ||               // Buildkite
  process.env.HUDSON_URL ||              // Hudson
  false
~~~

ferventcoder:
That way any CI system at play would switch the logging.

er0ck:
yeah, that's what i was referencing.  bunch of ENV.  i'm down with checking them all.  
where to put it... localhost/testrunner doesn't have a host object :sob:

PR #986, opened 2015-10-14
(BKR-587) added nil as a result to apply_manifest_on docs

kevpl:


kevpl:
failure unrelated to change

justinstoller:
:+1:

PR #989, opened 2015-10-15
(BKR-584) Change to use mkdir_p helper command

petems:
Makes command cross-compatible for PSWindows


kevpl:

PR #991, opened 2015-10-19
(BKR-590) parameterized ports in helpers

kevpl:


lib/beaker/dsl/helpers/puppet_helpers.rb
```diff
@@ -300,8 +300,18 @@ def puppet_conf_for host, conf_opts
...
+        # @param [String] service Name of the service to restart
+        # @param [Fixnum] curl_retries Number of times to retry the restart command
+        # @param [Fixnum] port Port to check status at
+        #
+        # @return [Result] Result of last status check
         # @!visibility private
-        def bounce_service host, service, curl_retries = 120
+        def bounce_service host, service, curl_retries = nil, port = nil
+          curl_retries = 120 if curl_retries.nil?
+          port = 8140 if port.nil?
```
er0ck: perhaps these magic numbers could go into a Defaults module somewhere?

lib/beaker/dsl/helpers/puppet_helpers.rb
```diff
@@ -309,7 +319,7 @@ def bounce_service host, service, curl_retries = 120
             host.exec puppet_resource('service', service, 'ensure=stopped')
             host.exec puppet_resource('service', service, 'ensure=running')
           end
-          curl_with_retries(" #{service} ", host, "https://localhost:8140", [35, 60], curl_retries)
+          curl_with_retries(" #{service} ", host, "https://localhost:#{port}", [35, 60], curl_retries)
```
er0ck: magic numbers again

er0ck:
i'd like to see some acceptance tests added for these, if possible.

er0ck:
nice work, this definitely will help my immediate problem. 
i guess i need to ticket finding the puppetdb and server ports from answers and/or options in do_install in pe_utils.

kevpl:
failures unrelated to change

lib/beaker/dsl/helpers/puppet_helpers.rb
kevpl: I like this.  Working on it now

kevpl:
acceptance tests would be nice for this, but the way this issue was scoped, it's not really included.  Having acceptance tests for this would be more effort than this ticket itself, and should be taken on as a separate ticket.

kevpl:
killed that run, as another change was needed

kevpl:

kevpl:

kevpl:

kevpl:
failures unrelated to change

lib/beaker/dsl/helpers/puppet_helpers.rb
```diff
@@ -300,8 +300,18 @@ def puppet_conf_for host, conf_opts
...
+        # @param [Host] host Host the service runs on
+        # @param [String] service Name of the service to restart
+        # @param [Fixnum] curl_retries Number of times to retry the restart command
+        # @param [Fixnum] port Port to check status at
+        #
+        # @return [Result] Result of last status check
         # @!visibility private
-        def bounce_service host, service, curl_retries = 120
+        def bounce_service host, service, curl_retries = nil, port = nil
+          curl_retries = 120 if curl_retries.nil?
```
rick: Hm.  I think just leaving this as `curl_retries = nil` will do the defaulting, unless the desired use-case is:

``` ruby
bounce_service node, 'fooservice', nil, 1234
```

i.e., specifying `nil` in order to get the default value of `curl_retries` when specifying `port`.

I'm not super :smile_cat: on that approach -- it's how hash arguments were invented, basically.  I'm willing to sign off on it here, as this is the interface we 

For the wider (not this PR) conversation, I wonder if we should be thinking about kwargs (keyword args) for our various parameters. The downside there would be interface change (not the end of the world, especially if it improves the interface). Depends on ruby version, and I'm not 100% clear on what we expect for ruby version.

lib/beaker/dsl/helpers/puppet_helpers.rb
```diff
@@ -300,8 +300,18 @@ def puppet_conf_for host, conf_opts
...
+          port = options[:puppetserver_port] if port.nil?
```
rick: I think this one we can default in the signature, fwiw.

spec/beaker/dsl/helpers/puppet_helpers_spec.rb
```diff
@@ -894,5 +894,92 @@ def stub_post_setup
...
+      expect( subject ).to receive( :curl_with_retries ).with( anything(), anything(), /4433/, anything() ).once.ordered
+      subject.sleep_until_nc_started( host )
+    end
+
+    it 'allows setting the port' do
+      host = hosts[0]
+      expect( subject ).to receive( :curl_with_retries ).with( anything(), anything(), /4435/, anything() ).once.ordered
+      subject.sleep_until_nc_started( host, 4435 )
+    end
+  end
```
rick: I'm generally pretty :-1: on mockist tests, these included. I agree we should be looking for acceptance tests here, though, to be fair, I'm roadblocking lots of things by not having some version of https://github.com/puppetlabs/beaker/pull/930 landed to make that easier for everyone.

rick:
I'm :+1: on this with two caveats: going forward we should be doing more acceptance testing, and we should look at keyword args for helper arguments as we look at extractions, surface area, etc. 

For the record, I don't like the idea of splitting out testing from the development work, sprint/point-wise, as it promotes test-later (as opposed to test-first or even TDD testing), or test-never development. No one who wants a feature ever puts testing ahead of features in the backlog, so splitting them out becomes the tragedy of the commons. :house: :fire:

kevpl:
I agree, I don't think we should split out tests either. I don't think they ever happen in that case (your test-never case).  It's the idea that they are scoped together which made me say that there wasn't enough in this ticket to include them.

I'm down with your two caveats. Comments on your inline comments coming...

lib/beaker/dsl/helpers/puppet_helpers.rb
kevpl: yes, that is the desired case, so that we can set `port` without needing to know/update `curl_retries` later.

I'm down to use better patterns when we can, so I'd like to have this conversation whenever you can have it.

spec/beaker/dsl/helpers/puppet_helpers_spec.rb
kevpl: cool (that we're pointing out what you've been talking about as mockist tests now).  Do you see a spec test being able to test this in a way that is not mockist, or is your point that spec testing here is mockist in general?

I originally thought that acceptance testing this would be kind of a huge burden for so small a change, but I guess since we're usually pinning versions of customer code in our acceptance tests that it might not be as error prone as I was thinking.

kevpl:
@rick ready for you to take another look!

spec/beaker/dsl/helpers/puppet_helpers_spec.rb
rick: There are definitely ways to write non-mockist spec tests; though the ability to do it varies with the specific code under test.  Sometimes working with existing code (as opposed to writing new code which could conceivably be done test-first or TDD with the tests influencing the design) can be difficult to test without resorting to stubs/mocks heavily. There's at least [one really good book](http://www.amazon.com/Working-Effectively-Legacy-Michael-Feathers/dp/0131177052) that I recommend regularly on the topic.

Sometimes difficulty of spec (or unit) testing code without mocks/stubs is a sign that refactoring to service classes, etc., might be helpful to find seams in the design where tests would work.

If I were to do work here to make the design more testable I would probably end up introducing at least one additional class -- either to act as a carrier for the various parameters, or a class to act as the final delegate to execute the curl functionality (independent of how it was called, etc.). 

Ironically `Command` classes tend to come into play for making code like our helpers more testable. I'm not sure that's the history behind the `Command` classes in Beaker, though.

rick:
OK, gotcha -- can't tell if anything code-wise changed, but I think no.  :+1:

doug-rosser:
:+1:

spec/beaker/dsl/helpers/puppet_helpers_spec.rb
er0ck: when i started to think about the issues related to heavy required mocking in the beaker-facter tests, for instance, i was thinking about how we could add adapter classes to get around the issues.  this also allows us to retain a current interface while starting to implement the new one with better encapsulation.  It's also a lot of work :100: and we can rapidly get inconsistencies if a bunch of folks are all rapidly providing their own adapters.  This would probably manifest in data carriers as well, and i generally don't like in/out data-pipeline classes, as passing around data between too many objects gets smelly, fast.
and... i'm talking myself in a circle.  :confused:

rick: Dealing with existing code is always more work (but I've found Big Rewrites to be worse, so I just accept it as the cost of doing business).  It's easy to talk oneself into being paralyzed while theorizing. Implementing an adapter, test subclass, one-off interface, etc., while solving a specific problem tends to be much more fruitful; the only caveat being a commitment to see things through once they're started. :construction:

PR #993, opened 2015-10-22
(BKR-596) Fix Windows Path Support

cowofevil:
The "copy_module_to" method does not detect if the target machine is Windows
after SCP copy of the module. It attempts to move the directory because path
joining does not account for "/" separators in the path.

### DIFF
--- a/lib/beaker/dsl/install_utils/module_utils.rb
+++ b/lib/beaker/dsl/install_utils/module_utils.rb
@@ -133,8 +133,12 @@ module Beaker
               #move to the host
               logger.debug "Using scp to transfer #{source_path} to #{target_path}"
               scp_to host, source_path, target_module_dir, {:ignore => ignore_list}
+
               #rename to the selected module name, if not correct
               cur_path = File.join(target_module_dir, source_name)
+              if host.is_powershell? #make sure our slashes are correct
+                cur_path = cur_path.gsub(/\//,'\\')
+              end
               host.mv cur_path, target_path unless cur_path == target_path
             when 'rsync'
               logger.debug "Using rsync to transfer #{source_path} to #{target_path}"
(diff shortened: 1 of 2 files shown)


cowofevil:
@kevpl The last failure is unrelated to the change.

kevpl:
couldn't see the failures, re-kicking to get visibility

kevpl:

kevpl:
failures unrelated to change

lib/beaker/dsl/install_utils/module_utils.rb
```diff
@@ -133,8 +133,12 @@ def copy_module_to(one_or_more_hosts, opts = {})
               #move to the host
               logger.debug "Using scp to transfer #{source_path} to #{target_path}"
               scp_to host, source_path, target_module_dir, {:ignore => ignore_list}
+
               #rename to the selected module name, if not correct
               cur_path = File.join(target_module_dir, source_name)
+              if host.is_powershell? #make sure our slashes are correct
+                cur_path = cur_path.gsub(/\//,'\\')
```
Iristyle: I'm concerned that we're conflating multiple things here.  There are 2 things that Beaker should really care about:

* Network transport - i.e. `ssh` vs `winrm`
* Interpreter - i.e. `bash`, `cmd`, `powershell`, etc

The problem is that @cowofevil is running Bitvise SSH, and he assumed we should be setting `is_cygwin: false`.

But that doesn't really convey what we care about, and since Beaker doesn't appear to track network transport  apart from interpreter, we end up in a strange state. `is_cygwin: false` appears to end up setting `is_powershell?` to true, which IMHO is not accurate.

Bitvise templates use `ssh` (like Cygwin), but the interpreter used is `cmd`, not `powershell`.

PR #995, opened 2015-10-23
(BKR-588) updated eos install support

kevpl:


lib/beaker/host.rb
```diff
@@ -499,6 +501,7 @@ def do_rsync_to from_path, to_path, opts = {}
     'freebsd',
     'windows',
     'pswindows',
+    'eos',
```
kevpl: @rick, is there a reason in Ruby why this should be at the bottom of the file?  Like how javascript makes the page load faster by being at the bottom kinda thing?

This was a little painful to find for me, so I'm wondering if there's any reason that I can't move this to the top like every other require ever.

rick: I mean... "kind of".  So, the thing there is that those files are declaring subclasses of `Beaker::Host` so ideally that base class is declared before they are encountered.

Various ways to do that:

 - Do what they did -- declare the full base class, all the code, etc., and load those classes after it, from the same file.
 - Declare the base class, and don't require the subclasses in the same file, relying on them to be included later, autoloaded, etc.
 - Declare the base class. Inside the base class, when something is actually instantiated (via a call to `.new` aka `initialize`; or on module inclusion if it's a module) go and require the relevant classes. This is often "require all the classes in this whole directory".
 - Declare the base class, require those files inside the declaration.
 - Declare the base class, and once that's declared, close it, require those files, and, then re-open the class.

So, there's a bunch of ways to do it (probably 3 or 4 more I didn't think of), and I agree, dropping those at the end of the file is less than ideal.

rick: Oh yeah, another less-than-ideal way: not declaring the base class and just loading the subclasses at the top of the file, and just ensuring the declaration (all nested modules, etc.) is done exactly the same way everywhere.  If you get it wrong, you'll get an error, so it's not really risky.

kevpl:

kevpl:
killed this test.  There are questions in the ticket that need to be answered before this can be looked at further.

kevpl:
@LuvCurves I've done the testing with a FOSS puppet master that we spoke of, and this is working fine.  I'm gonna take a little bit to update docs with the results of this work, but if you want to start doing pre-suite work testing with this branch, that should work.

I'd appreciate being able to know if any of our assumptions are wrong before this goes in if possible. Let me know if you have any issues or questions about doing this.

kevpl:

kevpl:

LuvCurves:
@kevpl I'm working on testing this in my environment now.  I'll try to have some feedback for you by the end of the day.

kevpl:
:+1:

kevpl:
failures unrelated to change

LuvCurves:
This is working for me.  Going to update JIRA now

lib/beaker/host.rb
kevpl: ok.  I think the current way is one of the more deterministic routes, so I'm down with just keeping it for now.

kevpl:
@rick, if you have some time this morning, I'd appreciate a look here! :wave:

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -1107,6 +1107,10 @@ def install_puppet_agent_dev_repo_on( hosts, opts )
             onhost_copy_base = opts[:copy_dir_external]
 
             case variant
+            when /^eos/
```
rick: Not for this PR, but @kevpl it wouldn't be the worst thing in the world for a set of platform / variant predicate methods to exist which wrap up our regexes and centralize them.  Something like `variant_is_eos?` (e.g.).  I may need a refresher course on when we should look to `variant` vs. `host['platform']` and what the differences are.

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -1193,11 +1197,17 @@ def install_puppet_agent_dev_repo_on( hosts, opts )
               raise "No repository installation step for #{variant} yet..."
             end
 
-            onhost_copied_file = File.join(onhost_copy_base, release_file)
-            fetch_http_file( release_path, release_file, copy_dir_local)
-            scp_to host, File.join(copy_dir_local, release_file), onhost_copy_base
+            if host['platform'] =~ /eos/
```
rick: Again, not a criticism of this PR, which is limited in scope, but I'm not excited about how liberally some of our code (`foss_utils` being an example) devolves into strings of `if` blocks. This makes me think again that we should be looking at a different breakdown of domain concepts that might enable us to use inheritance or delegation instead of conditionals.

lib/beaker/host/eos.rb
```diff
@@ -0,0 +1,56 @@
...
+    #
+    # @return [String, String] Path to the directory and filename of the package, respectively
+    def get_puppet_agent_package_info( puppet_collection = nil, puppet_agent_version = nil, opts = {} )
+      error_message = "Must provide %s argument to get puppet agent package information"
+      raise ArgumentError, error_message % "puppet_collection" unless puppet_collection
+      raise ArgumentError, error_message % "puppet_agent_version" unless puppet_agent_version
+
+      variant, version, arch, _ = self['platform'].to_array
+      release_path = "#{variant}/#{version}/#{puppet_collection}/#{arch}"
+      release_file = "puppet-agent-#{puppet_agent_version}-1.#{variant}#{version}.#{arch}.swix"
```
rick: I know nothing about this, but the "1." in here makes me want to ask -- when does this become a "2."?

rick:
:+1:

lib/beaker/dsl/install_utils/foss_utils.rb
kevpl: yeah, I agree, helpers here would be way better.  I'm down with fixing these up as a policy going forward.

lib/beaker/dsl/install_utils/foss_utils.rb
kevpl: I agree.  I thought about making a `get_remote_file` file in the `unix/file.rb` code so that that could just be `host.get_remote_file` rather than the if statement, but I backed down from that :-/

lib/beaker/host/eos.rb
kevpl: To be honest, I don't know, I just know that this works.  You'd have to ask RE for that info.  My first suspicion would have been puppet collection, but we already take that into account right above it.

doug-rosser:
:+1:

PR #996, opened 2015-10-26
(BKR-600) Import wiki docs, relativize markdown links

rick:
This is a spike to evaluate the workability of moving the Beaker wiki documentation to in-repo documentation.

/cc @puppetlabs/beaker @er0ck


rick:
[Preview link](https://github.com/puppetlabs/beaker/blob/bkr-xxx/move-wiki-to-in-repo-docs/Home.md)

kevpl:
it looks like this clutters up the base beaker dir.  Is there a reason you wanted to do that rather than having something like a `docs` directory?

rick:
I did want it in a docs dir; my bad, not sure what happened on that last attempt (I've done this a few times, getting the rebase/squash stuff right).

rick:
Give it a spin now.

kevpl:
It looks like this is a straight pull of the files, without any history, is that right?  I just want to make sure there's not any weird git history masking going on.  The way wiki editing works, I think it's better this way (as opposed to pulling over inconsequential/not helpful commit history), but I just want to be explicit about this.

I think at some point, we should take a look at putting in something that can render these locally in a browser, so that you can actually read the documentation as if on github (rather than having to read the markdown), even if you're on an airplane, but I think that's fine as a future improvement.

As probably a commit on top of this, but (hopeful) included in this PR, we should update the main project README to link to the docs home page to give people some inroads to the docs.

Other than bringing up this stuff, I think there are further conversations to have about things such as organizing the docs, but just getting these in as they are is the right first step! :+1:

rick:
This is explicitly a pull without history (see the commit message, but "--sqaush" to git subtree eliminates history).  I additionally relativized the intra-document links, and then also ran through a pass of trailing-whitespace / "DOS newline" cleanups.

I'm totally fine with linking the `README.md` into this documentation in this PR, I'll take a look at it.

rick:
Updated README.me ([rendered version](https://github.com/puppetlabs/beaker/blob/bkr-xxx/move-wiki-to-in-repo-docs/README.md) to point to this version of the documentation.

kevpl:
ah cool, then I'm :+1: on this.  I'm down with getting more opinions if you want, but if you're ready, then feel free to create a ticket for it, and I'll be ready to merge it.

rick:
Created [BKR-600](https://tickets.puppetlabs.com/browse/BKR-600) for this.

kevpl:
manually stopped the test

PR #997, opened 2015-10-26
(MAINT) Move docs to README.md, rename some files

rick:
Prior to this we were using `Home.md` as the root for documentation in docs/ (as a direct import from the wiki). For most purposes, `README.md` is going to be easier (as users get it rendered by default and don't have to know that `Home.md` is a thing).

Also normalize some file names that have punctuation in them, as these are just a headache on the cmdline, for linking on the web, etc.  

A few files also picked up strange names (completely my fault) at some point late in the automated cleanup process during import, so this also fixes those. Went through and audited links again to try to catch any breakage.

[Rendered view of README.md](https://github.com/puppetlabs/beaker/blob/6d45acb92caa335402cfe4d2f4680df470037b64/docs/README.md)

/cc @puppetlabs/beaker @er0ck


electrical:
Had a quick look around in the docs and looks good.

rick:
:sparkles: :guitar: :metal:

PR #998, opened 2015-10-27
(BKR-598) made AWS EC2 key name collision impractical

kevpl:
This property allows the user to make the AWS EC2 key name unique in their
Beaker runs. This property can be set from the CONFIG section of the hosts
file.

### DIFF
--- a/docs/hypervisors/aws.md
+++ b/docs/hypervisors/aws.md
@@ -1,0 +1,143 @@
+# Amazon Web Services - Elastic Compute Cloud (EC2)
+
+EC2 is a "web service that provides resizable compute capacity in the cloud."
+
+[EC2 site](https://aws.amazon.com/ec2/).
+
+# Getting Started
+
+### Requirements
+
+- Get EC2 access from your IT dept, particularly your `aws_access_key_id` & `aws_secret_access_key`.
+- put these values into your [~/.fog file](http://fog.io/about/getting_started.html).
+
+### Setup Amazon Image Config
+
+The Amazon Image Config file in Beaker is the file that specifies which Amazon
+Machine Image (AMI) should be used for a host and which EC2 region that host
+should be generated into.
+
+The text in this file follows this form:
+
+    AMI:
+      <host-vmname-value>:
+        :image:
+          :<type>: <ami-id>
+          :<type>: <ami-id>
+        :region: <region-id>
+      <host-vmname-value>:
+        ...
+
+The `host-vmname-value` is an ID used to reference one of these particular AMI
+definitions. It is applied to a host via the `vmname` key in the hosts file.
+
+The `type` variable is an arbitrary key that you can use to specify the different
+types of that host platform that you can be testing with. Note that this value
+will be grabbed automatically via the value for the host's `snapshot` key.
+For example, we tend to use `:pe` and `:foss` for these values.
+
+The `ami-id` variable is the AMI ID as specified by Amazon. You can see the AMI
+ID pattern in EC2's
+[Find a Linux AMI]
+(http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/finding-an-ami.html)
+page, particularly in the "using the Images page" section's step 7. For some
+examples of AMI IDs, check out their
+[Amazon Linux AMI page](https://aws.amazon.com/amazon-linux-ami/).
+
+The `region-id` variable represents the EC2 region ID from AWS. For reference,
+checkout EC2's 
+[Regions and Availability Zones page]
+(http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html).
+An example of a region ID is `eu-west-1` for the Ireland data center.
+
+This file is by default located at `config/image_templates/ec2.yaml`. This is a
+relative path from Beaker's execution location, and can be overridden using the
+`:ec2_yaml` key in a CONFIG section of a host file if required.
+
+### Create a Hosts File to Use
+
+An EC2 hosts file looks like a typical hosts file, except that there are a
+number of required properties that need to be added to every host in order for
+the AWS hypervisor to provision hosts properly.  They come in this form:
+
+    <hostname>:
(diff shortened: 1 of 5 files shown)


lib/beaker/hypervisor/aws_sdk.rb
```diff
@@ -687,7 +687,7 @@ def key_name_prefix
     # @return [String] safe key name for current host
     # @api private
     def key_name
-      "#{key_name_prefix}-#{@options[:timestamp].strftime("%F_%H_%M_%S")}"
+      "#{key_name_prefix}-#{@options[:aws_keyname_modifier]}-#{@options[:timestamp].strftime("%F_%H_%M_%S")}"
```
kbarber: Why not just add a random string instead? That way we (and anyone else using aws, who knows) don't have to change the job configuration. Since this bug is a regression from original behaviour, its a less user-impacting way of solving it.

lib/beaker/options/presets.rb
```diff
@@ -181,6 +181,7 @@ def presets
           :ec2_yaml               => 'config/image_templates/ec2.yaml',
           :help                   => false,
           :collect_perf_data      => 'none',
+          :aws_keyname_modifier   => '',
```
kbarber: Or default this to a random string, but once you have a random input, what requirements would bring about a non-random or predictable one?

Perhaps there is value in putting the job_name anyway, but that feels like a tangent to the regression.

electrical: Or we could also look at things that CI systems set as variables that should be unique for that job? and if its not in CI, fallback to a random number?

docs/hypervisors/aws.md
```diff
@@ -0,0 +1,32 @@
...
+
+The way to deal with this is to use the _aws_keyname_modifier_ variable.
+
+You can set the _aws_keyname_modifier_ variable in the host file in the options
+section:
+
+    CONFIG:
+      aws_keyname_modifier => unique_value,
+
+In CI systems, setting this to something like job name would make it unique enough
```
kbarber: Is JOB_NAME, what is recommended here? I'm a bit wary, so I went and tested what that string results to in our tests. Here is a JOB_NAME from one of our master runs:

```
+ echo 'Testing JOB_NAME'
Testing JOB_NAME
+ echo enterprise_puppetdb_integration-system-puppetdb_full-master/INSTALL_TYPE=install,LAYOUT=ec2-west-el7-64mda-el7-64a,PUPPETDB_DATABASE=embedded_db,label=beaker
enterprise_puppetdb_integration-system-puppetdb_full-master/INSTALL_TYPE=install,LAYOUT=ec2-west-el7-64mda-el7-64a,PUPPETDB_DATABASE=embedded_db,label=beaker
```

I'm not sure from memory what the rules are but we need to make sure this is going to be a valid resultant string that AWS would accept as a keypair name. I notice already we are being cautious around hostnames, and replacing the dots with hyphens from the hostname in #key_name_prefix. Those strings can sometimes contain weird stuff also.

Something to notice here, is that there is no other uniqueness element in JOB_NAME, the job number doesn't get included for example. If two jobs are running in parallel, they can still collide if JOB_NAME is the only thing used, and the tests start in the same second.

lib/beaker/options/presets.rb
kbarber: Why include anything external at all? Our goal isn't to add metadata, or is it? Milliseconds + random string creates a massive reduction in collision chance, probably enough for our needs.

kevpl:
thanks for the look guys!  I'll get on changing these, & have those up for you soon.

kbarber:
+1 to the new changes.

electrical:
Would it make sense to have tests for multiple hosts to ensure the random number is generated for each host? or am i just playing devils advocate now? :-)

kevpl:
The random number is generated per Beaker run, and has never been done on a per host level.

Before, we generated one key per host, and used it between runs, which could lead to corruption between them.

electrical:
Ahhh okay, i see.

rick:
:+1:

kevpl:
failure unrelated to change

docs/hypervisors/aws.md
```diff
@@ -0,0 +1,143 @@
...
+
+### How Do I Find Out My Key Prefix?
+
+In order to find out your key pattern as used by Beaker, just kick off a Beaker
+run to generate an EC2 host. When you do this, you should see lines that look
+like so:
+
+    aws-sdk: Launch instance
+    aws-sdk: Ensure key pair exists, create if not
+    [AWS EC2 200 0.142666 0 retries]^[[0m^[[1m describe_key_pairs(:filters=>[{:name=>"key-name",:values=>["Beaker-johnsmith-Johns-Ubuntu-2-local"]}]) ^[[0
```
justinstoller: It's so amazing to see hypervisor docs landing with any patch!

One little nit is the `^[[0m^[[1m`console escape sequences`^[[0` in here.

waynr:
:see_no_evil: :+1: after @justinstoller's comment is addressed.

kevpl:
failure unrelated to change

spec/beaker/hypervisor/aws_sdk_spec.rb
```diff
@@ -672,60 +679,14 @@ module Beaker
     describe '#ensure_key_pair' do
       let( :region ) { double('region', :name => 'test_region_name') }
       subject(:ensure_key_pair) { aws.ensure_key_pair(region) }
+      let( :key_name ) { "Beaker-rspec-SUT" }
 
-      context 'when a beaker keypair already exists' do
```
waynr: It's not clear to me why these tests are being removed...?

kevpl: they were actually bad tests that don't give us any meaningful results.  We haven't used cached keys and the `:keypair_generate_new` option hasn't generated new keys since BKR-522.  This should have been pulled there, but I missed them then.

waynr:
Okay, :+1: !

PR #999, opened 2015-10-27
(MAINT) Update main README.md docs link

rick:
This was missed in the final change-over from Home.md during BKR-602. Updating so that
this points to the new correct place.

/cc @puppetlabs/beaker

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ See [Beaker Installation](https://github.com/puppetlabs/beaker/wiki/Beaker-Insta
 
 #Documentation
 
-Documentation for Beaker can be found in this repository in [the docs/ folder](docs/Home.md).
+Documentation for Beaker can be found in this repository in [the docs/ folder](docs/README.md).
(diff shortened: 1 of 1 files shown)


kevpl:
cancelled the job by hand.  This is a good example of something that could use a flag to not run testing.  Perhaps the `(maint)` prefix is the right flag for that?  I dunno, but thought I'd put this out there.

justinstoller:
Jobs can be set up to not trigger, or only trigger on certain paths (a number of teams use this to not trigger on docs updates).

There's also a plugin that I've been dying to try out but don't have the bandwidth:
https://wiki.jenkins-ci.org/display/JENKINS/Ci+Skip+Plugin
http://docs.openstack.org/infra/jenkins-job-builder/wrappers.html#wrappers.ci-skip
I'm a little hesitant based on it using the ruby runtime and not having a wider install base (but if you wanted to try it out....  :)  )

PR #1002, opened 2015-10-30
(BKR-483) Add Huawei 6 Platform support

LuvCurves:


LuvCurves:
Going to wait until we have more logic on how to do the install automatically from a repo

PR #1003, opened 2015-11-02
(BKR-511) Beaker fails to launch Vagrant VMs when run within Bundler

GeoffWilliams:
Unset the RUBYLIB variable before shelling out to run vagrant.  Fixes dependency on outdated version of bundler


electrical:
Lgtm

PR #1005, opened 2015-11-02
[QENG-3063] [DO NOT MERGE] Introduce absurd logging for debugging purposes

rick:
Attempting to diagnose a weird in-jenkins-only test failure where on debian7 and centos7 sometimes hosts will not reboot. Not only this, but no output is received once the reboot sequence is started; so jenkins times out after 10 minutes of lack of output.

Here attempting to both flush logging destinations on all output for this test, as well as introducing a large amount of additional output.  Let's see if we can get more information on this test failure.

/cc @puppetlabs/beaker

PR #1006, opened 2015-11-03
[QENG-3063] Fix reboot test timeout issues on jenkins

rick:
- [QENG-3063](https://tickets.puppetlabs.com/browse/QENG-3063)

Prior to this change, the `host_test.rb` check for rebooting systems could sometimes fail, only on our Beaker acceptance jenkins, but in such a way that it was clear that something was happening after the attempt to reboot started, but no further output was available.

Exploratory work showed that when a rebooted host changes IP address, it is possible for not only the normal long sequence of Fibonacci-fallback connect retries to happen, but for also various low-level TCP connection failures (with their own long timeouts) to occur.

Since output was being buffered, and not flushed, and neither of these phases was generating a lot of output (enough to trigger a buffer flush on its own), it was possible for more than our 10 minute jenkins output timeout to pass without effectively changing the output buffer.

Flushing our log output overcomes the jenkins timeout problem. We do not here address questions of long IP-address change timeouts.

/cc @puppetlabs/beaker 
/related: https://github.com/puppetlabs/beaker/pull/1005/

### DIFF
--- a/acceptance/tests/base/dsl/helpers/host_helpers/create_remote_file_test.rb
+++ b/acceptance/tests/base/dsl/helpers/host_helpers/create_remote_file_test.rb
@@ -81,10 +81,18 @@ test_name "dsl::helpers::host_helpers #create_remote_file" do
       remote_filename = File.join(remote_tmpdir, "testfile.txt")
       contents = fixture_contents("simple_text_file")
 
-      create_remote_file default, remote_filename, contents, { :protocol => "rsync" }
+      result = create_remote_file default, remote_filename, contents, { :protocol => "rsync" }
 
-      remote_contents = on(default, "cat #{remote_filename}").stdout
-      assert_equal contents, remote_contents
+      fails_intermittently("https://tickets.puppetlabs.com/browse/BKR-612",
+        "default" => default,
+        "remote_tmpdir" => remote_tmpdir,
+        "remote_filename" => remote_filename,
+        "contents" => contents,
+        "result" => result
+        ) do
+          remote_contents = on(default, "cat #{remote_filename}").stdout
+          assert_equal contents, remote_contents
+      end
     end
   end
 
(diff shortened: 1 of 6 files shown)


kevpl:
:+1:, will merge after Jenkins goes green

justinstoller:
Bwahahahahah!

He means NEVER!

rick:
lololol

rick:
So, as I mentioned in the hipchat channel... the way we're spec testing our logger was to invent a `MockIO` class which prevents `.flush` from working (causes an error due to the IO stream not being initialized). That class is a subclass of `IO`, which is great, but it neuters all the methods so that they can't work.

Why? Who knows.

Instead of being overly mockist (and making it difficult to change our code and have confidence), we could actually test the IO paths more fully.  E.g., http://product.reverb.com/2014/12/19/testing-logging-in-ruby-with-stringio/

So, turns out, `StringIO` makes the tests pass just as well as our `MockIO` does, but actually does things like round-tripping through an IO-compatible structure (not just a blank object with stubbed method names).

The downside is that we're not really testing things through `StringIO` any more than we were with `MockIO`, since all the tests just do `expect()` calls on the object.  But we *could*. I think we should, at our next opportunity.

Will push that spec change in a second.

Also, I probably want to go ahead and neuter the intermittently failing `create_remote_file via rsync` tests (as discussed in hipchat) for this go-around so that we can actually get things green in the cases where this makes a difference. There will be a separate Jira issue for that related problem.

rick:
Pushed changes to use `StringIO` instead of `MockIO` (we should consider making those tests non-mockist in the future), also neutered [BKR-612](https://tickets.puppetlabs.com/browse/BKR-612) test here, since it will almost certainly fail whenever this failing behavior would be exposed.

rick:
Added `fails_intermittently` wrappers to more fully isolate rsync-related specs that can fail on debian7 and centos7 due to stale IP issues. Will build until our green build shows a rebooted host which has a stale IP -- *that* green means that this patch is working for QENG-3063 without tripping up on rsync issues (which will be resolved via BKR-612).

rick:
retest this please

rick:
:metal: :guitar:

[This build](http://jenkins-beaker.delivery.puppetlabs.net/job/qe_beaker_intn-sys_beaker-acceptance-base-vpool/1097/agent=centos7/consoleFull) includes a centos7 host which changes IP on reboot, reboot tests do not fail; rsync tests would fail but are caught in the `fails_intermittently` blocks.

:ship:

PR #1008, opened 2015-11-11
(BKR-622) Fix Regular Expression to give correct Host IP Address

developerinlondon:
Without this patch applied the hostname can sometimes be incorrectly picked up and this
creates problem with all the hostnames getting the same IP addresses on a multi-node setup.

Example Nodeset that can cause this problem:
```
HOSTS:
  staging:
    roles:
      - sta
      - master
    platform: el-6-x86_64
    box: puppetlabs/centos-6.6-64-puppet
    box_url: https://vagrantcloud.com/puppetlabs/boxes/centos-6.6-64-puppet
    hypervisor: vagrant
    shared_folder: abc
    NetworkSettings:
      IPAddress: 10.255.50.100
  etl:
    roles:
      - etl
    platform: el-6-x86_64
    box: puppetlabs/centos-6.6-64-puppet
    box_url: https://vagrantcloud.com/puppetlabs/boxes/centos-6.6-64-puppet
    hypervisor: vagrant
    NetworkSettings:
      IPAddress: 10.255.50.120
CONFIG:
  type: git
  destroy: no
```

In the above scenario, the hostname 'etl' is also available when you search 'puppetlabs'.

The patch fixes this by changing the regular expression to look for the string 'etl' (including
the quote marks).

### DIFF
--- a/lib/beaker/hypervisor/vagrant.rb
+++ b/lib/beaker/hypervisor/vagrant.rb
@@ -106,7 +106,7 @@ module Beaker
       ip = ''
       if File.file?(@vagrant_file) #we should have a vagrant file available to us for reading
         f = File.read(@vagrant_file)
-        m = /#{hostname}.*?ip:\s*('|")\s*([^'"]+)('|")/m.match(f)
+        m = /'#{hostname}'.*?ip:\s*('|")\s*([^'"]+)('|")/m.match(f)
         if m
           ip = m[2]
           @logger.debug("Determined existing vagrant box #{hostname} ip to be: #{ip} ")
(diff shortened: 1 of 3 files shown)


lib/beaker/hypervisor/vagrant.rb
```diff
@@ -106,7 +106,7 @@ def get_ip_from_vagrant_file(hostname)
       ip = ''
       if File.file?(@vagrant_file) #we should have a vagrant file available to us for reading
         f = File.read(@vagrant_file)
-        m = /#{hostname}.*?ip:\s*('|")\s*([^'"]+)('|")/m.match(f)
+        m = /'#{hostname}'.*?ip:\s*('|")\s*([^'"]+)('|")/m.match(f)
```
rick: Seems like this might need to allow for `"`s to be used here?

developerinlondon: well currently beaker always adds the ' mark anyways, so the above regex will always match the correct hostname's IP.

kevpl:

kevpl:
looks like this was a transient failure.  Re-kicking now

kevpl:

developerinlondon:
are we good with this @kevpl ?

PR #1010, opened 2015-11-12
(BKR-624) Added a setter for line_prefix

dylanratcliffe:
Added a setter for `line_prefix` that calculates the length as it is set. Also modified the `step_in` and `step_out` methods to use this new setter


lib/beaker/logger.rb
```diff
@@ -89,6 +89,12 @@ def initialize(*args)
       dests.each {|dest| add_destination(dest)}
     end
 
+    # Setter for line_prefix that also calculates line_prefix_length
+    # @param [String] prefix The line prefix
+    def line_prefix=(prefix)
+      @line_prefix = prefix
+      @line_prefix_length = prefix.length
```
kevpl: if we're going to give up `@line_prefix_length`, we should do so entirely.  Why is it being kept/updated here?

kevpl:
@dylanratcliffe please add a spec test, so that we can verify that we don't break this easily in the future.

lib/beaker/logger.rb
dylanratcliffe: @kevpl I was doing it for backwards compatibility bit it seems nothing is using it so i'll kill it

dylanratcliffe:
@kevpl Okay all done

PR #1012, opened 2015-11-13
(BKR-609) added checks to make aio_version? more robust

kevpl:


kevpl:
@puppetlabs/beaker ready for a look!

spec/beaker/dsl/roles_spec.rb
```diff
@@ -163,6 +163,14 @@ class ClassMixedWithDSLRoles
...
       expect( subject.aio_version?(agent1) ).to be === true
     end
+    it 'can take an empty string for pe_ver' do
+      agent1[:pe_ver] = ''
+      expect{ subject.aio_version?(agent1) }.not_to raise_error
+    end
+    it 'can take an empty string for FOSS version' do
+      agent1[:version] = ''
+      expect{ subject.aio_version?(agent1) }.not_to raise_error
+    end
```
rick: It seems like a useful set of tests here is to assert what the value of `aio_version?` would be when given a `nil` or `''` value for one or both of `:pe_ver`, and `:version`. I'm guessing that ordering (i.e., we check one before the other) is significant here.  What happens if both `:pe_ver` and `:version` are specified but one is `<` the other (and vice versa)?

rick: In looking deeper I see *some* tests for some of the cases in the prior tests. I think this could be made more rigorous -- i.e., step through all the possibilities, in truth-table fashion:

| pe_ver | version | aio_version?|
|-----------|----------|------------------|
| nil | nil | false |
| '' | nil | false |
| nil | '' | false |
| '' | '' | false |
| 3.8 | nil | false |
| 3.8 | '' | false | 
| etc. | etc. | etc. |
|-----|----|----|

There's probably a way to built a simple list of lists and just run through the assertions that way.  It would provide a lot better confidence in the behavior, and I wouldn't be surprised if it didn't turn up a couple of bugs.  Adding in the role and type strings to the matrix would probably be useful.

If this seems like drudgery, I can pair with you on it. :grinning:

kevpl: @rick added truth table-style tests.  Let me know what you think when you get some time.

rick:
:+1:

waynr:
Looks like spec tests are failing on this PR and the failures look legit.

rick:
Looks like legit [spec failures](http://jenkins-beaker.delivery.puppetlabs.net/job/qe_beaker_unit-rb_rspec/2416/ruby=1.9.3-p484/console).  Presumably the specs which mock out this method to test other things need to be updated (cue rants about mockist specs, etc.).

kevpl:
retest this please

waynr:
:+1:

PR #1013, opened 2015-11-16
[BKR-623] Reorganize Beaker test runner classes for introduction of minitest runner

rick:
See [BKR-623](https://tickets.puppetlabs.com/browse/BKR-623) for notes on direction. This will likely require a number of refactoring commits to get everything into place.

**TODO**

 - [x] Create `beaker/runner/beaker` + `Beaker::Runner::Beaker` and move  `Beaker::TestSuite` down, leaving a shim class behind
 - [x] Add support for a `--runner` cmdline flag, which defaults to `beaker`
 - [x] use `options[:runner]` to instantiate the appropriate runner from `cli.rb`
 - [x] move `Beaker::TestCase` down into `Beaker::Runner::Beaker`
   - [x] renamed `Beaker::Runner::Beaker` to `Beaker::Runner::Native`, because ruby
 - [x] migrate `TestSuite` tests into runner
 - [x] add a minitest runner `TestSuite` class
 - [x] migrate `TestCase` tests into runner
 - [x] unwind subclassing in `Beaker::TestSuite`
 - [x] remove `, minitest` from `--runner` cmdline docs (not ready for prime time)
 - [x] rename `--runner beaker` to `--runner native`
 - [ ] :eyes: + :ship:
 - [x] structure / outcomes refactoring - see #1023 for ongoing work
   - [ ] open a new Jira ticket for this
   - move "structure" / "outcomes" modules into includes for native `TestCase` class
   - move non-structure/non-outcomes modules into `Beaker::TestCase` class includes
   - clean up `Beaker::DSL`
   - unwind subclassing in `Beaker::TestCase`
   - alias `Beaker::DSL` and `Beaker::TestCase`
 - [ ] add a new Jira ticket for running minitest tests
   - add basic support for minitest runner to execute a series of tests (:warning: this will not work on a beaker suite unless we map structure/outcomes helpers to minitest syntaxes)
   - ensure that `Beaker::TestCase`/`Beaker::DSL` methods are available to minitest runner `TestCase`

/cc @puppetlabs/beaker


rick:
retest this please

rick:
Including the [mikado graph](https://pragprog.com/magazines/2010-06/the-mikado-method) here that I'm working from:

bl_ocks_org_cjrd_raw_6863459_

Also, the .json file which contains the graph information can be found [here](https://gist.github.com/rick/b42d0e6c32d8e1d23f01) To use this, go to [the editor](http://bl.ocks.org/cjrd/raw/6863459/) and upload via the little :cloud: icon.

rick:
For work related to "structure / outcomes" helpers, I'm referring to this code:

 - https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl.rb#L73-L82
``` ruby
  module DSL
    include Beaker::DSL::Roles
    include Beaker::DSL::Outcomes
    include Beaker::DSL::Structure
    include Beaker::DSL::Assertions
    include Beaker::DSL::Wrappers
    include Beaker::DSL::Helpers
    include Beaker::DSL::InstallUtils
    include Beaker::DSL::Patterns
  end
```

Of these modules, `Beaker::DSL::Structure` and `Beaker::DSL::Outcomes` are test-runner specific functionality. We should move these to a place where they can be included in the `Beaker::Runner::Native` hierarchy, and away from `Beaker::TestCase`, where they get included in all runners.

rick:
@ki pursuant to our discussions today, maybe we should review this PR and decide if it's ok to land as-is, and move the remainder of the work to next PR(s)?

kevpl:
I'm ok with moving to get this in, moving later work to later PRs.  Some comments to follow soon in-line

lib/beaker/options/command_line_parser.rb
```diff
@@ -98,6 +98,12 @@ def initialize
             @cmd_options[:timeout] = value
           end
 
+          opts.on '-r RUNNER', '--runner RUNNER',
+                  'Specify which test runner to use',
+                  'supported runners: beaker, minitest' do |value|
```
kevpl: The minitest runner itself doesn't look like it runs at this point, is that true?  I'm fine with that being the case, but I'd like to update this line that says it can until it's true.

lib/beaker/test_suite.rb
```diff
@@ -1,407 +1,17 @@
...
-      begin
-        run
-        return self if @test_suite_results.success?
-      rescue => e
-        #failed during run
-        report_and_raise(@logger, e, "TestSuite :run_and_raise_on_failure")
+    def self.runner(runner)
+      case runner
+      when "beaker"
+        ::Beaker::Runner::Native::TestSuite
```
kevpl: Why Native for the module wrapping TestCase?  Is this because other runners will have their own notion on what a TestCase is, so we're keeping these things in the Native namespace?

If that's the case, then I think this is an opportunity to clear up potential future confusion.  Having the runner be named Beaker means it'll easily be conflated with Beaker in general.  With the Native module, there's also a mapping that'll have to be maintained in people's understanding that the "beaker runner" is equivalent to the Native module in the code.  If we have this string (& the preset) named "native", then we eliminate both the confusion possibility, & stay consistent with our module path.

lib/beaker/options/command_line_parser.rb
rick: That's a good point :eyes: :sparkles:  -- it will not do anything useful as of the moment. We actually should probably drop the ", minitest" from here and only add it when we want to make minitest available for public use.

lib/beaker/test_suite.rb
rick: Ah, I get you.

For the moment, `TestCase` is basically a "native-only" construct, because it's moved from the default beaker runner into the native module space. I suspect minitest won't even have a TestCase, and wouldn't have to.  

But the main point, as I'm hearing it, is that users should call `beaker --runner native` instead of `beaker --runner beaker`, which makes sense. I originally tried to make this module namespace be `Runner::Beaker` but ran into headaches related to scoped constant resolution (i.e., when looking for `Beaker` as a constant name you could get `::Beaker` or `::Beaker::Runner::Beaker`, depending). I moved to `Native` but didn't re-think the `--runner` name.

:+1: to renaming the `--runner` arg.

kevpl:
@rick do you feel this is ready to be merged?  Would you mind either taking "[DO NOT MERGE]" off the PR title, or just commenting here to that effect, and I'll get this merged?

rick:
Updated checklists, etc., re-reviewed code. What could go wrong?  :ship: :shipit:

lib/beaker/options/command_line_parser.rb
```diff
@@ -98,6 +98,12 @@ def initialize
             @cmd_options[:timeout] = value
           end
 
+          opts.on '-r RUNNER', '--runner RUNNER',
+                  'Specify which test runner to use',
+                  'supported runners: beaker' do |value|
```
kevpl: @rick shouldn't this be 'native' instead of 'beaker'?

rick: You and your facts!  Yes, good :eyes:

kevpl:
other than my comment above (sorry for the :facepunch:), I'm :+1:

rick:
> Updated checklists, etc., re-reviewed code. What could go wrong?

:fire:

PR #1015, opened 2015-11-18
BKR-629: Adds compile masters to the array of roles 

erikPrime:
in the not_controller method, so that agents on compile_master systems will show as controllers

### DIFF
--- a/lib/beaker/dsl/roles.rb
+++ b/lib/beaker/dsl/roles.rb
@@ -100,7 +100,7 @@ module Beaker
       #     end
       #
       def not_controller(host)
-        controllers = ['dashboard', 'database', 'master', 'console']
+        controllers = ['dashboard', 'database', 'master', 'console', 'compile_master']
         matched_roles = host['roles'].select { |v| controllers.include?(v) }
         matched_roles.length == 0
       end


kevpl:
@erikPrime is there any kind of master that would be considered "not a controller"?  What is a master of master's role?  I'm wondering if perhaps we should have a separate check in that method to just make sure nothing that ends in `_master` is included.  What do you think?

rick:
So... while we're in here. I hate this method.  Is there a way we can define `def controller?`, with a positive statement about what it means to be a controller?  We could then make `def not_controller() !controller? end` and deprecate it.

kevpl:
last comments still current

kevpl:
last comments still current

kevpl:
It looks like this commit is tied to the wrong JIRA ticket ([BKR-629](https://tickets.puppetlabs.com/browse/BKR-629)). Please update the commit message & PR title to point to the correct one.

erikPrime:
@kevpl Sorry I lost contact on this PR.
Q1.  Is there any kind of master that would be considered "not a controller" ?
A: No.  The beaker role of the master of masters is "master".  You COULD argue that this ought to be handled by the "agent_only" logic, but that isn't where pry led me...
Q2: ...perhaps we should have a separate check in that method to just make sure nothing that ends in _master is included
A:  I don't think that would help, because "controller" means "master", "dashboard", "database", and now "compile_master"
@rick I completely agree.  Even worse, when I grep through the beaker repo the word "controller" is sometimes used as above [ master, dashboard, database, compile_master ] and sometimes used as the host on which beaker is executing, the 'test controller'.

So I'm going to look to see if we can refactor our file_sync tests to use the "agent_only" method instead of the "not_controller" method...

PR #1016, opened 2015-11-20
(BKR-472) (BKR-475) Fix solaris for install_pe_on with 2015.2

hunner:
When trying to install puppet-agent collection packages through
`install_pe_on`, eventually `install_puppet_agent_pe_promoted_repo_on`
gets called but has no entry for where the solaris packages are kept.
This commit adds the ability for solaris to install PC puppet-agent
packages.


kevpl:
retest this please

PR #1017, opened 2015-11-20
(BKR-634) Fix YAML deserialization for Beaker::Platform

adrienthebo:
The version of Psych shipped with Ruby 1.9.3-p194 did not support
subclasses of String that added additional ivars to the String class.
The Beaker::Platform class does just that - subclassing String and
additional ivars - which meant that round tripping a Beaker::Platform
object would create an object with improperly initialized instance
variables. Psych commit e2fcf9af9e95535401f816bc893839b9ad743a9e
resolved that issue but we still use platforms that have the old version
of psych.

To resolve this issue, this commit implements a custom #init_with method
that explicitly sets all instance variables on the Beaker object that
were defined inside of the YAML map instance, and then reconstructs the
string value based on those fields.

/cc @er0ck

### DIFF
--- a/lib/beaker/platform.rb
+++ b/lib/beaker/platform.rb
@@ -110,5 +110,13 @@ module Beaker
       [@variant, @version, @arch].join('-')
     end
 
+    if RUBY_VERSION =~ /^1\.9/
+      def init_with(coder)
+        coder.map.each do |ivar, value|
+          instance_variable_set("@#{ivar}", value)
+        end
+        replace("#{@variant}-#{@version}-#{@arch}")
+      end
+    end
   end
 end
--- a/spec/beaker/platform_spec.rb
+++ b/spec/beaker/platform_spec.rb
@@ -128,6 +128,22 @@ module Beaker
 
     end
 
-  end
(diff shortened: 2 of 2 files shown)


adrienthebo:
This should have some fairly exhaustive verification of the behavior across versions of Ruby, but I'm Pretty Confident (TM) that this should work correctly on all versions of Ruby that use Psych.

adrienthebo:
@er0ck could I get the magic jenkins incantation uttered to kick this build?

er0ck:
opened a ticket for this: BKR-634
@adrienthebo quick work on this one!
if you have a second can you associate this PR/commits with the ticket above?
if not, i can probably just push to your branch...

er0ck:
FYI, the magic incantation is 
'rete st
th1s
please' 

or some parts of it, all one one line, speled correctly :-D

adrienthebo:
@er0ck I just updated the commit message to reference BKR-634, do you need anything else for this?

er0ck:
looks good to me!

i guess the PR tests kickoff isn't smart enough to know when it's just a commit msg change...

PR #1020, opened 2015-11-23
(BKR-604) Windows BitVise install_pe fixes

kevpl:


kevpl:
lgtm failure. Looking into it.

kevpl:
retest this please

lib/beaker/host/pswindows.rb
```diff
@@ -23,5 +23,17 @@ def platform_defaults
...
+    attr_reader :network_separator, :external_copy_base, :system_temp_path
+    def initialize name, host_hash, options
+      super
+
+      @network_separator = '/'
+      @external_copy_base = '/programdata'
+      # %TEMP% == C:\Users\ADMINI~1\AppData\Local\Temp
+      # is a user temp path, not the system path.  Also, it doesn't work, there's
+      # probably an issue with the `ADMINI~1` section
+      @system_temp_path = 'C:\\Windows\\Temp'
```
Iristyle: Looks like we previously hardcoded this, but if we could avoid that, it would be better.

lib/beaker/host/pswindows/file.rb
```diff
@@ -2,17 +2,29 @@ module PSWindows::File
   include Beaker::CommandFactory
 
   def tmpfile(name)
-    execute("echo C:\\Windows\\Temp\\#{name}.%RANDOM%")
+    execute("echo #{system_temp_path}\\#{name}.%RANDOM%")
```
Iristyle: If we're going to run back to the host for this through `execute`, better to just use PowerShell IMHO.

```powershell
[IO.Path]::GetTempFileName()
```

lib/beaker/host/pswindows/file.rb
```diff
@@ -2,17 +2,29 @@ module PSWindows::File
...
   end
 
   def tmpdir(name)
     # generate name
-    tmpdirname = execute("echo C:\\Windows\\Temp\\#{name}.%RANDOM%")
-    # created named dir
+    tmpdirname = tmpfile(name)
+    # create named dir
```
Iristyle: Likewise, PowerShell has a way to do this already

```powershell
[System.IO.Path]::GetTempPath()
```

lib/beaker/host/pswindows/file.rb
```diff
@@ -2,17 +2,29 @@ module PSWindows::File
...
     execute("md #{tmpdirname}")
     tmpdirname
   end
 
+  # Turns a host file path into one that can be used by BitVise SSH
+  #
+  # @param [String] path Path to be changed
+  #
+  # @return [String] network version of that path
+  def network_path(path)
```
Iristyle: This seems like the wrong place to put this IMHO.  BitVise != PowerShell.

lib/beaker/host/windows.rb
```diff
@@ -24,5 +24,13 @@ def platform_defaults
       })
     end
 
+    attr_reader :network_separator, :external_copy_base
+    def initialize name, host_hash, options
+      super
+
+      @network_separator = '\\'
+      @external_copy_base = '`cygpath -smF 35`/'
```
Iristyle: This should work properly for real Cygwin hosts + Bitvise hosts with Git bash.

lib/beaker/dsl/install_utils/windows_utils.rb
```diff
@@ -5,14 +5,15 @@ module InstallUtils
...
         # Given a host, returns it's system TEMP path
+        #
         # @param [Host] host An object implementing {Beaker::Hosts}'s interface.
-        def get_temp_path(host)
-          # under CYGWIN %TEMP% may not be set
-          tmp_path = on(host, Command.new('ECHO %SYSTEMROOT%', [], { :cmdexe => true }))
-          tmp_path.output.gsub(/\n/, '') + '\\TEMP'
+        #
+        # @return [String] system temp path
+        def get_system_temp_path(host)
```
Iristyle: To avoid the new alias, wonder if it might be better to just change the signature of `get_temp_path` and pass in a new `kind` parameter with default value like:

```ruby
def get_temp_path(host, kind = :system)
```

lib/beaker/host/pswindows/file.rb
Iristyle: Also, I see where you're going with `network_path`, but it's a pretty generic term - I think `scp_path` might be more appropriate.

That said, this still seems like the wrong place to do this.  Might it be better to normalize paths inside the SCP helpers instead?  That way there aren't calls to `network_path` all over the code base every time such a request is initiated.  Let the slashification be deferred until copy time, keeping in mind that for Bitvise, `C:\` should become `/`?

/cc @joshcooper @ferventcoder

lib/beaker/host/pswindows.rb
```diff
@@ -23,5 +23,17 @@ def platform_defaults
       })
     end
 
+    attr_reader :network_separator, :external_copy_base, :system_temp_path
+    def initialize name, host_hash, options
+      super
+
+      @network_separator = '/'
+      @external_copy_base = '/programdata'
```
Iristyle: This can be retrieved with:

```
cmd.exe /c "for %I in (%ALLUSERSPROFILE%) do @echo %~sI"
```

lib/beaker/host/windows.rb
Iristyle: Also of note... I think that `@external_copy_base` should be resolvable at runtime like you're doing here.

However, if you look at how you've implemented in `host/pswindows.rb` and `host/unix.rb`, you'll see that `@external_copy_base` is set to a static value.  Can `opts[:copy_dir_external]` be resolved at runtime like this?

lib/beaker/host/pswindows/file.rb
kevpl: changed to `scp_path`.  I agree that doing this inside SCP helpers would be nice, but everywhere Beaker makes these calls, the platforms are dealt with before sending to the SCP calls so that the calls can focus on what they're supposed to do alone, so changing this now would be too big for this issue.

For your comment about bitvise not being powershell, what fact are we supposed to use to determine whether bitvise is in use from beaker?  It seems like template containing `-bv-` is the only distinguishing fact here.  I just want to make sure that's the right thing to do before potentially changing this up.

lib/beaker/dsl/install_utils/windows_utils.rb
kevpl: I actually like having the alias as a pointer to something that should be deprecated in the future.

Also, we already have a pattern in beaker where `host.tmpdir` should give you a user's temp dir.  Making this change just allows there to be two ways to get a temp directory for a user (even if `get_temp_path`'s would just be pointing to `host.tmpdir`), which is something I'd like to avoid.

lib/beaker/host/pswindows/file.rb
Iristyle: Ok - I understand not wanting to extend the scope of this issue... should we file another ticket then?

Since the `scp_to` accepts the host as a parameter, it should be able to canonicalize the path to be compatible with the current scp server in use, based on the host's details as discovered at runtime.  This would be a noop on most platforms, but take affect on Windows where different scp servers are being used.  Otherwise, any acceptance tests making calls to `scp_to` themselves would have to be written with up-front knowledge of which backend they're using (Cygwin vs Bitvise style pathing), which seems problematic,  unnecessary and wrong IMHO.  I feel this should be deferred until run-time.

Rather than use a convention as you've suggested for detecting whether BitVise is running, I would suggest introspecting the host to see if it is running Bitvise and set an internal `ssh_server` or similar value on the host (until a later date when host vs shell can be properly separated).  There are a couple of ways I can think to do this:

```
C:\Users\Administrator>cmd.exe /c sc query BvSshServer | find "RUNNING"
        STATE              : 4  RUNNING
```

OR

```
C:\Users\Administrator>cmd.exe /c netstat -a -b | find "BvSshServer"
 [BvSshServer.exe]
 [BvSshServer.exe]
 [BvSshServer.exe]
```

These mechanisms are not 100% perfect, but I think are probably good enough for now.  Does that make sense?

lib/beaker/host/pswindows.rb
kevpl: There's already enough change in this ticket, so I think this request is out of scope for this changeset.  If you'd like to see this change, please either submit a PR or a ticket for this.

lib/beaker/host/pswindows.rb
```diff
@@ -23,5 +23,17 @@ def platform_defaults
...
+    attr_reader :scp_separator, :external_copy_base, :system_temp_path
+    def initialize name, host_hash, options
+      super
+
+      @scp_separator = '/'
+      # %TEMP% == C:\Users\ADMINI~1\AppData\Local\Temp
+      # is a user temp path, not the system path.  Also, it doesn't work, there's
+      # probably an issue with the `ADMINI~1` section
+      @system_temp_path = 'C:\\Windows\\Temp'
+      @external_copy_base = '/programdata'
```
Iristyle: If you're going to use this for Cygwin:

```ruby
@external_copy_base = execute('echo `cygpath -smF 35`/')
```

You should use this here:

```ruby
@external_copy_base = execute('cmd.exe /c "for %I in (%ALLUSERSPROFILE%) do @echo %~sI"')
```

FWIW, `\ProgramData` is not valid on Windows 2003

lib/beaker/host/pswindows.rb
kevpl: I tried this, which returns `C:\PROGRA~3`, which doesn't work with scp.  I got it to present the directory to SCP just like having the value hard-coded does, and it would still fail, saying that the files don't exist:

      localhost $ scp ../my_beaker_files/tickets/bkr604/jenkins1.yml win-2008r2-x86_64-agent:/PROGRA~3 {:ignore => }
      RuntimeError: scp: /PROGRA\~3: Error opening file: No such file or directory.

I can login and verify that the files do exist after the run, and create the files before the SCP, and I still get this error message.

I'd rather keep what's there now and works, and create another ticket to improve this afterwards.

Iristyle: That doesn't look right... I just created a Bitvise host and tested manually with SCP (outside of Beaker), and what I suggested works fine:

```
[~] curl -d --url vmpooler.delivery.puppetlabs.net/vm/win-2012r2-wmf5-bv-x86_64                         13:18:53
{
  "ok": true,
  "win-2012r2-wmf5-bv-x86_64": {
    "hostname": "wu7b6kzeaaxxsmf",
    "ok": true
  },
  "domain": "delivery.puppetlabs.net"
}%                                                                                                                [~] scp ~/dotfiles.sublime-project "Administrator@wu7b6kzeaaxxsmf.delivery.puppetlabs.net:/PROGRA~3/dotfiles.sublime-project"
The authenticity of host 'wu7b6kzeaaxxsmf.delivery.puppetlabs.net (10.32.124.33)' can't be established.
ECDSA key fingerprint is SHA256:EKu2q+hGnVuwl36rb/2C0JlF2p3f90CMOTmwJkt4wSs.
Are you sure you want to continue connecting (yes/no)? yes
Warning: Permanently added 'wu7b6kzeaaxxsmf.delivery.puppetlabs.net,10.32.124.33' (ECDSA) to the list of known hosts.
dotfiles.sublime-project                                                        100%   59     0.1KB/s   00:00
```

While I was testing this, I also noted that 8.3 compatibility is not actually necessary for Bitvise SCP either:

```
[~] scp ~/dotfiles.sublime-project "Administrator@wu7b6kzeaaxxsmf.delivery.puppetlabs.net:/Program\ Files/dotfiles.sublime-project"
dotfiles.sublime-project                                                        100%   59     0.1KB/s   00:00
```

For reference, under Cygwin SCP, the following works (with slashes reversed direction):

```
[~] curl -d --url vmpooler.delivery.puppetlabs.net/vm/win-2012r2-wmf5-x86_64                            13:24:21
{
  "ok": true,
  "win-2012r2-wmf5-x86_64": {
    "hostname": "nd9gn3322qzrt78",
    "ok": true
  },
  "domain": "delivery.puppetlabs.net"
}%

[~] scp ~/dotfiles.sublime-project "Administrator@nd9gn3322qzrt78.delivery.puppetlabs.net:\\PROGRA~3\\dotfiles.sublime-project"
dotfiles.sublime-project                                                        100%   59     0.1KB/s   00:00


[~] scp ~/dotfiles.sublime-project "Administrator@nd9gn3322qzrt78.delivery.puppetlabs.net:\\Program\ Files\\dotfiles.sublime-project"
dotfiles.sublime-project                                                        100%   59     0.1KB/s   00:00
```

lib/beaker/host/pswindows.rb
kevpl: I've tried to use the `ALLUSERSPROFILE` command a number of times over this week, and it doesn't work in my testing.

The command returns `C:\PROGRA~3`, which is fine, but I can get the SCP call to match the working `/programdata` call, and it won't SCP correctly, giving me the message:

    localhost $ scp tmp/repo_configs/windows/puppet-agent-x64.msi win-2008r2-x86_64-agent:/PROGRA~3 {:ignore => }
    RuntimeError: scp: /PROGRA\~3: Error opening file: No such file or directory.

Regardless of whether the file(s) actually exist or not. I can verify that the directory does exist after the beaker run is complete (and before the command is called as well), and I've tried doing various things like creating the file before SCP-ing to it as a target.  I can even run this same command after a beaker run, and it'll work from my command line, but Beaker does not successfully make this call.

`/programdata` works in the cases that I've tested.  I'm confused about your point about Windows 2003.  When I run it by default, it uses the cygwin command and executes fine.  If I try to enable powershell by setting `is_cygwin: false` on the host, everything blows up, getting nowhere near running the command above.  I'm just using the host information below.  Is there something I'm missing to set this up correctly (like should I be pointing to a different template?)?

    windows2003r2-64-2:
      hypervisor: vmpooler
      platform: windows-2003r2-64
      template: win-2003r2-x86_64
      roles:
        - agent

kevpl:

cowofevil:
Just throwing in my 2 cents here. I think the "scp_to" method should implicitly know to strip the drive letter off of paths if the host is using BitVise. This is useful because we have many acceptance tests that pass fully qualified paths with drive letter using the "scp_to" method.

kevpl:
@cowofevil @Iristyle I took another look at putting the path correction into the `scp_to` method, and I think it was never there because it was just never needed, rather than it wasn't supposed to be there, so I've added it in as you've asked.

I've added in BitVise detection that you suggested, @Iristyle. I think the only issue left is the `@external_copy_base` on PSWindows.  Let me know if you'd like to talk more about any of the other topics, and hopefully we'll have this ready soon.

kevpl:
failures unrelated to change ([BKR-647](https://tickets.puppetlabs.com/browse/BKR-647))

lib/beaker/host/pswindows.rb
Iristyle: On Windows 2003, what I mean is that `C:\ProgramData` doesn't exist.  On that platform, `%ALLUSERSPROFILE%` resolves to `C:\Documents and Settings\All Users`

I think the issue in your example above might be with the escaping of `~`.  I don't believe that's necessary, and I think it's confusing SCP.

lib/beaker/host/unix.rb
```diff
@@ -23,5 +23,20 @@ def platform_defaults
       })
     end
 
+    def is_ssh_server_bitvise?
```
Iristyle: Why not just return a symbol for the type here?  Less code and more future proof IMHO.

```ruby
def ssh_server_type
  :openssl
end
```

lib/beaker/host/windows/file.rb
```diff
@@ -9,6 +9,32 @@ def tmpdir(name)
...
+
+  # Updates a file path for use with SCP, depending on the SSH Server
+  #
+  # @note This will fail with an SSH server that is not OpenSSL or BitVise.
+  #
+  # @param [String] path Path to be changed
+  #
+  # @return [String] Path updated for use by SCP
+  def scp_path(path)
+    if is_ssh_server_bitvise?
```
Iristyle: Same comment here about the API... why use the extra `is_xxx_` style methods, when the alternative is simpler / more resilient to change?

```ruby
case ssh_server_type
when :openssl
  path
when :bitvise
  network_path = path.gsub('\\', scp_separator)
  network_path.gsub('C:', '')
else
  raise ArgumentError
end
```

lib/beaker/host/pswindows.rb
Iristyle: @kevpl note in my examples above that I didn't escape the `~`

kevpl:
@Iristyle what do you think?

lib/beaker/host/unix.rb
```diff
@@ -23,5 +23,23 @@ def platform_defaults
...
 
+    # Determines which SSH Server is in use on this host
+    #
+    # @note This method is mostly a placeholder method, since only :openssl
+    #   can be returned at this time. Checkout {Windows::Host#determine_ssh_server}
+    #   for an example where work needs to be done to determine the answer
+    #
+    # @return [Symbol] Value for the SSH Server in use
+    def determine_ssh_server
+      :openssl
```
Iristyle: I think you had wanted to change this to `:openssh` @kevpl

lib/beaker/host/pswindows.rb
Iristyle: @kevpl you can also take the `s` out of the command I gave you to get the non-short version.

```
@external_copy_base = execute('cmd.exe /c "for %I in (%ALLUSERSPROFILE%) do @echo %~I"')
```

This will return `C:\ProgramData` instead of `C:\Progra~3`

kevpl:
manually killed. will re-kick if not picked up automatically

kevpl:

kevpl:

lib/beaker/host/unix/file.rb
```diff
@@ -9,6 +9,19 @@ def tmpdir(name)
     execute("mktemp -dt #{name}.XXXXXX")
   end
 
+  def system_temp_path
+    msg = "unix/file.rb: system_temp_path: Only available on Windows hosts"
```
Iristyle: When not return `/tmp` here?

kevpl:
failure unrelated to change

colinPL:
Waiting until the next beaker release before this is merged.

PR #1022, opened 2015-11-25
(BKR-636) Fix module howto

bleach:
I hit some issues when trying to use beaker for the first time. These changes would have prevented that.

Should fall under the trivial patch exemption from signing the CLA.


rick:
I'm :+1:, but definitely want another opinion from @puppetlabs/beaker.

PR #1028, opened 2015-12-04
(BKR-644) Prepend 'f' to the fedora version

joshcooper:
Previously, install_puppet_agent_dev_repo_on did not work for fedora,
because we were constructing a URL of the form:

    http://<host>/puppet-agent/<sha>/repos/fedora/22/PC1/x86_64/puppet-agent-1.3.2-1.fedora22.x86_64.rpm

However, we unfortunately prepend 'f' to the version string when
publishing the package, see RE-4191.

This commit prepends an 'f' to the version string for fedora-only, so
the new URL is:

    http://<host>/puppet-agent/<sha>/repos/fedora/f22/PC1/x86_64/puppet-agent-1.3.2-1.fedoraf22.x86_64.rpm

Note f22 appears in two places, "fedora/f22" and
"puppet-agent...-fedoraf22.x86_64.rpm"

### DIFF
--- a/lib/beaker/dsl/install_utils/foss_utils.rb
+++ b/lib/beaker/dsl/install_utils/foss_utils.rb
@@ -1112,6 +1112,7 @@ module Beaker
               release_path << release_path_end
             when /^(fedora|el|centos|sles)$/
               variant = ((variant == 'centos') ? 'el' : variant)
+              version = ((variant == 'fedora') ? "f#{version}" : version)
               release_path << "#{variant}/#{version}/#{opts[:puppet_collection]}/#{arch}"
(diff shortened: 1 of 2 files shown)


joshcooper:
This PR failed because of https://tickets.puppetlabs.com/browse/BKR-647. Basically, the test tries to curl `curl --retry 5 http://cygwin.com/setup-x86_64.exe -o /cygdrive/c/Windows/System32/setup-x86_64.exe` but cygwin.com is returning a 404.

kevpl:
failures unrelated to change. :+1: 

@joshcooper thanks for submitting this!

PR #1029, opened 2015-12-07
(BKR-569) added caching disable option

kevpl:


lib/beaker/dsl/helpers/web_helpers.rb
```diff
@@ -47,7 +47,7 @@ def fetch_http_file(base_url, file_name, dst_dir)
           base_url.chomp!('/')
           src = "#{base_url}/#{file_name}"
           dst = File.join(dst_dir, file_name)
-          if File.exists?(dst)
+          if options[:cache_files_locally] && File.exists?(dst)
```
colinPL: Should a warning to `STDOUT` be put here to provided notice of the up-coming change to change `:cache_files_locally` from `true` to `false`? Or will that change be communicated later?

kevpl: that change will be communicated later

spec/beaker/dsl/helpers/web_helpers_spec.rb
```diff
@@ -28,9 +34,20 @@ def fetch_allows
     describe "given valid arguments" do
 
       it "returns its second and third arguments concatenated." do
```
colinPL: Can this test also verify that `Already fetched` is logged?

kevpl: sure. done.

lib/beaker/options/presets.rb
```diff
@@ -185,6 +185,7 @@ def presets
           :puppetdb_port_nonssl   => 8080,
           :puppetserver_port      => 8140,
           :nodeclassifier_port    => 4433,
+          :cache_files_locally    => true, # TODO change to false in next major version
```
rick: If we're adding values to `presets`, we should probably bring in this spec which tests our defaults: https://github.com/puppetlabs/beaker/pull/1013/files#diff-e00c6e55b14629bb263ac9d4b5fb00b2

It was reverted along with the reversion of the `TestCase` refactoring for alternate test runners; it could be reintroduced here, or added in a separate PR.

colinPL:
retest this please

lib/beaker/options/presets.rb
kevpl: @rick would you mind doing that as a separate PR?

I have some questions about that code, so I'd rather not include it in my PR.

colinPL:
Failure is not related to patch.

PR #1031, opened 2015-12-08
(BKR-642) include aix exec, error checking to unix exec

kevpl:


lib/beaker/host/aix.rb
```diff
@@ -4,13 +4,14 @@
...
   class Host < Unix::Host
-    [ 'user', 'group', 'file' ].each do |lib|
+    [ 'user', 'group', 'file', 'exec' ].each do |lib|
         require "beaker/host/aix/#{lib}"
     end
 
     include Aix::User
     include Aix::Group
     include Aix::File
+    include Aix::Exec
```
joshcooper: Can we add specs that the correct commands are executed for an AIX host?

lib/beaker/host/unix/exec.rb
```diff
@@ -176,6 +176,8 @@ def ssh_service_restart
       exec(Beaker::Command.new("svcadm restart svc:/network/ssh:default"))
     when /(free|open)bsd/
       exec(Beaker::Command.new("sudo /etc/rc.d/sshd restart"))
+    else
+      raise ArgumentError.new("Unsupported Platform: '#{self['platform']}'")
```
rick: We don't have a style guide, so not a blocker, but this might be more idiomatic:

``` ruby
  raise ArgumentError, "Unsupported Platform: '#{self['platform']}'"
```

rick:
Other than adding more tests, :+1:.

kevpl:
killed it because I made another change. will kick in a sec

kevpl:

lib/beaker/host/aix.rb
kevpl: @joshcooper added tests

joshcooper:
:+1: pending CI checks

PR #1034, opened 2015-12-10
(BKR-237) disable updates.puppetlabs.com

erikPrime:


kevpl:
Thanks for submitting this @erikPrime!  I've got a few notes on the changes:

- please pre-pend your git commit messages with the ticket (`BKR-237`) reference
- please add some spec testing for this change
- `@options[:enable_updates]` needs a default set in the [`presets.rb` file](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/options/presets.rb#L120-L198)

The nice thing about adding your own setting, is that you could potentially default this to `false`, and not have to worry about setting it in hosts files.  However, I'm not sure if anyone else will be effected by this.

Another question to consider is whether or not the user should be notified of their update status with a log message or something similar.  Do you think this is fine as is, or do you think there'll be a group that'll be working on this functionality (or nearby) that might want to know what state their host is in in terms of update capabilities?

erikPrime:
@kevpl I've modified presets.rb, defaulting this new option to false, which modifies the hosts file to resolve updates.puppetlabs.com to 127.0.0.1.  I hope this will affect everyone.  We don't want anyone who is running beaker inside the company to be polluting the updates server with bad data.  It's possible this option is poorly named...

I added a log.notify to inform the user that their host file is being modified.  Personally I'd prefer to know.

kevpl:
@erikPrime it looks like your PR got polluted with recently merged commits, as well as a merge commit itself.  We don't usually merge PRs that don't have a clean commit history.  Would you please clean those up (you might have to cherry pick the commits)?

erikPrime:
@kevpl Yeah, I worked on that little annoyance last and hoped that you'd be slow enough as to not notice.  My bad.  Can you please look again?

PR #1036, opened 2015-12-15
(BKR-650) removes lines that modify self.hosts  Fixes confine :to logic

erikPrime:


lib/beaker/dsl/structure.rb
```diff
@@ -213,7 +213,6 @@ def confine(type, criteria, host_array = nil, &block)
           logger.warn "No suitable hosts with: #{criteria.inspect}"
           skip_test 'No suitable hosts found'
         end
-        self.hosts = hosts_to_modify
```
erikPrime: My understanding is that this modifies the global hosts array; We shouldn't do that.

lib/beaker/dsl/structure.rb
```diff
@@ -236,9 +235,7 @@ def confine_block(type, criteria, host_array = nil, &block)
           # a skip generated from the provided block, pass it up the chain
           raise e
         end
-      ensure
-        self.hosts = original_hosts
```
erikPrime: Don't modify that global hosts array.

erikPrime:
I need to consider the case where a call is made to confine with out a hosts argument ( and where hosts are iterated over in the block that is passed in.  Nate says I broke this.

PR #1041, opened 2015-12-29
(BKR-661) Change to use pkgng for package installs

petems:
* pkg_* is EOL as of Sept 2014 https://lists.freebsd.org/pipermail/freebsd-ports-announce/2014-February/000077.html
* On most base boxes, the pkg_* pointers are now 404:

```
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

pkg_add -r rsync

Stdout from the command:

Error: Unable to get ftp://ftp.freebsd.org/pub/FreeBSD/ports/amd64/packages-9-stable/Latest/rsync.tbz: File unavailable (e.g., file not found, no access)
```

* Also fixes (BK-438), as we're changing package install result to be the result, as is the same with other pkg helpers.

### DIFF
--- a/lib/beaker/host/freebsd/pkg.rb
+++ b/lib/beaker/host/freebsd/pkg.rb
@@ -2,17 +2,12 @@ module FreeBSD::Pkg
   include Beaker::CommandFactory
 
   def install_package(name, cmdline_args = nil, opts = {})
-    case self['platform']
-    when /freebsd-9/
-      cmdline_args ||= '-rF'
-      result = execute("pkg_add #{cmdline_args} #{name}", opts) { |result| result }
-    when /freebsd-10/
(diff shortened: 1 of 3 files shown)


colinPL:
:+1:

kevpl:
@petems please add a spec test to confirm this behavior.

One example of something that could be tested here is that in the past, we've had the situation come up where a subclass of `Beaker::Host` didn't include its own modules (`File`, `Pkg`, etc).  This led to Unix methods being used in place of the correct one, and this would error out because the code no longer took care of that platform.

Thanks for submitting this change!  I'm looking forward to getting it merged & released.

petems:
@kevpl Done! :smile:

PR #1042, opened 2015-12-29
(BKR-88) Split Validation Out Of Options Parser

colinPL:
This commit separates the options parser from the validation piece. This narrows the focus of Parser and makes the class smaller in code.

### DIFF
--- a/lib/beaker/options/parser.rb
+++ b/lib/beaker/options/parser.rb
@@ -16,13 +16,6 @@ module Beaker
       #The OptionsHash of all parsed options
       attr_accessor :options
 
-      # Raises an ArgumentError with associated message
-      # @param [String] msg The error message to be reported
-      # @raise [ArgumentError] Takes the supplied message and raises it as an ArgumentError
-      def parser_error msg = ""
-        raise ArgumentError, msg.to_s
-      end
-
       # Returns the git repository used for git installations
       # @return [String] The git repository
       def repo
@@ -32,7 +25,7 @@ module Beaker
       # Returns a description of Beaker's supported arguments
       # @return [String] The usage String
       def usage
-       @command_line_parser.usage
+        @command_line_parser.usage
       end
 
       # Normalizes argument into an Array.  Argument can either be converted into an array of a single value,
@@ -67,25 +60,25 @@ module Beaker
       #                         no .rb files are found overall
       def file_list(paths)
         files = []
-        if not paths.empty?
+        unless paths.empty?
           paths.each do |root|
             if File.file?(root)
               files << root
             elsif File.directory?(root) #expand and explore
               discover_files = Dir.glob(
-                File.join(root, "**/*.rb")
+                  File.join(root, "**/*.rb")
               ).select { |f| File.file?(f) }
               if discover_files.empty?
-                parser_error "empty directory used as an option (#{root})!"
+                @validator.parser_error "empty directory used as an option (#{root})!"
               end
-              files += discover_files.sort_by {|file| [file.count("/"), file]}
+              files += discover_files.sort_by { |file| [file.count("/"), file] }
             else #not a file, not a directory, not nothin'
-              parser_error "#{root} used as a file option but is not a file or directory!"
+              @validator.parser_error "#{root} used as a file option but is not a file or directory!"
             end
           end
         end
         if files.empty?
-          parser_error "no .rb files found in #{paths.to_s}"
+          @validator.parser_error "no .rb files found in #{paths.to_s}"
         end
         files
       end
@@ -106,13 +99,13 @@ module Beaker
         git_opts.map! { |opt|
           case opt
             when /^PUPPET\//
-              opt = "#{GITREPO}/puppet.git##{opt.split('/', 2)[1]}"
+              opt = "#{repo}/puppet.git##{opt.split('/', 2)[1]}"
             when /^FACTER\//
-              opt = "#{GITREPO}/facter.git##{opt.split('/', 2)[1]}"
+              opt = "#{repo}/facter.git##{opt.split('/', 2)[1]}"
             when /^HIERA\//
-              opt = "#{GITREPO}/hiera.git##{opt.split('/', 2)[1]}"
+              opt = "#{repo}/hiera.git##{opt.split('/', 2)[1]}"
             when /^HIERA-PUPPET\//
-              opt = "#{GITREPO}/hiera-puppet.git##{opt.split('/', 2)[1]}"
+              opt = "#{repo}/hiera-puppet.git##{opt.split('/', 2)[1]}"
           end
           opt
         }
@@ -125,8 +118,8 @@ module Beaker
       #defined.
       #@param [Hash] hosts A hash of hosts, each identified by a String name.  Each named host will have an Array of roles
       def set_default_host!(hosts)
-        default = []
-        master = []
+        default           = []
+        master            = []
         default_host_name = nil
 
         #look through the hosts and find any hosts with role 'default' and any hosts with role 'master'
@@ -139,12 +132,9 @@ module Beaker
           end
         end
 
-        if not default.empty?
-          #we already have a default set, do nothing
-          if default.length > 1
-            parser_error "Only one host may have the role 'default', default roles assigned to #{default}"
-          end
-        else
+        # default_set? will throw an error if length > 1
+        # and return false if no default is set.
+        unless @validator.default_set?(default)
           #no default set, let's make one
           if not master.empty? and master.length == 1
             default_host_name = master[0]
@@ -160,8 +150,9 @@ module Beaker
       #Constructor for Parser
       #
       def initialize
-         @command_line_parser = Beaker::Options::CommandLineParser.new
-         @presets = Beaker::Options::Presets.new
+        @command_line_parser = Beaker::Options::CommandLineParser.new
+        @presets             = Beaker::Options::Presets.new
+        @validator           = Beaker::Options::Validator.new
       end
 
       # Parses ARGV or provided arguments array, file options, hosts options and combines with environment variables and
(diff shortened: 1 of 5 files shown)


lib/beaker/options/parser.rb
```diff
@@ -335,55 +284,64 @@ def normalize_args
...
+        roles
+      end
+
+      # Get a unique list of hypervisors from list of host.
       #
-      # @return nil
-      # @api public
-      def normalize_and_validate_tags()
+      # @param [Array] hosts beaker hosts
+      # @return [Array] unique list of hosts
```
kevpl: should this be a list of hypervisors?

colinPL: Yes.

lib/beaker/options/validator.rb
```diff
@@ -0,0 +1,131 @@
...
+
+      # resolves all file symlinks that require it.
+      #
+      # @param [Beaker::OptionsHash] options Beaker Options hash
+      # @note doing it here allows us to not need duplicate logic, which we
+      #   would need if we were doing it in the parser (--hosts & --config)
+      #
+      # @return nil
+      # @api public
+      def resolve_symlinks(options)
```
kevpl: I'm confused as to why this was moved.  Is there a situation where you can fail this?

lib/beaker/options/parser.rb
```diff
@@ -67,25 +60,25 @@ def split_arg arg
...
             if File.file?(root)
               files << root
             elsif File.directory?(root) #expand and explore
               discover_files = Dir.glob(
-                File.join(root, "**/*.rb")
+                  File.join(root, "**/*.rb")
               ).select { |f| File.file?(f) }
               if discover_files.empty?
-                parser_error "empty directory used as an option (#{root})!"
+                @validator.parser_error "empty directory used as an option (#{root})!"
```
kevpl: These appear to be actual errors in the parsing.  If this is the case, why doesn't the parser just throw parser errors rather than validator's parser errors?

lib/beaker/options/validator.rb
```diff
@@ -0,0 +1,131 @@
...
+          # select items from exclude set that match included_tag
+          # no match is an empty list/array/[]
+          if exclude.select { |ex| ex == included_tag } != []
+            validator_error "tag '#{included_tag}' cannot be in both the included and excluded tag sets"
+          end
+        end
+
+      end
+
+      def valid_frictionless_roles?(role_array)
```
kevpl: weird that this can return `nil` when it passes validation.  The question mark ending implies that you can treat this method's return as a boolean value.  These methods would be clearer without the question mark & with `validate_`.. as the prefix, or if they returned boolean values.

lib/beaker/options/validator.rb
colinPL: Only if the file path causes `File.realpath` to throw an error. I am going to put it back in `parser.rb`.

lib/beaker/options/validator.rb
colinPL: This has been fixed for all methods that end in `?`, but do not return `bool`.

lib/beaker/options/parser.rb
colinPL: The intention was to provide one way to raise an error. The parser knows about the validator, but not the other way around.

lib/beaker/options/parser.rb
```diff
@@ -67,29 +60,39 @@ def split_arg arg
       #                         no .rb files are found overall
       def file_list(paths)
         files = []
-        if not paths.empty?
+        unless paths.empty?
```
rick: We don't have a style guide here, so there's no objective way to rule on this. In style guides I've lived under in the past, we had a couple of things that could apply here:

 - "Don't use the block form of `unless`" (i.e., only use inline `unless`)
 - "Always use `!` instead of `not`"

No change needed, just highlighting this and our continued lack of a real ruby style guide.

lib/beaker/options/parser.rb
```diff
@@ -32,7 +25,7 @@ def repo
       # Returns a description of Beaker's supported arguments
       # @return [String] The usage String
       def usage
-       @command_line_parser.usage
+        @command_line_parser.usage
```
rick: Not for this PR, as beaker is littered with this, but raising awareness (at more than a "style guide" level, as it has fairly deep practical implications): I recommend almost always using accessor methods (whether defined by `attr_*` or defined explicitly) instead of instance variable references (obvious exceptions include when the reference is inside an accessor method definition, when setting initially in a constructor, and so on).

The value of going this route shows up when refactoring, deprecating, auditing/instrumenting access, subclassing, etc., as instance variables work counter to these efforts. I've seen this as a "best practice" in smalltalk and java as well for similar reasons.

Again, not something to change here, just taking the opportunity to have the discussion.

lib/beaker/options/parser.rb
```diff
@@ -67,29 +60,39 @@ def split_arg arg
...
-              files += discover_files.sort_by {|file| [file.count("/"), file]}
-            else #not a file, not a directory, not nothin'
-              parser_error "#{root} used as a file option but is not a file or directory!"
+              path_files = Dir.glob(File.join(root, '**/*.rb'))
+                               .select { |f| File.file?(f) }
+                               .sort_by { |file| [file.count('/'), file] }
             end
+
+            @validator.validate_files(path_files, root)
+            files += path_files
```
rick: :+1: much easier to read and understand.

lib/beaker/options/parser.rb
```diff
@@ -67,29 +60,39 @@ def split_arg arg
...
           end
         end
-        if files.empty?
-          parser_error "no .rb files found in #{paths.to_s}"
-        end
+
+        @validator.validate_files(files, paths.to_s)
```
rick: Would it be sufficient for this to be the only call to `validate_files`? It seems like it throws an exception, and all we've done is add files here.

Another approach (I don't know without deeper spelunking if that method is used outside the files in this PR) would be to simply do the empty check that the method does, here inline. Is an exception needed, or could we pass back an empty result and expect callers to do the right thing?

lib/beaker/options/parser.rb
```diff
@@ -139,12 +142,9 @@ def set_default_host!(hosts)
           end
         end
 
-        if not default.empty?
-          #we already have a default set, do nothing
-          if default.length > 1
-            parser_error "Only one host may have the role 'default', default roles assigned to #{default}"
-          end
-        else
+        # default_set? will throw an error if length > 1
```
rick: Might be outside the scope of this PR, but if we're interested in cleanup in here, this block might be a candidate for a method extraction. This `set_default_host!` method is getting a bit unwieldy.

lib/beaker/options/validator.rb
```diff
@@ -0,0 +1,132 @@
...
+          validator_error("No files found for path: '#{path}'")
+        end
+      end
+
+      def validate_path(path)
+        unless File.file?(path) || File.directory?(path)
+          validator_error("#{path} used as a file option but is not a file or directory!")
+        end
+      end
+
```
rick: Stylistic nit, I'm personally not a fan of the extra newlines before the `end`s at the end of files. I see it a lot here (and even more idiosyncratic newline applications in some other projects) and I'm not sure I understand it.

lib/beaker/options/parser.rb
colinPL: The old code for this method (`file_list`) raises errors when:

* an empty directory is provided (`@validator.validate_files(path_files,root)`)
* a path is not a file or directory (`@validator.validate_path(root)`)
* if the final `files` array is empty (`@validator.validate_files(files, paths.to_s)`)

This was an attempt to make the code easier to read, move errors to `Validator`, while trying to retain the detailed error messages.

rick:
Going back to my "[Working Effectively with Legacy Code](http://www.amazon.com/Working-Effectively-Legacy-Michael-Feathers/dp/0131177052)" roots, this sort of refactoring is somewhat risky.  There are really a few phases of things that I would expect to be happening here that would show me, the reviewer, that this is a safe refactoring:

 - If there is insufficient test coverage (likely), then I would expect to see up-front commits bolstering test coverage to support the class extraction.
 - If it is impossible / difficult to do that literal extraction, I would expect to also see a series of commits that introduce sensing variables, split up methods, create seams, etc., so that the original class is modified only to the extent that it needs to be to get test coverage around the extraction that's about to happen.
 - Then I would expect to see test-supported extraction of the literal class (i.e., without cleanups or refactoring) in a short series of commits.
 - Then I would expect to see a set of refactoring commits that does various bits of cleanup to make things cleaner, more readable, etc.

This is one of those things where, even if the ultimate result is to squash everything down to one or a few commits, for the review process, and for the historical record, the proof-by-commits that we're doing a refactoring safely is of value.

rick:
Just wanting to follow-up on the last comment, with a couple of thoughts:

I can definitely :+1: a PR that doesn't have that sort of commit history that I'm talking about, but know that   we reduce our risk a lot (with or without the explicit commit history) by following that sort of process when we do an extraction -- especially on beaker, which has weird usages all over the place inside and outside the walls (every weird edge case is being exploited by someone).

If we want to work more collaboratively on how to do a refactoring like I'm talking about ("we" being me and you, a larger group, ESO CodeCraft, etc.), I'm game, as I think this is exactly the sort of not huge but not trivial refactoring that's a good case study for such an effort.

I definitely recommend the book I linked.  At the risk of gaining eternal ridicule (well, more than I'm already getting generally), I'll also link to a talk I gave back in 2008 (I shudder at my appearance and the more intense southern accent, but I think the material is still relevant) on bringing test coverage to the ruby `flog` tool, in support of a feature-addition refactoring:

[flog << Test.new](http://confreaks.tv/videos/rubyhoedown2008-flog-test-new)

colinPL:
Failure is unrelated to PR code changes.

colinPL:
@puppetlabs-jenkins ok to retest

colinPL:
retest this please

kevpl:
I'm :+1:.  @colinPL let me know if you're ready for merge.

rick:
:+1:

PR #1043, opened 2016-01-02
(MAINT) Correcting method name in docs

timhughes:


kevpl:
@timhughes good catch!  Would you update your git message, prefixing it with `(MAINT)`, so this can be classified correctly?

Thanks!  I'm :+1: other than the contributor change above, and I look forward to getting this merged & released!

timhughes:
Is that what you mean or would you like me to make a new commit and rebase

kevpl:
new commit and rebase, please.  Actually, internally, we usually amend the last commit for something like this.  But that does require a forced push, so whichever you'd prefer.

timhughes:
done :-)

kevpl:
perfect!  Thanks so much!

PR #1046, opened 2016-01-11
(BKR-666) added wrlinux5 provisioning steps

kevpl:


kevpl:
rebased

lib/beaker/host/unix.rb
```diff
@@ -138,6 +138,23 @@ def external_copy_base
       @external_copy_base
     end
 
+    # Tells you whether a host platform supports beaker's
+    #   {Beaker::HostPrebuiltSteps#set_env} method
+    #
+    # @return [String,nil] Reason message if set_env is not supported,
+    #   nil if it is supported
+    def supports_set_env
```
justinstoller: nit: It seems odd to me that a method `supports_set_env` returns `nil` to communicate that it does support that action and a truthy value to say that it doesn't support that action.

justinstoller:
:+1: on the approach and the code, save for method name vs outputs of `supports_set_env`.

lib/beaker/host/unix.rb
kevpl: I do agree that it's odd, but I think it's actually the correct way, in this case.

If the method were to be named according to its return value's truthiness, then it would have to be negative.  Negatively named methods are more complex, because the two situations you end up with are these:

1. not something (the method name)
2. not not something (negating the method name)

whereas having just a postively named method ends up having these scenarios:

1. something (method name)
2. not something (negating)

I would also contend that the fact that Ruby has a specific convention for boolean return methods (ending with a question mark) means that my method name implies that it shouldn't be used as a boolean return, and is documented appropriately to match.

If I were to take the altogether different approach of having two return values, one for the actual boolean test and the other being the failure message, then I still wouldn't be able to add a question mark to the end, because it would always return an array (I believe that's what the return value would be boolean resolved to since it's usually implicitly destructured), which would always resolve to true.

@rick, it seems like the most OOP way to resolve this issue would be to have a return object that holds two values, the test value (boolean) and message (string).  The object itself's truthiness would resolve to the test value, and have an accessor to get the message.  However, this seems like a serious amount of overhead for something this small.  What do you think?

colinPL:
retest this please

colinPL:
retest this please

lib/beaker/host/unix.rb
justinstoller: So I don't know that this is necessarily worth a lot of discussion (I'm not going to die on this hill). I grok your observation about negative methods, though I think the current method name and return values are just as confusing as the double negative instance.

If the method name can't be in line with the return value (truthy values answer "yes" to the question the method name asks) like `skip_setting_env()`.

Then I walk through the same path that you outlined, but I see it as, we're answering two questions with the one method, "do we skip setting the environment up for the platform?" and "why are we skipping setting up the environment?" I would argue that those two questions are similar enough that they don't warrant two method calls, or its own object[1], and would return two answers to those questions.

Again, not my hill. So feel free to ignore me. What do other's think? @colinPL ? @waynr ? (since Rick was already pinged).

1. unless this, and more complicated issues, are happening in many places and could be cleaned up with a Result object. But I would still put that out of scope of this PR.

kevpl: looking at this this morning, I'm realizing your suggested name is much clearer.

I've updated the name.  Thanks for standing up to the wall of text @justinstoller !

justinstoller: Walls of text are my preferred form of communication   :)


(srsly, you can give _what_ you think concisely, but I find it much harder to concisely explain _why_ you think that way, and understanding the why is the interesting part of any discussion)

justinstoller:
:+1: pending test outcome

kevpl:

kevpl:

colinPL:
restest this please

PR #1049, opened 2016-01-14
(MAINT) Remove keepalive_interval

colinPL:
This commit removes the keepalive_interval setting in presets.rb. This
was introduced to detect disconnections that were otherwise not noticed
unless beaker tried to write to the SSH session.

This change does still have keep alive enabled, but is now adhering to
the default interval of 300 (5 minutes).


joshcooper:
:+1:

johnduarte:
:+1: 
waiting for pr test run to complete

PR #1050, opened 2016-01-22
(BKR-685) Support openstack user_data

azhurbilo:


rick:
Looks :+1: to me, but I'm not familiar with the code in this area.

/cc @puppetlabs/beaker

colinPL:
Please add tests.

azhurbilo:
@colinPL , there is no any test for Openstack hypervisor.
Do I need create new spec file with 'user_data' test only?

azhurbilo:
@colinPL , I've added tests. Could you check PR?

spec/beaker/hypervisor/openstack_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+require 'fog'
+
+module Beaker
+  describe OpenStack do
+
+    let(:options) { make_opts.merge({'logger' => double().as_null_object}) }
+
+    let(:openstack) {
+      stub_const("Fog::Compute", MockComputeClient)
+      stub_const("Fog::Network", MockNetworkClient)
```
kevpl: @azhurbilo is there a reason that you need two mock files for these, rather than just creating a few stubbed objects in this file?

azhurbilo: @kevpl I'am not experienced in ruby and rspec, but I don't know how to test such cases without mock when you need response and execute one more method of returned object.

For example:
```
vm = @compute_client.servers.create(options)
```

kevpl: @azhurbilo if you checkout this [puppet_helpers_spec code](https://github.com/puppetlabs/beaker/blob/master/spec/beaker/dsl/helpers/puppet_helpers_spec.rb#L778-L781), you can see an example of how we do it.  A better way would be to have a delegator method so you don't have to mock so deeply, but stub instead.  But if you're just looking to get a test for this going quickly I would do something like this for that line:
```ruby
    mock_servers = Object.new
    allow( mock_servers ).to receive( :create )
    allow( @compute_client ).to receive( :servers ).and_return( mock_servers )
```
or potentially change the create line to an expect if that's what you're testing
```ruby
    expect( mock_servers ).to receive( :create ).with( options_answer )
```
let me know if any of this is unclear, and we can discuss it further if needed.

kevpl:
last comment is still current. If this is not commented on or touched in a week, it'll be closed due to inactivity. Please let me know if you have any questions about this.

PR #1051, opened 2016-01-25
(BKR-668) wrlinux5 puppet-agent install support added

kevpl:
Adds support for both puppet-agent install methods on wrlinux-5:
`install_puppet_agent_on` (from yum.puppetlabs.com)
`install_puppet_agent_dev_repo_on` (from builds.puppetlabs.net)

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -331,7 +331,7 @@ module Beaker
     # to that source dir.
     #
     # @param source [String] The path to the file/dir to upload
-    # @param target [String] The destination path on the host
+    # @param target_path [String] The destination path on the host
     # @param options [Hash{Symbol=>String}] Options to alter execution
     # @option options [Array<String>] :ignore An array of file/dir paths that will not be copied to the host
     # @example
@@ -340,9 +340,8 @@ module Beaker
     #
     #   do_scp_to('source/file.rb', 'target', { :ignore => 'file.rb' }
     #   -> will result in not files copyed to the host, all are ignored
-    def do_scp_to source, target, options
-      self.scp_prep_operations( target )
-      target = self.scp_path(target)
+    def do_scp_to source, target_path, options
+      target = self.scp_path( target_path )
       @logger.notify "localhost $ scp #{source} #{@name}:#{target} {:ignore => #{options[:ignore]}}"
 
       result = Result.new(@name, [source, target])
@@ -410,7 +409,7 @@ module Beaker
         end
       end
 
-      self.scp_post_operations( target )
+      self.scp_post_operations( target, target_path )
       return result
     end
 
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -341,6 +341,7 @@ module Beaker
     #   do_scp_to('source/file.rb', 'target', { :ignore => 'file.rb' }
     #   -> will result in not files copyed to the host, all are ignored
     def do_scp_to source, target, options
+      self.scp_prep_operations( target )
       target = self.scp_path(target)
       @logger.notify "localhost $ scp #{source} #{@name}:#{target} {:ignore => #{options[:ignore]}}"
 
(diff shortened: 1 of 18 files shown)


kevpl:
ping @LuvCurves for test & review.

kevpl:
:open_mouth: missed yard doc tests.  It was just a copied method signature comment that didn't get refactored correctly.  Updated, amended, & pushed.

LuvCurves:
I have verified this works in the test environment like we expect.

LuvCurves:
Found issue with call "on agent, puppet('--version')". Getting 2 prepends instead of 1
"source /etc/profile; sudo ip netns exec management source /etc/profile; sudo ip netns exec management puppet --version"

kevpl:
@LuvCurves I've fixed the command prefix doubling.  Please test this out, and let me know if you're still seeing any issues there.

LuvCurves:
scp is failing when it's done from test runner to cisco because it appears to be doing it with the 'root' account instead of 'beaker'

kevpl:
@LuvCurves it's SCPing as beaker, but the files are owned by root at that time, that's why the permission failure.  I've put in a fix that will solve the issue for all SCPing to cisco hosts going forward.

For fun:

    Finished in 15.19 seconds (files took 2.51 seconds to load)
    1337 examples, 0 failures

LuvCurves:
Nice work. You're correct, the file was owned by root and had permissions set to 600.  This last bit resolved my issue with the hiera acceptance test.

kevpl:

kevpl:
@LuvCurves things look good to me from here.  I know your testing is ongoing, so let me know here what happens as you go through that, and please give a final :+1: when your testing passes so we can get this merged.  Thanks for your help testing this!

kevpl:
killed the test myself, as I had a code update for that particular fix.  Will re-kick manually

kevpl:

kevpl:

LuvCurves:
I've done some extensive testing with the latest changes and it's working great.

kevpl:
We've found that the failure here is unrelated to these changes, as it's failing on master & in the modules pipelines in [this issue](https://tickets.puppetlabs.com/browse/BKR-695).

Since that's the case, this is ready for merge @puppetlabs/beaker!

lib/beaker/host.rb
```diff
@@ -340,8 +340,8 @@ def exec command, options={}
     #
     #   do_scp_to('source/file.rb', 'target', { :ignore => 'file.rb' }
     #   -> will result in not files copyed to the host, all are ignored
-    def do_scp_to source, target, options
-      target = self.scp_path(target)
+    def do_scp_to source, target_path, options
+      target = self.scp_path( target_path )
```
waynr: yay for more semantically correct parameter names!

waynr:
:+1: Looks good to me!

lib/beaker/host/unix/file.rb
```diff
@@ -17,7 +17,13 @@ def system_temp_path
   #
   # @note This is really only needed in Windows at this point. Refer to
```
doug-rosser: This comment seems odd, given the code below.

doug-rosser:
Other than my quibble/question above, :+1:

rick:
Only thing that stands out to me is additions of `/cisco/`, `/cisco-/`, and `/cisco-5/` -- is there a reason why one would be used over another, or are they all effectively equivalent?

LuvCurves:
@rick There are actually 2 Cisco platforms we need to support cisco-wrlinux-5 (aka cisco-5) and cisco-wrlinux-7 (aka cisco-7).  I don't think we actually need /cisco/ but /cisco-/ would cover both platforms.

kevpl:
@rick it should be that variant checks should be made against `cisco` itself, and platform checks should at least be against `cisco-`, unless the code really is specific for a version, hence `cisco-5`.

rick:
:+1:

PR #1052, opened 2016-01-26
Add info log when teardown begins

samwoods1:
I didn't consult anyone on this, but it is confusing to me when looking at logs, when my test case completes and when the teardown begins.  This message would help remove some of that confusion, especially in the case where you run into https://tickets.puppetlabs.com/browse/BKR-570

### DIFF
--- a/lib/beaker/test_case.rb
+++ b/lib/beaker/test_case.rb
@@ -136,6 +136,7 @@ module Beaker
             rescue StandardError, ScriptError, SignalException => e
               log_and_fail_test(e)
             ensure
+              @logger.info('Begin teardown')
               @teardown_procs.each do |teardown|
                 begin
                   teardown.call
@@ -143,6 +144,7 @@ module Beaker
                   log_and_fail_test(e)
                 end
               end
+              @logger.info('End teardown')
             end
           end
           @sublog = @logger.get_sublog


lib/beaker/dsl/structure.rb
```diff
@@ -70,6 +70,7 @@ def test_name my_name, &block
       #       'ensure=absent', 'purge=true'))
       #   end
       def teardown &block
+        logger.info('Begin teardown')
```
nwolfe: This is a good idea! We might also consider adding an `End teardown` message too.

I admit I haven't tested this, but just looking at the code it seems this is the wrong place for this log message to be printed.  I believe this method is how a test registers a block for callback during the teardown phase -- this isn't actually where/when the test suite is performing teardown.

I think we want to add the log messages before/after we call the blocks, which looks like here: https://github.com/puppetlabs/beaker/blob/master/lib/beaker/test_case.rb#L139-L144

samwoods1:
Ah, good call.  I was forgetting that we can have multiple teardown blocks.  Updated.

kevpl:
:+1:.  Please add a `(MAINT)` prefix to your git commit message, and we'll get this merged.

samwoods1:
Ping, I think this is ready for merge?

nwolfe:
:+1:

PR #1053, opened 2016-01-27
(BKR-637) Support mounting volumes with docker

nhemingway:
Wasn't sure whether I was allowed to take over the JIRA ticket, so I left it "Accepted".  Please let me know if I should do otherwise.

### DIFF
--- a/docs/Docker-Support.md
+++ b/docs/Docker-Support.md
@@ -89,11 +89,11 @@ You can mount folders into a docker container:
         hypervisor: docker
         mount_folders:
           name1:
-            from: host_path1
-            to: container_path1
+            host_path: host_path1
+            container_path: container_path1
           name2:
-            from: host_path2
-            to: container_path2
+            host_path: host_path2
+            container_path: container_path2
             opts: rw
     CONFIG:
       type: foss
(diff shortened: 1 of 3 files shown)


docs/Docker-Support.md
```diff
@@ -79,6 +79,25 @@ ssh:
...
+You can mount folders into a docker container:
+
+    HOSTS:
+      ubuntu-12-10:
+        platform: ubuntu-12.10-x64
+        image: ubuntu:12.10
+        hypervisor: docker
+        mount_folders:
+          name1:
+            from: host_path1
```
electrical: Maybe call from/to host_path/container_path ?
makes it a bit more verbose but a lot clearer.

electrical:
Looks good code wise and API call. left a small remark about the naming.

docs/Docker-Support.md
nhemingway: from/to was taken from the ticket.  I assume that's because that's what they're named for vagrant.

I prefer your more explicit names, and have changed them

electrical: Ahh okay. I see.
I think its much clearer this way for people using it.

kevpl:

kevpl:
@nhemingway, yeah, please assign the ticket to yourself.

PR #1054, opened 2016-01-27
(maint) move recipes from confluence

er0ck:
This change adds a document containing some recipes for solving common
problems with Beaker.  It is currently housed on confluence.


kevpl:
Would you be opposed to having the code in code blocks rather than in lists?

Also, please add a link to your doc from the main docs page.

er0ck:
done

PR #1057, opened 2016-02-04
added option to skip provisioning of a host

hdoedens:
For Puppet beaker testing I want to reuse a puppetmaster host in vsphere. I do not want it to reset to a specific point in time, because it would require updating all puppet modules again which costs time. 
From beaker-rspec I need to copy resources to the puppetmaster like hieradata, so having the host available through the nodeset config is really helpful.

I have added a parameter `provision` to skip vsphere provisioning, but still be able to reach the host from rspec.

### DIFF
--- a/spec/beaker/hypervisor/vsphere_spec.rb
+++ b/spec/beaker/hypervisor/vsphere_spec.rb
@@ -58,6 +58,20 @@ module Beaker
 
       end
 
+      it 'does not provisions hosts if provision is set to false' do
+        MockVsphereHelper.powerOff
+        hosts = make_hosts()
+        hosts[0]["provision"] = false
+        vsphere = Beaker::Vsphere.new( hosts, make_opts )
+
+        vsphere.provision
+
+        hosts.each do |host|
+          expect( MockVsphereHelper.find_vm( host.name ).powerState ) == "poweredOff"
+        end
+
+      end
+
     end
 
     describe "#cleanup" do
--- a/lib/beaker/hypervisor/vsphere.rb
+++ b/lib/beaker/hypervisor/vsphere.rb
@@ -18,7 +18,7 @@ module Beaker
       vsphere_helper = VsphereHelper.new( vsphere_credentials )
 
       vsphere_vms = {}
-      @hosts.each do |h|
+      @hosts.reject do |host| host[:provision] == false end.each do |h|
         name = h["vmname"] || h.name
         vsphere_vms[name] = h["snapshot"]
       end


kevpl:
Hi @hdoedens!  Thanks for contributing these changes to beaker, we're looking forward to talking about these with you.

In order for these to be merged into beaker, the changes will have to meet our contributor standards, which can be seen in the top level CONTRIBUTING.md doc.  The only changes required for your PR to meet our standards are as follows:

1. The git commit message will need to be prefixed by either a `(MAINT)` or Beaker ticket ref (ie `(BKR-999)`).  You can create a beaker ticket at our Jira project [here](https://tickets.puppetlabs.com/browse/BKR).
2. Please include a spec test, so that we can be sure we don't break your functionality in the future.

Once these are done, we can look at getting your changes through our pipline and merged!  Let me know if you have any questions about this.

hdoedens:
@kevpl Thank you for your response. 
I have created an issue (BKR-709) and amended the commit message. I have also added a spec test although I am not what the value of those tests are as I was unable to let mine or existing tests fail by modifying the expectations.
Anyhow, please let me know if I need to undertake more action.

kevpl:
You weren't able to get your test to fail?  I would imagine in this test case you'd set up your hosts to have `host[:provision]` be false, and then verify that the `name`, `vsphere_vms`, and whatever other consequences you're setting are behaving appropriately.

I usually comment out my changes after making a test, just to run one last verification that it does fail without my code, & then succeed with my code just before submitting.

Let me know when you feel better about your testing code, and we'll take a look at it.

kevpl:
PR for [BKR-709](https://tickets.puppetlabs.com/browse/BKR-709).

hdoedens:
What worries me most is that other tests also succeed when I change the expectation. For example when I change the expectation of the test on `line 47` of `spec/beaker/hypervisor/vsphere_spec.rb` (provisions hosts if no snapshot is provided) from:
`expect( MockVsphereHelper.find_vm( host.name ).powerState ) == "poweredOn"`
to 
`expect( MockVsphereHelper.find_vm( host.name ).powerState ) == "foobar"`
the test still succeeds. It seems as these tests are userless but I can't figure out what is wrong.

spec/beaker/hypervisor/vsphere_spec.rb
```diff
@@ -58,6 +58,20 @@ module Beaker
...
+      it 'does not provisions hosts if provision is set to false' do
+        MockVsphereHelper.powerOff
+        hosts = make_hosts()
+        hosts[0]["provision"] = false
+        vsphere = Beaker::Vsphere.new( hosts, make_opts )
+
+        vsphere.provision
+
+        hosts.each do |host|
+          expect( MockVsphereHelper.find_vm( host.name ).powerState ) == "poweredOff"
```
kevpl: it looks like you might be using the wrong syntax for the expect statement here.  When we're looking to test a matching state on an object, we typically use this syntax:

```ruby
expect( object.property ).to be === correct_answer
```

it looks like this line is missing the `.to be ===` section.  I'm not sure if what the `expect` method returns can be compared to a desired state.

hdoedens: ahh, thanks. That does indeed what I would expect from my previous code.

Please note that existing tests are also wrong and should be fixed as well.

hdoedens: @kevpl Your comment is entirely true. That also means that 3 existing vsphere tests are useless at this time...

I tried to write a good test but I can't get it to work. This is the closest I could get.
```
it 'does not resets to the given snapshot if provision is set to false' do
  MockVsphereHelper.powerOff
  hosts = make_hosts()
  hosts[0]["provision"] = false
  vsphere = Beaker::Vsphere.new( hosts, make_opts )
  expect( MockVsphereHelper ).to receive( :find_snapshot ).exactly( hosts.length - 1 ).times
  vsphere.provision
end
```
Rather dull, but it seems that the effect of this feature can already be achieved by simply not providing a snapshot... my bad :|

PR #1058, opened 2016-02-05
(maint) Fix Windows user_list functionality

Iristyle:
- Existing functionality used an overly restrictive regex that didn't
   account for characters like - or . in Windows user names, which are
   completely valid.

   Use a less restrictive regex to capture the value correctly.  The
   results of wmic always include the username immediately after the
   = in the Name= string


Iristyle:
FYI @joshcooper / @puppetlabs/windows

kevpl:

PR #1060, opened 2016-02-08
(BKR-696) Fix :ignore corruption in host.rb

jeffcoat:
do_rsync_to was overwriting its :ignore option,
which broke things when it was called in a loop.

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -482,10 +482,10 @@ module Beaker
       end
 
       if opts.has_key?(:ignore) and not opts[:ignore].empty?
-        opts[:ignore].map! do |value|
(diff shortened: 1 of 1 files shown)


kevpl:
looks good, and thanks for contributing this @jeffcoat! Would you please add a spec test, so we can make sure that we don't break your fix going forward?

kevpl:
last comment still current

kevpl:
last comment still current.  If there's still no movement on this PR in a week we'll close it out due to inactivity.

tvpartytonight:
As there hasn't been any movement for spec tests on this PR, I'm just going to close it. Please feel free to reopen with tests!

PR #1062, opened 2016-02-10
(BKR-667) added wrlinux7 provisioning support

kevpl:


kevpl:

kevpl:
@puppetlabs/beaker ready for review!

waynr:
:+1: looks good to me

lib/beaker/host/unix/exec.rb
```diff
@@ -233,15 +233,8 @@ def ssh_permit_user_environment
...
-
     environment_string = env_array.join(' ')
-    command = 'env'
-    punctuation = ''
-    if self[:platform] =~ /cisco-5/
-      command = 'export'
-      punctuation = ';'
-    end
-    "#{command} #{environment_string}#{punctuation}"
+    "env #{environment_string}"
```
sbeaulie: the previous code used 'export' when platform is cisco-5, presumably because 'env' cannot be used to load variables? I have no idea if that's the case but this env vs export logic is not here

kevpl: in the refactor of the cisco code, we made a copy of the `environment_string` method for cisco.  You'll see that the logic has been moved there, and should be able to check that Cisco hosts inherit from Unix hosts correct, are instantiated in `host.rb` as needed for this to work.

pinkypie:
:+1:

PR #1063, opened 2016-02-10
(BKR-48) Change how --dry_run is defined - should not be global variable.

pinkypie:
Removed $dry_run as a global variable.
Should be able to overwrite value from calling methods or use the parsed options hash value as default.
Returning from the original method if dry_run instead of continuing on and attempting a host connection.

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -265,8 +265,17 @@ module Beaker
     end
 
     def exec command, options={}
+      result = nil
       # I've always found this confusing
       cmdline = command.cmd_line(self)
+
+      # use the value of :dry_run passed to the method unless
+      # undefined, then use parsed @options hash.
+      options[:dry_run] ||= @options[:dry_run]
+
+      if options[:dry_run]
+        @logger.debug "\n Running in :dry_run mode. Command #{cmdline} not executed."
+      end
 
       if options[:silent]
         output_callback = nil
@@ -279,11 +288,10 @@ module Beaker
         end
       end
 
-      unless $dry_run
+      unless options[:dry_run]
         # is this returning a result object?
         # the options should come at the end of the method signature (rubyism)
         # and they shouldn't be ssh specific
-        result = nil
 
         @logger.step_in()
         seconds = Benchmark.realtime {
@@ -320,9 +328,8 @@ module Beaker
             raise CommandFailure, "Host '#{self}' exited with #{result.exit_code} running:\n #{cmdline}\nLast #{@options[:trace_limit]} lines of output were:\n#{result.formatted_output(@options[:trace_limit])}"
           end
         end
-        # Danger, so we have to return this result?
-        result
       end
+      result
     end
 
     # scp files from the localhost to this test host, if a directory is provided it is recursively copied.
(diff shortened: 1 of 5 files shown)


waynr:
All the style/formatting changes make it difficult to understand what is significant in this PR. Is this something you did intentionally or does your text editor apply coding style preferences automatically?

lib/beaker/host.rb
```diff
@@ -42,33 +43,33 @@ def [](k)
 
     def self.create name, host_hash, options
       case host_hash['platform']
-      when /windows/
-        cygwin = host_hash['is_cygwin']
-        if cygwin.nil? or cygwin == true
-          Windows::Host.new name, host_hash, options
+        when /windows/
```
kevpl: in beaker, we've always standardized on the `when` clause going at the same level of the `case` in a case statement.

lib/beaker/host.rb
```diff
@@ -504,13 +530,13 @@ def do_rsync_to from_path, to_path, opts = {}
...
-    'windows',
-    'pswindows',
-    'eos',
+      'unix',
+      'aix',
+      'mac',
+      'freebsd',
+      'windows',
+      'pswindows',
+      'eos',
```
kevpl: our old spacing is what we want to use here, since we use 2 spaces rather than 4 for indenting.  If there's some other reason to do this that I'm not seeing, please let me know, and we can discuss this further.

kevpl:
I'm fine with most of these formatting changes, but some of them go against conventions we've settled on (although haven't documented :\) in beaker.  I've pointed these out in the code, & they will need to be set back to move forward with this, unless there are cases to be made for them.

pinkypie:
I had a ruby style applied in RubyMine for Sweep stuff, I'll switch it back.

kevpl:
Nice work getting these tests to pass!

Looking at this code, I'm a little worried that we're changing the usage pattern of these methods by doing this.  In this world, in order for all callers of these methods to be able to handle dry-run calls, they would need to be able to handle (& check for) the null return.  That can be pretty painful, especially because these methods are so central to beaker's execution.

I've done some local testing with this branch (using an install puppet pre-suite and simple puppet smoke test), and I am seeing that pain locally in `NoMethodError`s returned.  Making some simple changes to use the Null Object pattern ([1](http://refactoring.com/catalog/introduceNullObject.html), [2](https://en.wikipedia.org/wiki/Null_Object_pattern)) instead of returning `nil`, I'm able to see this same beaker setup pass as you would expect dry-run to (or at least more like you'd expect dry-run to :wink:).  Check out [my changes](https://github.com/pinkypie/beaker/compare/BKR-48...kevpl:bkr48?expand=1), and let me know what you think.

pinkypie:
Yes that makes a lot of sense. It seemed like the way it worked previously, it didn't assign the result a value so probably returned nil?, or eventually returned nil (scp_to and scp_from). But it wasn't super clear and I think returning a result object is much better in general. Should we do the same for the do_scp_from method to keep them all in sync?

kevpl:
Yep, looks great!  I'm sorry to block merge on this, but would you please update your git commit messages to all be prefixed with the ticket in the form `(BKR-48)` ?  Sorry about that, I should have seen that yesterday, so that we could have gotten you these sprint points before planning :cry:

pinkypie:
@kevpl - I think that's them fixed up if you get a chance before the planning meeting!

kevpl:
Sorry to be a stickler on this, but the ticket reference needs to be surrounded by parentheses to match our [CONTRIBUTING](https://github.com/puppetlabs/beaker/blob/master/CONTRIBUTING.md) standards

pinkypie:
@kevpl  no probs, should have read the docs :-)

PR #1064, opened 2016-02-10
(MAINT) added cisco platform doc

kevpl:
Dependent on #1062.

Just docs changes, shouldn't need to wait/worry about Jenkins testing.  Probably should have included that in the git message, but oh well.  I'll try here.

[skip ci]


waynr:
:+1:

PR #1065, opened 2016-02-17
(BKR-704) Allow relative `mount_folders` in docker

op-ct:
Without this patch, a nodeset using relative `from:` paths under
`mount_folders:` with the `docker` hypervisor will encounter a
test-halting `Docker::Error::ClientError`.  This is a problem because
it forces docker-based nodesets to specify host-side mounts with
absolute paths, which isn't portable.

This patch fixes the issue by using Ruby's `File.expand_path()` method to
translate any relative `mount_folders:`/`from:` entries into absolute
paths before passing them on to docker.


electrical:
Good catch.

kevpl:

PR #1066, opened 2016-02-19
(BKR-705) Change Cisco platform names

LuvCurves:


lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -845,6 +845,11 @@ def install_puppetlabs_release_repo_on( hosts, repo = nil, opts = options )
             when /^(fedora|el|centos|sles|cisco)$/
               variant_url_value = (($1 == 'centos') ? 'el' : $1)
               variant_url_value = 'cisco-wrlinux' if variant == 'cisco'
+              if variant == 'cisco'
+                version = '5' if version == 'nexus'
+              else
+                version = '7' if version == 'ios_xr'
```
chrisvanheuveln: I don't think you want the `else` here - you won't get to the `ios_xr` check with that in place.

spec/beaker/host/cisco_spec.rb
```diff
@@ -14,10 +14,10 @@ module Cisco
 
     describe '#prepend_commands' do
 
-      context 'for cisco-5' do
+      context 'for cisco-nexus' do
 
         before :each do
-          @platform = 'cisco-5-x86'
+          @platform = 'cisco-nexus-x86'
```
glennmatthews: This seems wrong still - we still need the ability to specify an OS version in case there are behavior differences between, say, NX-OS 7.0 and NX-OS 7.1. Rather than treating `nexus` or `ios_xr` as the 'version', they should be part of the 'os' - so something more like `cisco_nexus-7.0(3)I2(1)-x86_64` or `cisco_ios_xr-6.0.0-x86_64` is what I would expect Beaker to accept.

docs/hosts/cisco.md
```diff
@@ -7,7 +7,7 @@ can get more details on this from their
 Wind River Linux support came to Beaker through testing Cisco network
```
glennmatthews: This sentence gives me pause, TBH - a lot of the logic below is specific to Cisco's OS and is probably not applicable to WRL in general. Does Beaker really support arbitrary WRL distros?

docs/hosts/cisco.md
```diff
@@ -25,7 +25,7 @@ file line that looks like this in our configuration:
         ...
         vrf: management
     
-B. WRLinux NXOS (5) hosts will also require a user to be set on the
+B. Cisco Nexus hosts will also require a user to be set on the
```
glennmatthews: This requirement 'B' is true for both NX-OS and IOS XR. Requirement A above only applies to NX-OS (IOS XR *can* use a VRF but doesn't have to)

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -890,7 +895,7 @@ def install_repo_configs(host, buildserver_url, package_name, build_version, cop
                                   repo_filename,
                                   copy_dir )
 
-          if host[:platform] =~ /cisco-5/
+          if host[:platform] =~ /cisco-nexus/
             to_path = "#{host.package_config_dir}/#{File.basename(repo)}"
```
glennmatthews: Can you add a comment explaining what this is doing on NX-OS and why? Are you confident that it doesn't apply to IOS XR?

lib/beaker/host/cisco.rb
```diff
@@ -21,7 +21,7 @@ def skip_set_env?
     #
     # @return nil
     def scp_post_operations(scp_file_actual, scp_file_target)
-      if self[:platform] =~ /cisco-5/
+      if self[:platform] =~ /cisco-nexus/
```
glennmatthews: Not sure what's gained specifically by gating this for Nexus. Why not instead `if scp_file_actual != scp_file_target`?

lib/beaker/host/cisco.rb
```diff
@@ -72,7 +72,7 @@ def prepend_commands(command = '', user_pc = '', opts = {})
       return user_pc unless command.index('vsh').nil?
 
       prepend_cmds = 'source /etc/profile;'
-      if self[:platform] =~ /cisco-5/
+      if self[:platform] =~ /cisco-nexus/
```
glennmatthews: This should probably instead be `if self[:vrf]` - use the VRF if defined, don't use it if not.

lib/beaker/host/cisco.rb
```diff
@@ -95,7 +95,7 @@ def environment_string env
       env_array = self.environment_variable_string_pair_array( env )
       environment_string = env_array.join(' ')
 
-      command = self[:platform] =~ /cisco-5/ ? 'export' : 'env'
+      command = self[:platform] =~ /cisco-nexus/ ? 'export' : 'env'
```
glennmatthews: What's the reason for this?

lib/beaker/host/cisco.rb
```diff
@@ -108,8 +108,8 @@ def validate_setup
       msg = nil
       msg = 'Cisco hosts must be provided with a :vrf value.' unless self[:vrf]
```
glennmatthews: vrf is only mandatory for Nexus.

lib/beaker/host/cisco.rb
```diff
@@ -108,8 +108,8 @@ def validate_setup
...
 
-      if !msg && self[:platform] =~ /cisco-5/ && self[:user] == 'root'
-        msg = 'Cisco-5 hosts must be provided with a :user value, as they can not SSH in as root.'
+      if !msg && self[:platform] =~ /cisco-nexus/ && self[:user] == 'root'
```
glennmatthews: root ssh is blocked on both Nexus and IOS XR.

spec/beaker/host/cisco_spec.rb
chrisvanheuveln: @glennmatthews Won't the prepend commands be the same for nexus models/versions though? If this changes to include version names then you'll need separate context for n9k = 7.0(3)I2(1), 7.0.(3)I2(2), n7k = 7.3(0)D1, n5/6k = 6.0..., etc, which then needs to be updated with every release?

glennmatthews: Well, at this point we don't need different commands for a particular version - we can treat all 'cisco_nexus-XXXXX' as equivalent. I'm just trying to think forward-compatibility here in case something does change.

chrisvanheuveln: Yup, agreed.

lib/beaker/dsl/install_utils/foss_utils.rb
LuvCurves: I'll remove the if.  The versions are used by beaker to get the correct version of the package.  Our package names are 'cisco-wrlinux-<version>-noarch.rpm'  So we need that to install the package correctly.

spec/beaker/host/cisco_spec.rb
LuvCurves: How about we cross that bridge when we get there.

docs/hosts/cisco.md
LuvCurves: We can take the WRL reference out.

docs/hosts/cisco.md
LuvCurves: Ok.

lib/beaker/dsl/install_utils/foss_utils.rb
LuvCurves: It's getting the correct file from our repo.  I am confident, because we have tested this.

lib/beaker/host/cisco.rb
LuvCurves: Ok, we'll remove it.

lib/beaker/host/cisco.rb
LuvCurves: We have test that will set the environment prior to executing a command. On Nexus, we found the using the 'env' command didn't work for this, we had to use the export command.

lib/beaker/dsl/install_utils/foss_utils.rb
glennmatthews: Interesting - I hadn't known Beaker had support for installing the puppet agent. Does this make https://github.com/cisco/cisco-network-puppet-module/blob/develop/utilities/installer/install_puppet.rb somewhat redundant now?

spec/beaker/host/cisco_spec.rb
glennmatthews: Just trying to avoid another breaking change in the future (`cisco-nexus-x86_64` ==> `cisco_nexus-VERSION-x86_64`). If that's not a concern from Puppet's side of things, fine by me.

lib/beaker/dsl/install_utils/foss_utils.rb
glennmatthews: Let me rephrase. :-) Why does nexus need a different destination path than any other platform?

lib/beaker/host/cisco.rb
LuvCurves: The only access to an IOS XR platform we have available to us is a VM that one of our release engineers was able to get running in our VMpooler.  That particular VM lets us use root.  I'll change this to match the Nexus.

lib/beaker/dsl/install_utils/foss_utils.rb
LuvCurves: Yes, it makes your install_puppet.rb look redundant.

spec/beaker/host/cisco_spec.rb
LuvCurves: Let me discuss this with the primary developer of Beaker.

lib/beaker/host/cisco.rb
LuvCurves: @glennmatthews We have a IOS XR VM that we are using for testing.  There does not appear to be a need to use the 'ip netns exec' to get things that require networking to work.  There is also no 'sudo' installed.  Is this normally available on IOS XR systems?  The only thing we've need to do for our own Puppet test is make sure puppet binaries are in the PATH.

lib/beaker/dsl/install_utils/foss_utils.rb
LuvCurves: We found that we could not SCP to the directory on the Nexus, we had to specify the target filename.  That is not required by any other platform.

lib/beaker/host/cisco.rb
LuvCurves: Same as above: "We found that we could not SCP to the directory on the Nexus, we had to specify the target filename.  That is not required by any other platform."

spec/beaker/host/cisco_spec.rb
LuvCurves: We're going with cisco_nexus-MAJOR_VERSION-x86_64 & cisco_ios_xr-MAJOR_VERSION-x86_64.  Which is currently cisco_nexus-7-x86_64 & cisco_ios_xr-6-x86_64

lib/beaker/host/cisco.rb
glennmatthews: I think you have an older 'beta' IOS XR VM at present. There've been a number of enhancements and recommendations for the workflow that have been made in the last few months. I'll make sure we cover this in our next interlock meeting with PuppetLabs. Thanks for the details - these will be helpful.

spec/beaker/host/cisco_spec.rb
glennmatthews: cisco_nexus-7-x86_64 and cisco_ios_xr-6-x86_64 sounds good to me.

lib/beaker/platform.rb
```diff
@@ -3,7 +3,7 @@ module Beaker
   # all String methods while adding several platform-specific use cases.
   class Platform < String
     # Supported platforms
-    PLATFORMS = /^(cisco|(free|open)bsd|osx|centos|fedora|debian|oracle|redhat|scientific|sles|ubuntu|windows|solaris|aix|el|eos|cumulus|f5|netscaler)\-.+\-.+$/
+    PLATFORMS = /^(cisco_nexus|cisco_ios_xr(free|open)bsd|osx|centos|fedora|debian|oracle|redhat|scientific|sles|ubuntu|windows|solaris|aix|el|eos|cumulus|f5|netscaler)\-.+\-.+$/
```
glennmatthews: Looks like you're missing a '|' here  between `cisco_ios_xr` and `(free|open)bsd`.

spec/beaker/host/cisco_spec.rb
```diff
@@ -150,9 +150,9 @@ module Cisco
...
       end
 
-      context 'on the cisco-7 platform' do
+      context 'on the cisco_ios_xr platform' do
         before :each do
-          @platform = 'cisco-7-x86'
+          @platform = 'cisco_ios_xr-x86_64'
         end
 
         it 'errors when no :vrf value is provided' do
```
glennmatthews: This shouldn't be true on IOS XR.

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -842,14 +842,24 @@ def install_puppetlabs_release_repo_on( hosts, repo = nil, opts = options )
...
+              if variant == 'cisco_ios_xr'
+                variant_url_value = 'cisco-wrlinux'
+                version = '7'
+              end
               remote = "%s/puppetlabs-release%s-%s-%s.noarch.rpm" %
                 [opts[:release_yum_repo_url], repo_name, variant_url_value, version]
 
-              if variant == 'cisco'
-                # cisco requires using yum to install the repo
+              if variant == 'cisco_ios_xr'
```
stahnma: Why break these apart if they are the same set of actions?

lib/beaker/host.rb
```diff
@@ -57,8 +57,10 @@ def self.create name, host_hash, options
         FreeBSD::Host.new name, host_hash, options
       when /eos/
         Eos::Host.new name, host_hash, options
-      when /cisco/
+      when /cisco_nexus/
```
stahnma: Since cisco_nexus and cisco_exr both match` /cisco/` this, why break it out?

stahnma:
I don't understand most of this PR. In nearly all cases `/cisco/` matches both platform names and performs identical actions. Why break them out? It's just more code to maintain. I understand updating the README or user-facing documentation, but internally to the code, just keep `/cisco/` as the matcher unless there a specific difference across them both. 

We treat EL as /el/ even though that could be RHEL 4, Scientific 6 x86_64 Centos 7 or whatever. It seems like Cisco is a similar pattern.

lib/beaker/dsl/install_utils/foss_utils.rb
LuvCurves: Because on cisco_ios_xr, we have to do a localinstall of the repo because there are no repos configured at that moment on the VM.

LuvCurves:
@stahnma That is what we have tried to do, however, as @kevpl can attest, there are enough differences that we can't keep them the same everywhere.  I have removed one of the case that you were referring to where it was clear /cisco/ was adequate.

stahnma:
Thanks @LuvCurves . :+1:  on the idea. I assume tests need to pass.

LuvCurves:
Due to the amount of drift that has happened since I first pulled the master branch. I will close this PR and open one that is up to date with master.

PR #1070, opened 2016-02-23
(BKR-671) - Add error handing for Beaker failing to acquire a VM from vmpooler

pinkypie:
The issue is when one of the specified host cannot be acquired.
Raise an error and attempt again until timeout (the same as with other provisioning errors).

### DIFF
--- a/lib/beaker/hypervisor/vmpooler.rb
+++ b/lib/beaker/hypervisor/vmpooler.rb
@@ -127,6 +127,10 @@ module Beaker
           domain = parsed_response['domain']
 
           @hosts.each_with_index do |h, i|
+            # If one of the requested hosts is not available on vmpooler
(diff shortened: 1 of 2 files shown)


kevpl:
@pinkypie please add a spec test so that we don't break this in the future

pinkypie:
@kevpl trying to. Having problems with getting the method to return an error in the spec test.

rick:
It could be that wrapping some part of the `parsed_response[h['template']]` bit in a helper method would (a) give us a more intention-revealing name, and (b) make this easier to test.

pinkypie:
Thanks @rick , could also probably do that for some of the other error handling logic in the provision method, and then add some more tests. Will make a note.

kevpl:
@rick look good?

rick:
:+1:

PR #1073, opened 2016-02-29
(BKR-308) moved PE-specific functionality to beaker-pe

kevpl:
beaker-pe refs:

- [github repo](https://github.com/puppetlabs/beaker-pe)
- [rubygems location](https://rubygems.org/gems/beaker-pe)


kevpl:

kevpl:

lib/beaker/dsl.rb
```diff
@@ -79,5 +80,10 @@ module DSL
     include Beaker::DSL::Helpers
     include Beaker::DSL::InstallUtils
     include Beaker::DSL::Patterns
+    include Beaker::DSL::PE
+
+    def self.register(helper_module)
```
objectverbobject: are we actually registering modules with this? I remember exploring this a while back, and found that the register method wasn't going to work out real well...although I can't remember specifics.

rick: I believe this came in as part of the work to support alternate test runners as part of Beaker.next (or some similar such name).

I don't know of a downside to the approach (though a question is whether it's in use by anything already or should have been rolled back when we rolled back an earlier PR). 

The downside to the current approach widely in use is that various modules directly open and reference the beaker class hierarchy and muck around with the internals. To eventually have sane libraries and to support things like alternate test runners we're likely to need to move away from that practice. Providing a registration hook is one way to do that.

I can talk more at some length about the bigger picture thinking (there's a mikado graph around here somewhere too) that motivates this.

objectverbobject: I think that fixing the current approach—or at least not encouraging directly opening and referencing the beaker class hierarchy—is definitely something we should be doing. However, this `register` method basically just seems to be sidestepping ruby's privatization of `include`, not necessarily stopping people from mucking with the internals. 

Perhaps I'm just not getting it; @kevpl is this in use anywhere yet in any of the beaker libraries, so I could see it in use and get a better grasp of this?

lib/beaker/dsl/install_utils.rb
```diff
@@ -1,13 +1,14 @@
...
 end
+require "beaker/dsl/install_utils/pe_defaults"
 
 module Beaker
   module DSL
     # Collection of installation methods and support
     module InstallUtils
       include DSL::InstallUtils::PuppetUtils
-      include DSL::InstallUtils::PEUtils
+      include DSL::InstallUtils::PEDefaults
```
objectverbobject: Is there a reason why this wasn't moved to beaker-pe? That seems like the more logical location for this set of info.

lib/beaker/dsl.rb
objectverbobject: Alright, I've tested this out now and feel pretty good about it after some hemming and hawing; I look forward to expanding this register function!

objectverbobject:
@kevpl Outside of my comment about `PEDefaults` module, I think this code is good to merge. That also doesn't HAVE to be addressed in this PR either...

lib/beaker/dsl/install_utils.rb
kevpl: this is left in due to backwards compatibility.  It was tried to pull this out previously, but it got left in because tests wouldn't work without the paths being set properly on SUT setup.  This is still the case, and I've decided to leave this at this point rather than try to solve this issue at this time.

objectverbobject:
LGTM.

I meant to do it

PR #1076, opened 2016-03-02
(BKR-715) Use puppet_agent_version for MSI from dev repos

MikaelSmith:
If a `puppet_agent_version` is specified, use it in
`install_puppet_agent_dev_repo_on` to fetch the MSI, so we get an MSI
unique to that SHA. This provides a work-around for BKR-712, and
normalizes behavior with `install_puppet_agent_on`.

### DIFF
--- a/lib/beaker/host/windows/pkg.rb
+++ b/lib/beaker/host/windows/pkg.rb
@@ -65,7 +65,13 @@ module Windows::Pkg
     # - we do not have install_32 set on host
     # - we do not have install_32 set globally
     arch_suffix = should_install_64bit ? '64' : '86'
-    release_file = "puppet-agent-x#{arch_suffix}.msi"
+    # If a version was specified, use it; otherwise fall back to a default name.
+    # Avoid when puppet_agent_version is set to a SHA, which isn't used in package names.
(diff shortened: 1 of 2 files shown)


lib/beaker/host/windows/pkg.rb
```diff
@@ -65,7 +65,11 @@ def puppet_agent_dev_package_info( puppet_collection = nil, puppet_agent_version
     # - we do not have install_32 set on host
     # - we do not have install_32 set globally
     arch_suffix = should_install_64bit ? '64' : '86'
-    release_file = "puppet-agent-x#{arch_suffix}.msi"
+    if puppet_agent_version
+      release_file = "puppet-agent-#{puppet_agent_version}-x#{arch_suffix}.msi"
```
MikaelSmith: I'm not sure if a similar change is needed for `pe_puppet_agent_promoted_package_info` below.

MikaelSmith:
This breaks when we set `puppet_agent_version` to `opt[:version]`.

MikaelSmith:
Updated to fix specs and ignore if puppet_agent_version is a SHA.

PR #1080, opened 2016-03-11
(maint) Correct install_puppet_agent_on args

MikaelSmith:
The docs for `install_puppet_agent_on` specify that the `opts` arg is
optional, but it doesn't have a default value so in practice it's
required. Update the call signature to specify a default empty hash, so
the function matches its description.


kevpl:
for tracking purposes, this was filed as a [BKR ticket](https://tickets.puppetlabs.com/browse/BKR-315).

PR #1081, opened 2016-03-15
(BKR-722) Don't clobber user_pc in prepend_commands for cisco host

glennmatthews:
The implementation of `#prepend_commands` for cisco hosts returns `'source /etc/profile;'` and optionally `"sudo ip netns exec #{vrf}"` but does not retain any user-specified prepends. Additionally, it supplies `'sudo'` always when the vrf is specified (regardless of whether the user is root) and never supplies `'sudo'` when the vrf is not specified.
(edit) Additionally, because 'sudo' is inserted in `prepend_commands`, any values from `environment_string` are lost when switching users

Fix 1: preserve user_pc.
Fix 2: prepend 'sudo' depending on whether user is root, independent of whether vrf is present.
(edit) Fix 3: Add `'source /etc/profile/; [sudo]'` in `environment_string` rather than in `prepend_commands`.

```
$ rspec spec/beaker/host/cisco_spec.rb 

Cisco::Host
  #prepend_commands
    for cisco_nexus-7
      ends with the :vrf host parameter
      guards against "vsh" usage (only scenario we dont want prefixing)
      retains user-specified prepend commands when adding vrf
    for cisco_ios_xr-6
      does use the :vrf host parameter if provided
      retains user-specified prepend commands when adding vrf
  #environment_string
    starts with sourcing the /etc/profile script
    uses `sudo` if not root
    for cisco_nexus-7
      uses `sudo` if not root
      uses `export` if root
      ends with a semi-colon
      turns env maps into paired strings correctly
    for cisco_ios_xr-6
      uses `sudo` if not root
      uses `env` if root
      does not end with a semi-colon
      turns env maps into paired strings correctly
  #package_config_dir
    returns correctly for cisco platforms
  #repo_type
    returns correctly for cisco platforms
  #validate_setup
    on the cisco_nexus-7 platform
      errors when no :vrf value is provided
      errors when no :user value is provided
      does nothing if the host is setup correctly
    on the cisco_ios_xr-6 platform
      does nothing if no :vrf value is provided
      errors when no user is provided
      does nothing if the host is setup correctly

Finished in 0.09871 seconds (files took 1.13 seconds to load)
23 examples, 0 failures
```

### DIFF
--- a/lib/beaker/host/cisco.rb
+++ b/lib/beaker/host/cisco.rb
@@ -71,11 +71,14 @@ module Cisco
     def prepend_commands(command = '', user_pc = '', opts = {})
       return user_pc unless command.index('vsh').nil?
 
-      prepend_cmds = 'source /etc/profile;'
+      prepend_cmds = ''
       if self[:vrf]
-        prepend_cmds << "sudo ip netns exec #{self[:vrf]}"
+        prepend_cmds << "ip netns exec #{self[:vrf]} "
       end
-      return prepend_cmds
+      if user_pc && !user_pc.empty?
+        prepend_cmds << "#{user_pc} "
+      end
+      prepend_cmds.strip
     end
 
     # Construct the environment string for this command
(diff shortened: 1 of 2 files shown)


glennmatthews:
@cisco/cisco-network-node-project-committers FYI

LuvCurves:
:+1:  I've been using this to test with over the last 2 days and it works just fine for our needs.

tvpartytonight:
:+1:

glennmatthews:
Thanks!

PR #1082, opened 2016-03-17
(BKR-724) Run docker_image_commands sooner

mcanevet:


electrical:
Any specific reasoning for this?

mcanevet:
@electrical we still validate our code on squeeze (I know it's EOL but you don't always do what you want), and official `debian:squeeze` image didn't change the sources.list to point to archive.debian.org, so I have to fix it before the first `apt-get update`.

electrical:
Ahh okay. i see. sounds reasonable to me :-)

electrical:
:+1: from me if it wasn't clear.

tvpartytonight:
Hi @mcanevet ! Thanks for these commits, we look forward to getting them into the project.

@electrical as the community maintainer of the docker hypervisor, do you think that this change requires an additional spec test so we don't break this functionality in the future? In general we would like all non `maint` PR's to include spec tests verifying the behavior intended in the PR. I'd like to see a spec test for this, but I am open to hearing about why it may not be necessary.

electrical:
I agree it would be nice to have a spec test in there to ensure we don't break it by accident in the future.

tvpartytonight:
@mcanevet based on @electrical 's comment, I'd like to see a spec test added to this PR before merging. Also, if you could add a short description in your commit, that would make this PR align with the guidelines in the [contributing](https://github.com/puppetlabs/beaker/blob/master/CONTRIBUTING.md) doc.

Please feel free to ping me if you have any other questions. Thanks!

mcanevet:
@tvpartytonight I don't need this anymore.

PR #1083, opened 2016-03-22
(maint) change junit skip to skipped to match schema

samwoods1:
Noticed that Jenkins isn't picking up skipped status, so I updated the junit output based on schema at - https://svn.jenkins-ci.org/trunk/hudson/dtkit/dtkit-format/dtkit-junit-model/src/main/resources/com/thalesgroup/dtkit/junit/model/xsd/junit-7.xsd


pinkypie:
:+1:

PR #1084, opened 2016-03-23
(BKR-750) Add pre-cleanup option

highb:
Adds a `--pre-cleanup` suite, that will *always* run before the
VM cleanup step. Useful for saving logs before the VMs disappear.


lib/beaker/cli.rb
```diff
@@ -113,6 +113,15 @@ def execute!
         end
       #cleanup phase
       rescue => e
+        begin
+          if @options[:fail_mode].to_s =~ /slow/
```
highb: Wasn't sure if the pre-cleanup should ignore the slow fail mode or not. Maybe it should, since I'm billing that it will "always run"?

tvpartytonight:
@highb thanks for submitting this; I want to wait until @kevpl has time to review this also before it gets merged in. Thanks again!

lib/beaker/cli.rb
kevpl: @highb yeah, I think you're right, that we should keep with the definition that it should always run, so it shouldn't check for slow mode like the post suite does. Other than that, everything looks good here. Once that change is through, we'll run it through one last automation pass & we should be good to merge this.

highb:
Updated to ignore the fail_mode, but I just noticed the specs are failing so I'm going to fix that up.

highb:
There we go, specs fixed. @kevpl this is ready for review again when you have time.

highb:
jenkins retest this please

kevpl:

PR #1085, opened 2016-03-30
(BKR-749) Wrap Nexus Pre-commands

LuvCurves:


tvpartytonight:
👍

glennmatthews:
This broke IOS XR. :frowning: 

```
    Beaker::Host::CommandFailure: Host 'sunstone-gfm.cisco.com' exited with 1 running:
     source /etc/profile; sudo sh -c " /opt/puppetlabs/bin/puppet config print certname
    Last 10 lines of output were:
        sh: -c: line 0: unexpected EOF while looking for matching `"'
        sh: -c: line 1: syntax error: unexpected end of file
```

lib/beaker/command.rb
```diff
@@ -72,9 +72,13 @@ def initialize command, args = [], options = {}
     def cmd_line host, cmd = @command, env = @environment, pc = @prepend_cmds
       env_string = host.environment_string( env )
       prepend_commands = host.prepend_commands( cmd, pc, :cmd_exe => @cmdexe )
+      if host[:platform] =~ /cisco_nexus/ && host[:user] != 'root'
```
glennmatthews: Doesn't this go in the opposite direction of #1051 / #1062 by putting platform-specific logic back into this module?

glennmatthews:
@mikewiebe @chburket FYI.

lib/beaker/command.rb
```diff
@@ -72,9 +72,13 @@ def initialize command, args = [], options = {}
...
+          append_command = '"'
+        cmd = cmd.gsub('"') { '\\"' }
```
glennmatthews: This doesn't fix the case where we have `"` in options_string or args_string...

LuvCurves:
@glennmatthews Our only IOS XR test platform here is a VM that allows us to run everything as root.  Even if I make the changes here, I don't have anything to test it against.  Please submit a PR, like you did with user_pc clobber.

glennmatthews:
@LuvCurves See #1087

PR #1086, opened 2016-03-30
(BKR-430) Adding rake tasks for running beaker acceptance tests.

pinkypie:
'rake acceptance' for the main acceptance run.
Also the following individual tasks for individual sets of tests.
rake test:base
rake test:hypervisor
rake test:puppetpe
rake test:puppetgem
rake test:puppetgit
rake test:puppetpkg

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,4 +1,8 @@
 require 'open3'
+require 'securerandom'
+require 'beaker-hostgenerator'
+
+HOSTS_PRESERVED  = 'log/latest/hosts_preserved.yml'
 
 task :default => [ 'test:spec' ]
 
@@ -9,6 +13,10 @@ end
 task :spec do
   Rake::Task['test:spec'].invoke
 end
+
+
+task :acceptance => ['test:base', 'test:puppetgit', 'test:hypervisor']
+
 
 task :yard do
   Rake::Task['docs:gen'].invoke
@@ -23,7 +31,118 @@ task :travis do
   Rake::Task['spec'].invoke
 end
 
+module HarnessOptions
+  defaults = {
+      :tests  => ['tests'],
+      :log_level => 'debug',
+      :preserve_hosts => 'onfail',
+  }
+
+  DEFAULTS = defaults
+
+  def self.get_options(file_path)
+    puts "Attempting to merge config file: #{file_path}"
+    if File.exists? file_path
+      options = eval(File.read(file_path), binding)
+    else
+      puts "No options file found at #{File.expand_path(file_path)}... skipping"
+    end
+    options || {}
+  end
+
+  def self.get_mode_options(mode)
+    get_options("./acceptance/config/#{mode}/acceptance-options.rb")
+  end
+
+  def self.get_local_options
+    get_options('./acceptance/local_options.rb')
+  end
+
+  def self.final_options(mode, intermediary_options = {})
+    mode_options = get_mode_options(mode)
+    local_overrides = get_local_options
+    final_options = DEFAULTS.merge(mode_options)
+    final_options.merge!(intermediary_options)
+    final_options.merge!(local_overrides)
+  end
+
+end
+
+def hosts_file_env
+  ENV['BEAKER_HOSTS']
+end
+
+def hosts_opt(use_preserved_hosts=false)
+  if use_preserved_hosts
+    "--hosts=#{HOSTS_PRESERVED}"
+  else
+    if hosts_file_env
+      "--hosts=#{hosts_file_env}"
+    else
+      "--hosts=tmp/#{HOSTS_FILE}"
+    end
+  end
+end
+
+def agent_target
+  ENV['TEST_TARGET'] || 'redhat7-64af'
+end
+
+def master_target
+  ENV['MASTER_TEST_TARGET'] || 'redhat7-64default.mdcal'
+end
+
+def test_targets
+  ENV['LAYOUT'] || "#{master_target}-#{agent_target}"
+end
+
+HOSTS_FILE = "#{test_targets}-#{SecureRandom.uuid}.yaml"
+
+def beaker_test(mode = :base, options = {})
+
+  preserved_hosts_mode = options[:hosts] == HOSTS_PRESERVED
+  final_options = HarnessOptions.final_options(mode, options)
+
+  options_opt  = ""
+  # preserved hosts can not be used with an options file (BKR-670)
+  #   one can still use OPTIONS
+
+  if !preserved_hosts_mode
+    options_file = 'merged_options.rb'
+    options_opt  = "--options-file=#{options_file}"
+    File.open(options_file, 'w') do |merged|
+      merged.puts <<-EOS
+# Copy this file to local_options.rb and adjust as needed if you wish to run
+# with some local overrides.
+      EOS
+      merged.puts(final_options)
+    end
+  end
+
+  tests = ENV['TESTS'] || ENV['TEST']
+  tests_opt = ""
+  tests_opt = "--tests=#{tests}" if tests
+
+  overriding_options = ENV['OPTIONS'].to_s
+
+  args = [options_opt, hosts_opt(preserved_hosts_mode), tests_opt,
+          *overriding_options.split(' ')].compact
+
+  sh("beaker", *args)
+end
+
+
 namespace :test do
+  USAGE = <<-EOS
+You may set BEAKER_HOSTS=config/nodes/foo.yaml or include it in an acceptance-options.rb for Beaker,
+or specify TEST_TARGET in a form beaker-hostgenerator accepts, e.g. ubuntu1504-64a.
+You may override the default master test target by specifying MASTER_TEST_TARGET.
+You may set TESTS=path/to/test,and/more/tests.
+You may set additional Beaker OPTIONS='--more --options'
+If there is a Beaker options hash in a ./acceptance/local_options.rb, it will be included.
+Commandline options set through the above environment variables will override settings in this file.
+  EOS
+
   desc 'Run specs and check for deprecation warnings'
   task :spec do
     original_dir = Dir.pwd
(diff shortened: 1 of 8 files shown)


Rakefile
```diff
@@ -10,6 +14,12 @@ task :spec do
   Rake::Task['test:spec'].invoke
 end
 
+task :acceptance do
+  Rake::Task['test:base'].invoke
```
rick: Wondering if this could just be done by listing these as dependencies on the task, rather than explicit `invoke` calls?

Rakefile
```diff
@@ -23,7 +33,116 @@ task :travis do
...
+    if File.exists? file_path
+      options = eval(File.read(file_path), binding)
+    else
+      puts "No options file found at #{File.expand_path(file_path)}... skipping"
+    end
+    options || {}
+  end
+
+  def self.get_local_options
+    get_options("./local_options.rb")
```
rick: This is probably dependent on the current working directory.  Probably want an incantation that involves `__FILE__`, maybe `expand_path`, etc., to make sure this works from wherever rake might be run.

Rakefile
```diff
@@ -23,7 +33,116 @@ task :travis do
...
+  end
+
+  def self.final_options(intermediary_options = {})
+    local_overrides = get_local_options
+    final_options = DEFAULTS.merge(intermediary_options)
+    final_options.merge!(local_overrides)
+    return final_options
```
rick: Nitpick, but trailing `return`s aren't necessary in ruby -- last expression will be the return value from the method.

Rakefile
```diff
@@ -23,7 +33,116 @@ task :travis do
...
+
+def master_target
+  ENV['MASTER_TEST_TARGET'] || 'redhat7-64default.mdcal'
+end
+
+def test_targets
+  ENV['LAYOUT'] || "#{master_target}-#{agent_target}"
+end
+
+HOSTS_FILE = "#{test_targets}-#{SecureRandom.uuid}.yaml"
```
rick: Do we need to require a lib to get access to `SecureRandom`?  I might be missing how we're getting access to this.

Rakefile
```diff
@@ -1,4 +1,8 @@
 require 'open3'
+require 'pp'
```
rick: This doesn't appear to be in use any longer, could probably be dropped.

Rakefile
```diff
@@ -23,7 +33,116 @@ task :travis do
...
+end
+
+def test_targets
+  ENV['LAYOUT'] || "#{master_target}-#{agent_target}"
+end
+
+HOSTS_FILE = "#{test_targets}-#{SecureRandom.uuid}.yaml"
+
+def pe_dir
+  ENV['PE_DIR'] || "http://neptune.puppetlabs.lan/archives/releases/3.7.2"
```
rick: This seems fairly internal to add to the general beaker project. Wondering if there's a way we can specify our internal defaults in an out-of-band (env variables, etc.) way so we don't end up publishing these things.  This is also somewhat fragile even for internal use -- if things move around we have to ship a new release of beaker.

Rakefile
```diff
@@ -23,7 +33,116 @@ task :travis do
...
+  options_opt  = ""
+  # preserved hosts can not be used with an options file (BKR-670)
+  #   one can still use OPTIONS
+
+  if !preserved_hosts_mode
+    options_file = 'merged_options.rb'
+    options_opt  = "--options-file=#{options_file}"
+    File.open(options_file, 'w') do |merged|
+      merged.puts <<-EOS
+      EOS
```
rick: Is this just a newline?  Maybe the body of the heredoc is missing?

Rakefile
```diff
@@ -23,7 +33,116 @@ task :travis do
...
+      merged.puts(final_options.pretty_inspect)
+    end
+  end
+
+  tests = ENV['TESTS'] || ENV['TEST']
+  tests_opt = ""
+  if tests
+    tests_opt = "--tests=#{tests}"
```
rick: Probably better as an inline `if` unless something else is coming here.

Rakefile
er0ck: yep.  could be just 
`task acceptance: :test:base`

Rakefile
```diff
@@ -23,7 +33,116 @@ task :travis do
...
+    options_file = 'merged_options.rb'
+    options_opt  = "--options-file=#{options_file}"
+    File.open(options_file, 'w') do |merged|
+      merged.puts <<-EOS
+      EOS
+      merged.puts(final_options.pretty_inspect)
+    end
+  end
+
+  tests = ENV['TESTS'] || ENV['TEST']
```
er0ck: should this have a default, like acceptance/tests or ./tests (with expand_path magic done for us)?

Rakefile
```diff
@@ -23,7 +33,116 @@ task :travis do
...
+
+
 namespace :test do
+  USAGE = <<-EOS
+You may set HOSTS=config/nodes/foo.yaml or include it in an options.rb for Beaker,
+or specify TEST_TARGET in a form beaker-hostgenerator accepts, e.g. ubuntu1504-64a.
+You may override the default master test target by specifying MASTER_TEST_TARGET.
+You may set TESTS=path/to/test,and/more/tests.
+You may set additional Beaker OPTIONS='--more --options'
+If there is a Beaker options hash in a ./local_options.rb, it will be included.  Commandline options set through the above environment variables will override settings in this file.
```
er0ck: nit: long line

Rakefile
```diff
@@ -48,8 +167,69 @@ namespace :test do
...
+  task :hypervisor  => 'gen_hosts' do
+    beaker_test({:tests => 'acceptance/tests/hypervisor'})
+  end
+
+  desc <<-EOS
+Run the puppet beaker acceptance tests on a pe install.
+#{USAGE}
+  EOS
+  task :puppetpe  => 'gen_hosts' do
+    beaker_test({:pre_suite => 'acceptance/pre_suite/pe/install.rb', :tests => 'acceptance/tests/puppet'})
```
er0ck: specifying the pre_suite here *might* break preserved_hosts_mode, but maybe that's ok

Rakefile
```diff
@@ -48,8 +167,69 @@ namespace :test do
...
+  task :puppetpkg => 'gen_hosts' do
+    beaker_test({:pre_suite => 'acceptance/pre_suite/puppet_pkg/install.rb', :tests => 'acceptance/tests/puppet'})
+  end
+
+  desc 'Generate Beaker Host Config File'
+  task :gen_hosts do
+    if hosts_file_env
+      next
+    end
+    cli = BeakerHostGenerator::CLI.new([test_targets, "--pe_dir=#{pe_dir}"])
```
er0ck: should this --pe-dir flag be conditional?  it's being specified for all test types, it appears.

Rakefile
pinkypie: It is part of Ruby's Standard Lib.

Rakefile
pinkypie: This is actually relative to the rakefile itself, so it shouldn't matter where rake is called from. I've tested this by running rake from different directories. The local_options.rb should be at the same level as the Rakefile. I will update the task usage to make it clearer.

Rakefile
pinkypie: This was based on what's currently used in the Jenkins runs for this pre-suite/test. It is difficult to know what to include as defaults (we are also using beaker-hostgenerator if no hosts file is defined and by default it sets hypervisor to vmpooler) as the ticket was originally raised by a Puppet dev who was having to go and hunt through Jenkins jobs and amend them/run them to check his local changes to Beaker. To make it easy for internal people to just run 'rake acceptance' vs community members having to set extra variables/ include extra files is the dilemma.  We could just remove the puppetpe task altogether and have puppetgit (puppet git install) as part of the acceptance run.

Rakefile
pinkypie: @er0ck  - it seems that because of BKR-670, the options we have written to merged_options.rb are ignored when running in preserved_hosts_mode, so the :pre-suite we have defined here is ignored anyway when a hosts file is specified. Would have to set ENV['TESTS'] and/or ENV['OPTIONS'] for any additional options. Unless I am missing an easier/better way to do this.

er0ck: ah, yes.  i forgot about that issue when using merged_options.
i think in one of my project Rakefiles, when running in preserved_hosts mode i ensure we don't use the options file.  In that same project we use the options file(s) to set the presuite.
https://github.com/puppetlabs/orchestrator/blob/2016.2.x/acceptance/Rakefile#L71

will that fix the issue here?

Rakefile
rick: Hm.  I'm confused. I'm seeing this locally:

```
Mon Apr 04 09:54:00 chenrezig ~
 % ▸ irb
>> SecureRandom
NameError: uninitialized constant SecureRandom
	from (irb):1
	from /opt/boxen/rbenv/versions/2.1.5/bin/irb:11:in `<main>'
>> require 'securerandom'
=> true
>> SecureRandom
=> SecureRandom
>> RUBY_VERSION
=> "2.1.5"
```

Rakefile
rick: Yeah, this is a bit thorny and probably deserves more big-picture thought (as a general beaker testing "how do we do the right thing by outside users, and still not leave our internal customers in a confusing place).

I'd be down to highlight this as tech debt (we know it's not the perfect solution, and we promise to have a conversation about it) -- might be something good for ESO CodeCraft or the QE team meeting.

Rakefile
pinkypie: I think rake loads that library at some point, which is why it was working for me. Will add it in for clarity etc..

rick: :ok: :cool: -- yeah, I was just confused on that point.

Rakefile
pinkypie: I removed the assignment in the Rakefile itself and added to the options file specific to the pe test. I added the puppetgit test to the acceptance bundle instead for the meantime. Yes it would be good to think this through more and get others opinions. I'll add a card to both boards and what the consensus is.

Rakefile
pinkypie: I added the specific mode options files and set all the options in there, I think this looks similar to what is being done elsewhere. Let me know if you see anything that looks out of place.

acceptance/config/puppetpkg/acceptance-options.rb
```diff
@@ -0,0 +1,4 @@
+{
+    :pre_suite            => 'acceptance/pre_suite/puppet_pkg/install.rb',
+    :tests                => 'acceptance/tests/puppet'
+}.merge(eval File.read('acceptance/config/acceptance-options.rb'))
```
er0ck: nice.  i think this is a good... option

rick:
I'm :+1:

er0ck:
:shipit:

tvpartytonight:
👍 merging.

PR #1090, opened 2016-04-06
(BKR-680) Add Huaweios support

tvpartytonight:
This PR adds support for the Huaweios platform. The LXC that this support adds is a Debian jessie variant, so the Huaweios platform will just be the same as any other Unix::Host.


LuvCurves:
:+1:

pinkypie:
👍

PR #1091, opened 2016-04-07
(BKR-759) Update dmg url for puppet-agent install

johnduarte:
The url path structure for osx dmg packages has changed. Also,
the `puppet-agent-latest.dmg` alias is no longer generated when
publishing packages to https://downloads.puppetlabs.com.

This commit updates the url structure to reflect the current
pattern. It also attempts to introspect the 'latest' published
version. If a `puppet_agent_version` has not been specified and
a 'latest' version cannot be found, an exception is raised.

### DIFF
--- a/lib/beaker/dsl/install_utils/foss_utils.rb
+++ b/lib/beaker/dsl/install_utils/foss_utils.rb
@@ -696,19 +696,45 @@ module Beaker
             add_role(host, 'aio') #we are installing agent, so we want aio role
 
             variant, version, arch, codename = host['platform'].to_array
-            agent_version = opts[:puppet_agent_version] || 'latest'
-            pkg_name = "puppet-agent-#{agent_version}*"
-            if agent_version == 'latest'
-              dmg_name = "puppet-agent-#{agent_version}.dmg"
-              on host, "curl -O #{opts[:mac_download_url]}/#{dmg_name}"
-            else
-              dmg_name = "puppet-agent-#{agent_version}-osx-#{version}-x86_64.dmg"
-              on host, "curl -O #{opts[:mac_download_url]}/#{opts[:puppet_collection]}/#{dmg_name}"
+
+            download_url = "#{opts[:mac_download_url]}/#{version}/#{opts[:puppet_collection]}/#{arch}"
+
+            latest = get_latest_puppet_agent_build_from_url(download_url)
+
+            agent_version = opts[:puppet_agent_version] || latest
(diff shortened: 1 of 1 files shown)


lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -689,23 +689,38 @@ def install_puppet_from_dmg_on( hosts, opts )
         # @return nil
         # @api private
         def install_puppet_agent_from_dmg_on(hosts, opts)
+          require 'nokogiri'
```
kevpl: @johnduarte have you guys seen [what PE does](https://github.com/puppetlabs/beaker-pe/blob/4fc88d8c83bce01ec9a918d3dfdf7e8496e8fc49/lib/beaker-pe/options/pe_version_scraper.rb) for this?

The latest file is usually just a version for them, are you trying to get additional information from there?

If so, I wonder if potentially having that file be a JSON or ruby hash would be a simpler method of providing these values, rather than parsing HTML.  We'd prefer to limit our exposure to Nokogiri if possible.

johnduarte: @kevpl I was attempting to preserve the internal expectation of support for `latest` as an introspected version as stated by the documentation for this function.

"latest" is not a valid version for published puppet-agent dmg files.

I would also prefer to remove the dependence on Nokogiri. If there is a way to achieve this without changing the API of this method, I am open to suggestions. I am not sure how to get that data without scraping the HTML for the `opts[:mac_download_url]`.

kevpl: The code referenced above doesn't specify that it needs to be the "latest", or use the latest idea that we've developed for PE. You should be able to point at whatever file resource you'd like at any URL, it just matters that the file only contain the version that you'd like to use for that beaker run.

You wouldn't want to use that code anyway because it's for PE, and we don't want FOSS code depending on it. You would just be copying the pattern of how versions are scraped from there.  If you copy it over in a nice general method, then we can actually just make beaker-pe use your method, leaving that method as an adapter if needed.

If this is unclear, maybe we should just get together for a quick face-to-face to get on the same page about what's needed here & what would be the best way to do it.

johnduarte:
@kevpl the nokogiri introspection of 'latest' has been broken out into a separate method as per your suggestion.

kevpl:
thanks @johnduarte!

kevpl:

PR #1092, opened 2016-04-11
WIP (RE-6487) update the aio windows paths.

McdonaldSeanp:
Paths to the "bin" directory for puppet, ruby, facter etc.
have been updated for the AIO agent for windows. With the
new cut over to windows vanagon the path to that directory
will now be /Program Files/Puppet Labs/Puppet/puppet/bin

### DIFF
--- a/lib/beaker/dsl/install_utils/aio_defaults.rb
+++ b/lib/beaker/dsl/install_utils/aio_defaults.rb
@@ -17,13 +17,13 @@ module Beaker
           },
           'windows' => { #windows
             'puppetbindir'      => '/cygdrive/c/Program Files (x86)/Puppet Labs/Puppet/bin:/cygdrive/c/Program Files/Puppet Labs/Puppet/bin',
-            'privatebindir'     => '/cygdrive/c/Program Files (x86)/Puppet Labs/Puppet/sys/ruby/bin:/cygdrive/c/Program Files/Puppet Labs/Puppet/sys/ruby/bin',
+            'privatebindir'     => '/cygdrive/c/Program Files (x86)/Puppet Labs/Puppet/sys/ruby/bin:/cygdrive/c/Program Files/Puppet Labs/Puppet/puppet/bin',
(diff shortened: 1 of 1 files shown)


lib/beaker/dsl/install_utils/aio_defaults.rb
```diff
@@ -17,13 +17,13 @@ module AIODefaults
           },
           'windows' => { #windows
             'puppetbindir'      => '/cygdrive/c/Program Files (x86)/Puppet Labs/Puppet/bin:/cygdrive/c/Program Files/Puppet Labs/Puppet/bin',
-            'privatebindir'     => '/cygdrive/c/Program Files (x86)/Puppet Labs/Puppet/sys/ruby/bin:/cygdrive/c/Program Files/Puppet Labs/Puppet/sys/ruby/bin',
+            'privatebindir'     => '/cygdrive/c/Program Files (x86)/Puppet Labs/Puppet/sys/ruby/bin:/cygdrive/c/Program Files/Puppet Labs/Puppet/puppet/bin',
```
Iristyle: Making this change this will break existing pipelines...

McdonaldSeanp: Yep, that's why I listed this PR as "don't merge". I have put up a card for the vanagon/windows sync up to talk about what to do about the issue.

Iristyle: One thing that you could do is append the path, but I'll have to think a bit about whether or not that's a bad idea.

My preference would be to not have to make any Beaker adjustments, and to keep everything consistent with what we were shipping.  There might be tests scattered about (in other projects) that don't use `privatebindir`, and instead hardcode these paths.  That would then require updating the existing tests in addition to Beaker.

Iristyle: /cc @joshcooper

jcoconnor:
.

Iristyle:
Closing PR as I believe https://tickets.puppetlabs.com/browse/RE-7027 supercedes this effort.

PR #1093, opened 2016-04-12
(BKR-219) Add support for :is_puppetserver per host

tvpartytonight:
This commit adds a check for each host's :is_puppetserver key when using
the `with_puppet_running_on` command. Previously, it would only check
the global option for this key.


kevpl:
:+1:

tvpartytonight:

PR #1098, opened 2016-04-13
(BKR-766) Install puppet agent on huaweios

LuvCurves:


lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -1179,6 +1179,8 @@ def remove_puppet_on( hosts )
             case host[:platform]
             when /cumulus/
               pkgs = on(host, "dpkg-query -l  | awk '{print $2}' | grep -E '(^pe-|puppet)'", :acceptable_exit_codes => [0,1]).stdout.chomp.split(/\n+/)
+            when /huaweios/
+                pkgs = on(host, "dpkg-query -l  | awk '{print $2}' | grep -E '(^pe-|puppet)'", :acceptable_exit_codes => [0,1]).stdout.chomp.split(/\n+/)
```
kevpl: @LuvCurves is there a reason that I'm missing why we need a separate huaweios line here?  It looks to me like this line matches the cumulus one above, so can we just change that expression to the one below?

```ruby
    when /cumulus|huaweios/
```

LuvCurves:
@kevpl Fixed.

PR #1100, opened 2016-04-14
(BKR-767) Only use Solaris major version for pkg

johnduarte:
Puppet only publishes packages under the major version of a Solaris
release. Currently, if Beaker is passed a Solaris host with a minor
version (eg. solaris 11.2), it fails to find and install package.

This commit truncates the Solaris version to its minor component
in the solaris_puppet_agent_dev_package_info and
solaris_install_local_package  methods.


tvpartytonight:
👍

PR #1101, opened 2016-04-15
(BKR-521) handle priority of :acceptable_exit_codes

pinkypie:
Make acceptable_exit_codes override accept_all_exit_codes => true as the more specific of the two options.

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -324,7 +324,8 @@ module Beaker
           # exit codes at the host level and then raising...
           # is it necessary to break execution??
           if options[:accept_all_exit_codes] && options[:acceptable_exit_codes]
-            @logger.warn ":accept_all_exit_codes & :acceptable_exit_codes set. :accept_all_exit_codes overrides, but they shouldn't both be set at once"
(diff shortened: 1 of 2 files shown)


lib/beaker/host.rb
```diff
@@ -324,7 +324,8 @@ def exec command, options={}
           # exit codes at the host level and then raising...
           # is it necessary to break execution??
           if options[:accept_all_exit_codes] && options[:acceptable_exit_codes]
-            @logger.warn ":accept_all_exit_codes & :acceptable_exit_codes set. :accept_all_exit_codes overrides, but they shouldn't both be set at once"
+            @logger.warn ":accept_all_exit_codes & :acceptable_exit_codes set. :acceptable_exit_codes overrides, but they shouldn't both be set at once"
```
er0ck: should this state that we're falling back to `:acceptable_exit_codes`?

er0ck:
otherwise :+1:

lib/beaker/host.rb
pinkypie: I put ':acceptable_exit_codes overrides' instead of what was there already - ':accept_all_exit_codes overrides', but I can reword?

er0ck: oh.  sorry, i guess i missed it.  it could probably use a re-word, but i wouldn't hold the PR up for it

PR #1103, opened 2016-04-18
(BKR-770) Update EPEL package name for EL7

james-stocks:
Change the default EPEL package name for EL7 from "epel-release-7-5.noarch.rpm" to "epel-release-7-6.noarch.rpm"


james-stocks:
Have squashed in a commit that updates the spec tests - is there a way to re-run the Jenkins checks?

MikaelSmith:
😹

PR #1112, opened 2016-04-28
(BKR-789) enable --no-provision for Solaris 11 guests

apollo-io:
Without this change, it is not possible to run beaker against solaris
hosts with --no-provision, which means for every test run the VMs must
be re-provisioned, which has a big impact on the length of time to be
able to run tests during development lifecycle.

For Solaris the 'rolemod' command is used, but this can only be run
successfully once, the second time it runs it fails because the root
user has already been changes from a role to a user.  This change adds a
check to ensure rolemod is only run if it is needed.


kevpl:

PR #1113, opened 2016-04-28
(BKR-758) Fix Bitvise SSH issues and Various vagrant versions not forwarding ssh properly

james-powis:
Bitvise has largely been unable to function due to hard coded linux/cygwin commands, this merge request fixes bitvise ssh windows issues.

Additionally Vagrant has a regression bug where intermittent versions fail to forward ssh when os is windows. Reference: https://github.com/mitchellh/vagrant/issues/7202

### DIFF
--- a/lib/beaker/host_prebuilt_steps.rb
+++ b/lib/beaker/host_prebuilt_steps.rb
@@ -293,7 +293,11 @@ module Beaker
     def get_domain_name(host)
       domain = nil
       search = nil
-      resolv_conf = host.exec(Command.new("cat /etc/resolv.conf")).stdout
+      if ((host['platform'] =~ /windows/) and not host.is_cygwin?)
+        resolv_conf = host.exec(Command.new('type C:\Windows\System32\drivers\etc\hosts')).stdout
+      else
+        resolv_conf = host.exec(Command.new("cat /etc/resolv.conf")).stdout
+      end
       resolv_conf.each_line { |line|
         if line =~ /^\s*domain\s+(\S+)/
           domain = $1
@@ -322,6 +326,8 @@ module Beaker
     def set_etc_hosts(host, etc_hosts)
       if host['platform'] =~ /freebsd/
         host.echo_to_file(etc_hosts, '/etc/hosts')
+      elsif ((host['platform'] =~ /windows/) and not host.is_cygwin?)
+        host.exec(Command.new("echo '#{etc_hosts}' >> C:\\Windows\\System32\\drivers\\etc\\hosts"))
       else
         host.exec(Command.new("echo '#{etc_hosts}' >> /etc/hosts"))
       end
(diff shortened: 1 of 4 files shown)


james-powis:
Upon further investigation found that the vagrant bug was due to regression, instead of testing each possible effected version, decided best route was to explicitly set ssh forward.
    Effected vagrant versions 1.6.2, 1.6.3, 1.6.4, 1.7.3+
    Uneffected vagrant versions <= 1.6.2, 1.7.0 - 1.7.2
    
The bug causes ssh vagrant@#{SUT} to fail to map to ssh -p 2222 vagrant@localhost. Explicitly setting absolutely ensures ssh forwarding functions.

james-powis:
Bump?

tvpartytonight:
@james-powis thanks for this PR; I think that this needs a rebase from puppetlabs:master so we can get clean history with no merge commits. Thanks!

james-powis:
Well that was a bit of a battle, but I got it all rebased and squashed down. I made sure credit is paid where credit is due to all the contributors in the commit message.

lib/beaker/host_prebuilt_steps.rb
```diff
@@ -367,13 +367,15 @@ def copy_ssh_to_root host, opts
     def hack_etc_hosts hosts, opts
       etc_hosts = "127.0.0.1\tlocalhost localhost.localdomain\n"
       hosts.each do |host|
-        ip = host['vm_ip'] || host['ip'].to_s
-        hostname = host[:vmhostname] || host.name
-        domain = get_domain_name(host)
-        etc_hosts += "#{ip}\t#{hostname}.#{domain} #{hostname}\n"
+        unless host['platform'] =~ /windows/
```
tvpartytonight: @james-powis can you provide some context on why this is restricted to non-windows platforms? I'm pretty sure that internally here at puppet, we will still want this functionality on windows as well as unix based systems.

james-powis: @petems could provide more insight but looking at the logic the get_domain_name and set_etc_hosts functions both are coded with *nix / cygwin paths, I would agree that that omitting host file hacking works just fine for single node test cases, but it does in fact break multi node test cases across the board. 

I will get the appropriate logic in place.

james-powis:
@tvpartytonight I removed the offending "if/unless windows" commands and fixed the underlying problems in the get_domain_name and set_etc_hosts functions to provide native windows commands when bitvise is in use my local tests confirm all is well.

lib/beaker/host_prebuilt_steps.rb
petems: Yep, basically it should be restricted to platform windows and is_cygwin false, as the standard approaches for things like domain and hosts dont work in PS Windows 👍

petems:
@james-powis Thanks for taking the reins on this, had it on the backburner for a while, keep forgetting to bring back upstream! 👍

tvpartytonight:

tvpartytonight:

tvpartytonight:

tvpartytonight:

tvpartytonight:
Thanks for being so responsive for this work, @james-powis ! It'll be rolled up into the next beaker release.

PR #1120, opened 2016-05-19
(maint) Change spec test to use simplest assertion

tvpartytonight:
Changes to minitest 5.9 broke a single test that used assert_equal to
generate a minitest assertion; this change uses the simplest assertion
possible, so changes to minitest are less likely to break a spec test.


kevpl:
I'm confident the update shouldn't effect the CI results, so I'm down with this. 👍

PR #1125, opened 2016-05-25
(MAINT) clarifying between install docs

kevpl:
This confusion isn't actually about the docs themselves. Both phrases "beaker-installation" and "how to install" link to the same document. But they are listed in the same TOC, which doesn't make much sense.

This PR is to clarify between the two entries and gain a little more consistency here.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -3,10 +3,9 @@
 ## Beaker Info
 
 * [MAINTAINERS](/MAINTAINERS.md)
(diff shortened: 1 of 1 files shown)


tvpartytonight:
@kevpl I think this needs a rebase.

kevpl:
@tvpartytonight it's ready. I killed those jobs with my bare hands!

Not sure if any of the things we put in for doc fixes worked, since it seems to have kicked off anyway... 💨

PR #1129, opened 2016-05-27
(QA-2501) add ability to install hosts in parallel

samwoods1:
Add :install_in_parallel option.  Default to false (existing behavior)
Run timesync in parallel if :install_in_parallel
Updated stop_agent_on to accept an array of hosts (as it was documented)
Updated sign_certificate_for to take an array of hosts (as it was documented)
install_puppet_on method now checks :install_in_parallel option
Added ability to run in parallel to run_block_on - closes host SSH connection in run_block_on after each parallel command is issued
Added run_in_parallel argument (defaulting to false) to:
- run_block_on
- block_on
- on
- install_puppet_agent_on
- apply_manifest_on
- stop_agent_on
- execute_powershell_script_on


samwoods1:

samwoods1:

samwoods1:

samwoods1:

lib/beaker/dsl/helpers/puppet_helpers.rb
```diff
@@ -714,35 +716,51 @@ def wait_for_host_in_dashboard(host)
...
         #                            or a role (String or Symbol) that identifies one or more hosts.
+        #                            No argument, or an empty array means no validation of success
+        #                            for specific hosts will be performed.
         #
         # @return nil
         # @raise [FailTest] if process times out
-        def sign_certificate_for(host)
-          block_on host do | host |
-            if [master, dashboard, database].include? host
+        def sign_certificate_for(host = [])
```
james-stocks: The change on this line stops someone getting an arity exception if they pass nothing for "host". I suspect the method would now just do nothing at all if someone made that mistake? I'm not sure if this is worth worrying about though - the method name plus the yardoc makes it sound strange that anyone would expect to not pass a value for host

lib/beaker/dsl/helpers/puppet_helpers.rb
```diff
@@ -401,11 +401,12 @@ def bounce_service host, service, curl_retries = nil, port = nil
         # @param [Block] block This method will yield to a block of code passed
         #                      by the caller; this can be used for additional
         #                      validation, etc.
+        # @param [Boolean] run_in_parallel Whether to run on each host in parallel.
         #
         # @return [Array<Result>, Result, nil] An array of results, a result object,
         #   or nil. Check {#run_block_on} for more details on this.
-        def apply_manifest_on(host, manifest, opts = {}, &block)
-          block_on host do | host |
+        def apply_manifest_on(host, manifest, opts = {}, run_in_parallel=false, &block)
```
james-stocks: Should run_in_parallel be a new and separate param when we have an opts hash?

lib/beaker/dsl/helpers/puppet_helpers.rb
```diff
@@ -714,35 +716,51 @@ def wait_for_host_in_dashboard(host)
...
+        def sign_certificate_for(host = [])
+          hostnames = []
+          hosts = host.is_a?(Array) ? host : [host]
+          hosts.each{ |current_host|
+            if [master, dashboard, database].include? current_host
 
-              on host, puppet( 'agent -t' ), :acceptable_exit_codes => [0,1,2]
-              on master, puppet( "cert --allow-dns-alt-names sign #{host}" ), :acceptable_exit_codes => [0,24]
+              on current_host, puppet( 'agent -t' ), :acceptable_exit_codes => [0,1,2]
+              on master, puppet( "cert --allow-dns-alt-names sign #{current_host}" ), :acceptable_exit_codes => [0,24]
```
james-stocks: Can this line end up being done in parallel? The master might not like trying to do several signings at once if it's locking files in the ssl dir or elsewhere.

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -245,21 +245,23 @@ def install_puppet(opts = {})
...
             opts[:puppet_agent_version] ||= opts[:version]
-            install_puppet_agent_on(hosts, opts)
+            install_puppet_agent_on(hosts, run_in_parallel, opts)
 
           else
-            block_on hosts do |host|
+            block_on hosts, run_in_parallel do |host|
               if host['platform'] =~ /el-(5|6|7)/
                 relver = $1
                 install_puppet_from_rpm_on(host, opts.merge(:release => relver, :family => 'el'))
```
james-stocks: If this is one of the beaker methods that scp's from the source to the test executor and then scp's from the executor to the destination host then it might behave oddly if running in parallel - host 1 might touch the file and begin transferring to it; then host 2 might see the touched (but only partially downloaded) file and assume it's the complete file.
If this method doesn't; there are definitely methods that do, such as create_remote_file

lib/beaker/dsl/helpers/puppet_helpers.rb
samwoods1: Yeah, this method originally took a single host, issued a command to the master to sign all certs, and then validated that a certificate for that single host exists.  I changed it so that you could pass an array of hosts and that it would sign all certs and then validate that certs for all of the hosts were signed.  If you pass nothing in, it will execute the command to sign all certs, but do no validation that any specific hosts are included in the list of certs.

lib/beaker/dsl/helpers/puppet_helpers.rb
samwoods1: Yeah, I think it has to be.  The main opts hash is typically passed around Beaker from method to method.  If you want to be able to execute the same method in parallel or not in parallel depending on the context, it pretty much needs to be outside of the options hash, unless I'm misunderstanding something.

lib/beaker/dsl/helpers/puppet_helpers.rb
samwoods1: This isn't done in parallel. This is a special case for any host that is a master, dashboard or database.  Any other hosts will fall into the else below where 'cert --sign --all' is executed a single time.

lib/beaker/dsl/install_utils/foss_utils.rb
samwoods1: Hmm... I have not exhaustively tested this, but I would expect in this case - or any case where it would otherwise cause things to blow up - you would opt not to run in parallel.  This would likely require some experimentation to determine whether it would work or not.

tvpartytonight:
@samwoods1 this PR needs acceptance tests; let's remove that label after they have been submitted for this work.

lib/beaker/dsl/helpers/puppet_helpers.rb
nwolfe: Hm, but it's provided as an option in the `opts` elsewhere?  And it truly is an option.

I think we should keep it in the `options` hash, that way we won't need to modify a bunch of methods to accept a new parameter, and we can avoid lines that simply pull it out of the hash and pass it down explicitly; everything already accepts the `opts` hash and passes it around, we should use that.

> If you want to be able to execute the same method in parallel or not in parallel depending on the context, it pretty much needs to be outside of the options hash, unless I'm misunderstanding something

The location of the value shouldn't inhibit this; it can remain in the options hash, but would just need to be updated in their between method calls.

> any calls to these methods will still default to run_in_parallel=false

We can still do this without using method parameter default values, right?

> The reason for having this as a separate argument to the method is so you can opt on a per call basis whether to execute in parallel or not.

That's still true when it's an `options`, it just takes a few extra characters of code, right?

Overall, I think it would be wise to organize this based on the structure of the code and not on the amount of effort/characters necessary to use it.  Since we've already got `options` plumbed throughout the code, and we're introducing a new option, then it would make sense to fold it into the existing convention.  It does require a tiny bit more code, but that's pretty cheap.

lib/beaker/dsl/helpers/puppet_helpers.rb
nwolfe: The name of this method is misleading.  Without looking at anything other than the method signature, I would expect that `host` is an array of hostname strings that will be passed to `puppet cert sign <hostname>`, and if I gave it an empty list then it would be a NOOP.

I see that this was already misleading prior to these changes, but since we're in there and altering the behavior, we should consider improving this.  Ideally we'd rename the method name to be more clear, but if we want to avoid "API" changes then we should at least improve the docstring a bunch.

I think the docstring should first and foremost say that it will sign all certificates, regardless of the parameter value.

lib/beaker/dsl/helpers/host_helpers.rb
```diff
@@ -59,8 +60,8 @@ module HostHelpers
         #   on :dashboard, "echo hello"
         # @return [Result]   An object representing the outcome of *command*.
         # @raise  [FailTest] Raises an exception if *command* obviously fails.
-        def on(host, command, opts = {}, &block)
-          block_on host do | host |
+        def on(host, command, opts = {}, run_in_parallel=false, &block)
```
nwolfe: I think we should put this in the `opts` rather than introduce a new parameter.  A nice benefit of that is that we no longer have to make a breaking change to this method.  And if we ever remove the `run_in_parellel` option, similarly, we won't have to make a breaking change to this method (or any method) signature.

nwolfe:
@samwoods1 Nice job, I like these changes.  Overall, I think we should pursue keeping this as an options in the various `opts` parameters, and we definitely need acceptance tests.  I'm happy to help with either!

samwoods1:

lib/beaker/dsl/helpers/puppet_helpers.rb
samwoods1: @nwolfe I updated the yardoc for this method.

samwoods1:
Much of the acceptance testing will be performed by this: https://jenkins-beaker.delivery.puppetlabs.net/job/In%20parallel%20job%20(In%20progress)/ which will execute some basic validation after running a parallel install for the following configurations: pe, pe with frictionless agent, FOSS. This will run on a matrix of agent OS's.  I'm looking at possibly doing some upgrade tests as well.

samwoods1:

beaker.gemspec
```diff
@@ -40,6 +40,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'inifile', '~> 2.0'
   s.add_runtime_dependency 'rsync', '~> 1.0.9'
   s.add_runtime_dependency 'open_uri_redirections', '~> 0.2.1'
+  s.add_runtime_dependency 'in-parallel', '~> 0.1.8'
```
tvpartytonight: Does this have to specify the z component? I think that `0.1` is actually what you want specify, unless you don't plan on revving up to the `0.2.0` version anytime soon...

samwoods1: Anything prior to 0.1.8 would not work, I suppose that means I should update it to 0.2.0 since there were breaking changes.

tvpartytonight: The way the dependency is set now, beaker will not automatically get the `0.2` release, it will only get the latest `0.1.x` release. By specifying `~> 0.1` it will install the latest available gem in the range of `> 0.1.0` to `< 1.0`.

See this article for more pessimistic operator explaining: https://robots.thoughtbot.com/rubys-pessimistic-operator

lib/beaker/dsl/helpers/puppet_helpers.rb
```diff
@@ -714,35 +717,52 @@ def wait_for_host_in_dashboard(host)
...
+              end
+              on master, puppet("cert --sign --all --allow-dns-alt-names"), :acceptable_exit_codes => [0,24]
+              out = on(master, puppet("cert --list --all")).stdout
+              if hostnames.all? { |hostname| out =~ /\+ "?#{hostname}"?/ }
+                hostnames.clear
+                break
               end
 
+              sleep next_sleep
+              (last_sleep, next_sleep) = next_sleep, last_sleep+next_sleep
```
tvpartytonight: 💤

lib/beaker/dsl/helpers/host_helpers.rb
```diff
@@ -21,10 +21,11 @@ module HostHelpers
         #   @option opts [Boolean] :pty (false) Execute this command in a pseudoterminal.
         #   @option opts [Boolean] :expect_connection_failure (false) Expect this command
         #     to result in a connection failure, reconnect and continue execution.
+        #   @option opts [Boolean] :run_in_parallel Whether to run on each host in parallel.
         #   @option opts [Hash{String=>String}] :environment ({}) These will be
         #     treated as extra environment variables that should be set before
         #     running the command.
-        #
+        #   @option opts [Boolean] :run_in_parallel Whether to run on each host in parallel.
```
tvpartytonight: Is this a duplicated of line 24 ^^?

samwoods1: Yep, it was.  Removed it.

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -334,7 +340,7 @@ def install_puppet_agent_on(hosts, opts = {})
           opts[:puppet_collection] ||= 'pc1' #hi!  i'm case sensitive!  be careful!
           opts[:puppet_agent_version] ||= opts[:version] #backwards compatability with old parameter name
 
-          block_on hosts do |host|
+          block_on hosts, { :run_in_parallel => @options[:run_in_parallel].include('install') } do |host|
```
kevpl: should this not be passable from the `opts` hash as well?

should this go into a little helper method so that it's done the same way in all of these methods?

docs/Lets-Write-a-Test.md
```diff
@@ -18,6 +18,8 @@ We prefer to install PE once and then run a set of tests, so PE installation sho
     $ cd setup
     $ cat > install.rb << RUBY
     test_name "Installing Puppet Enterprise" do
+    # Particularly when installing a larger number of hosts, this can reduce the execution time.
+    options[:run_in_parallel] = ["configuration", "install"]
```
kevpl: The code instances below are using the value 'configure' for this.

Running these tests in parallel seems like a big enough thing that it could use it's own doc. Would you mind writing up a small one, so that you can highlight these two use cases (these coded values vs running something like `on()` with `:run_in_parallel` set to `true`)?  I think a question that will come up when people see this is whether or not these are all the values that can be had for this setting as well. Having its own doc where you clearly delineate these different use cases would be helpful, rather than having people need to poke around in various docs for the different ways.

You can also highlight the environment variable method for passing as well there.

lib/beaker/shared/host_manager.rb
```diff
@@ -1,3 +1,5 @@
+require 'in_parallel'
```
kevpl: it looks like this require is set here to be a one-off, or to limit the scope of this change. But we're changing `run_block_on`, so there's no real limit to the scope of this change :). Is there another reason that this require is here rather than with the others?

kevpl:
@samwoods1, we were discussing the potential timing of merging & releasing this change earlier today, and were thinking of potentially giving it its own Y release, in order to minimize risk. This would be a follow-on to the [June 29th currently planned release](https://github.com/puppetlabs/beaker/wiki).

One thing that would be helpful from both a minimizing risk standpoint and a "getting your changes out faster" standpoint would be if you separated out the unrelated bugfixes from the parallel work. Then we could merge those into the next release, and make this as small as possible for it's own release. Is this something that you have time for?

tvpartytonight:
@samwoods1 to summarize our discussion yesterday:

1) Alter/make configurable the behavior of in-parallel to wait for all forks to finish before raising exceptions; I think the default should be to fail "slow", so all forks have a chance to complete.

2) Add a default timeout for child process completion; this is just a failsafe so that something doesn't hang forever. I'm guessing that 30 min is a decent default to start with? I'm also assuming that this is something outside of the code in this PR...I'd love to take a look at that code when it's ready.

3) Solving the "in-parallel" level setting. We had talked about a none, conservative, and aggressive settings. I'm still unsure of what the best solution might be...

Overall, my ad hoc testing of this has been pretty good. I'm excited to merge in the functionality.

samwoods1:
@kevpl  the only bug fixes I did were things that were required to do this work.  Was there particular stuff you wanted to see in a separate PR?

samwoods1:
Ok... I believe I have addressed all of the feedback up to this point.  
@tvpartytonight 
1) I updated in-parallel to wait for all forked processes to complete before re-raising an exception from the child process by default (and optionally fail fast)
2) Added default timeout https://github.com/puppetlabs/in-parallel/pull/6
3) Changed install_in_parallel to run_in_parallel with array of values.  Currently only supports 'configure' and 'install'

samwoods1:

docs/Argument-Processing-and-Precedence.md
```diff
@@ -211,77 +220,60 @@ Values already included in Beaker as defaults for required arguments.
...
           :help                   => false,
-          :collect_perf_data      => false,
+          :collect_perf_data      => 'none',
+          :puppetdb_port_ssl      => 8081,
+          :puppetdb_port_nonssl   => 8080,
+          :puppetserver_port      => 8140,
+          :nodeclassifier_port    => 4433,
+          :cache_files_locally    => true,
+          :aws_keyname_modifier   => rand(10 ** 10).to_s.rjust(10,'0'), # 10 digit random number string
+          [:run_in_parallel](runner/run_in_parallel.md)        => [],
```
kevpl: this link doesn't appear to work viewing the page in github

docs/runner/run_in_parallel.md
```diff
@@ -0,0 +1,59 @@
...
+- install_puppet_agent_on
+- apply_manifest_on
+- stop_agent_on
+- execute_powershell_script_on
+
+## Using InParallel in your test scripts
+In addition to the options, you can use InParallel within your test scripts as well.
+
+Examples:
+```
```
kevpl: I believe if you add ruby to this, you'll get nice syntax highlighting. Doesn't need to block merging

docs/runner/run_in_parallel.md
```diff
@@ -0,0 +1,59 @@
+# run_in_parallel global and command options
+
+## run_in_parallel global option
+The run_in_parallel global option is an array with the following possible values: ['configure', 'install']
+It defaults to an empty array []
+It can be set in an options file, or overriden by the BEAKER_RUN_IN_PARALLEL environment variable
+example:
+
+export BEAKER_RUN_IN_PARALLEL=configure,install
```
kevpl: 💯, thanks for explaining how this env variable will work!

docs/Argument-Processing-and-Precedence.md
samwoods1: Github is having issues updating... This is corrected in my last push, you can see it is correct in the files tab, but still showing up here...

kevpl:

pinkypie:
retest this please

samwoods1:
retest this please

samwoods1:
I made some minor updates to the in-parallel library to address some of the things Tony mentioned to me, mainly the logging.  Now it is using a logger, and will swap out it's own simple logger with the Beaker logger if it is available.  I also consolidated a lot of the logging so there are fewer unique log messages.  

I Also made a couple of bug fixes, one to address the timeouts we were seeing on Fedora and Ubuntu tests, and one to fix an issue where pressing CTRL-C or stopping a debugging session would always kill the child proc's, but wouldn't abort execution of the program, you always had to do it a second time to truly stop.  I think how things are should be pretty stable now.

waynr:
@samwoods1 so in ESO Codecraft today I mentioned the fact that (at least for *nix-like systems) when forking a new subprocess each new subprocess will inherit a COW clone of all the parent process's memory segments as well as copies of the parent process' file descriptors. From that point on, any changes made in the child process will not be reflected in the parent process' memory space, ie changes made to variables defined outside the parellelized block. 

I understand that you have provided for interprocess communication by allowing forked processes to return values to the parent process but given the proclivity of ruby programmers to modify non-local state it should be made more clear than it currently is that what happens in the child process stays in the child process (unless explicitly returned--but of course, that part you have already documented).

lib/beaker/shared/host_manager.rb
```diff
@@ -100,8 +102,18 @@ def run_block_on hosts = [], filter = nil, &block
...
-            result = block_hosts.map do |h|
-              run_block_on h, nil, &block
+            if opts[:run_in_parallel]
+              # Pass caller[1] - the line that called block_on - for logging purposes.
+              result = block_hosts.map.each_in_parallel(caller[1]) do |h|
+                run_block_on h, &block
+              end
+              hosts.each{|host| host.close}# For some reason, I have to close the SSH connection
+              # after spawning a process and running commands on a host,
+              # or else it gets into a broken state for the next call.
```
waynr: This may have something to do with the behavior of the SSH library in use. Is it possible that the the SSH library closes the connections on process termination? In that case, the SSH library would probably only be deleting the fd from the child process' fd table--the original copy of the fd in the parent process would still be open even though the connection itself is closed. Closing the connection manually here would clean up that fd and allow the host object to properly reestablish the connection on the next attempt. For what it's worth, I am not personally familiar with the SSH connection code in use here, @KevinCorcoran or @tvpartytonight can provide more insight.

waynr: Whoops, I meant @kevpl

waynr: I also meant "can maybe provide more insight"

samwoods1:
@waynr Good point, I updated documentation in run_in_parallel.md to include this info.

samwoods1:

tvpartytonight:
LGTM!

tron in parallel

PR #1130, opened 2016-05-27
(BKR-814) added beaker-hostgenerator syntax to hosts arg

kevpl:


nwolfe:
@kevpl If a BHG string is passed in, do we need to involve files on disk at all?  

I can't quite tell from the code whether the generated file is an implementation detail or something that the user would be aware of and perhaps use.

Or, is it that the majority of code in Beaker will parse the hosts file directly, and so you're trying to avoid changing all that code at once?

I was assuming the hosts file was only parsed once at startup, and from there on everything would just work with the parsed `Hash` map.  If that was the case, we wouldn't need to involve a file since we could just do `YAML.parse_string( BeakerHostGenerator::CLI.new(<hosts input string>).execute )`

kevpl:
@nwolfe interesting, this would be simpler than the method I took, which I suspect might be due to outdated reasoning. I'll take a look at this when I get some time.

kevpl:
a question just to clarify: I'm wondering if beaker users would expect to have a hosts file available when they use this feature, regardless of whether they use `--preserve-hosts`.

@puppetlabs/beaker ?
@tvpartytonight ? @er0ck ?

er0ck:
i'd expect this to NOT generate a hosts file, other than a preserved_hosts file.  ie, i'd expect startup to be faster by some (unmeasurable?) amount.

lib/beaker/options/parser.rb
```diff
@@ -215,6 +221,37 @@ def parse_args(args = ARGV)
...
+        hosts_options = Beaker::Options::HostsFileParser.parse_hosts_file(@options[:hosts_file])
+        return hosts_options, nil
+      rescue Errno::ENOENT
+        require 'tempfile'
+        require 'beaker-hostgenerator'
+        bhg_cli = BeakerHostGenerator::CLI.new( [ @options[:hosts_file] ] )
+        hosts_file_content = bhg_cli.execute
+        hosts_file = Tempfile.open( 'hosts.yml' )
+        begin
+          hosts_file.write( hosts_file_content )
```
er0ck: yeah, what @nwolfe said. I don't really understand what `:hosts_file_generated` options is doing, in general.  but also because it appears that we're always writing a `./hosts.yml` here.  current dir seems like a non-great place, but in addition i'd expect beaker to just write to the preserved_hosts file (created another ticket for earlier serialization of preserved_hosts: https://tickets.puppetlabs.com/browse/BKR-828).

er0ck:
BTW, in general this is looking great, and i'm very excited for this functionality!

lib/beaker/options/parser.rb
tvpartytonight: I'm also not quite sure why we are tracking whether or not we generated the file with the `:hosts_file_generated` key; it seems like we are tracking it just to ensure we delete it in the end, but since it is generated with the `Tempfile` class, I'm not sure why we care? @er0ck this file wouldn't get generated in current dir, fyi, but in the system's temporary directory.

tvpartytonight:
I'm ok with this generating a system tmpfile, but it could prove to be a simpler and easier to maintain if we sidestepped that whole process and just dealt with the in-memory hash, as @nwolfe described. I could go either way on this issue...

Besides that issue, this PR seems to necessitate some [command_line_parser.rb](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/options/command_line_parser.rb) changes to reflect that it can also take a host-generator string.

tvpartytonight:
@kevpl this could also coincide with the initiative to have beaker look for a default yaml configuration, rather than force the user to define where that config is. Maybe that is conflating this PR with too much intention, but it does seem super convenient...

kevpl:

tvpartytonight:

tvpartytonight:
@kevpl this looks good to me, other than the case of possibly returning a host-generator parse error when we may really want to return a `Errno::ENOENT` instead. It's an edge case that we can patch up in future ticket.

nwolfe:
Oh...I didn't realize comments were addressed on this.  I was hoping to review again before merge.

lib/beaker/options/parser.rb
```diff
@@ -90,7 +90,9 @@ def file_list(paths)
       # @return nil
       # @api public
       def resolve_symlinks!
-        @options[:hosts_file] = File.realpath(@options[:hosts_file]) if @options[:hosts_file]
+        if @options[:hosts_file] && !@options[:hosts_file_generated]
```
nwolfe: So we weren't able to hide the fact that the file wasn't pre-generated and had to introduce this "option" to detect that?

er0ck: i'm with you on this one.  will vote on a ticket to move to some singleton state object, or similar

PR #1137, opened 2016-06-07
(BKR-832) Add `box_download_insecure` for vagrant

op-ct:
Without this patch, vagrant nodesets cannot disable SSL server certificate
verification.  This patch fixes the issue by introducing a
`box_download_insecure` option for vagrant hosts.


kevpl:
Thanks for the improvement @op-ct! Would you add a small section describing this change just like you did above in our [Vagrant doc](https://github.com/puppetlabs/beaker/blob/master/docs/hypervisors/vagrant.md#vagrant-specific-hosts-file-settings), so that other's can more easily learn about and use it?

kevpl:
If no movement is made on this in a week, we will close it due to inactivity.

kevpl:
Closing due to inactivity

PR #1140, opened 2016-06-09
(BKR-355) - Adding acceptance tests for web_helper methods.

pinkypie:
First pass at acceptance tests for web_helpers, for comment.

### DIFF
--- a/acceptance/tests/base/dsl/helpers/web_helpers/web_helpers_test.rb
+++ b/acceptance/tests/base/dsl/helpers/web_helpers/web_helpers_test.rb
@@ -2,65 +2,54 @@ require 'helpers/test_helper'
 require 'webrick'
 require 'webrick/https'
 
+confine :except, :platform => %w(windows solaris-11 osx)
+
 test_name 'dsl::helpers::web_helpers #link_exists?' do
-
-  http_cmd = "ruby -rwebrick -e'WEBrick::HTTPServer.new(:Port => 80, :DocumentRoot => Dir.pwd).start' > /tmp/mylogfile 2>&1 &"
-  host = hosts[0]
-  on(host, http_cmd)
-  on(host, 'mkdir tmp')
-  create_remote_file(host, 'tmp/test.txt', 'test')
-
-  dst_dir = 'web_helpers/'
+  cert_name = [
+      %w[CN localhost],
+  ]
+  http_cmd = "ruby -rwebrick -e'WEBrick::HTTPServer.new(:Port => 80, :DocumentRoot => \"/tmp\").start' > /tmp/mylogfile 2>&1 &"
+  https_cmd = "ruby -rwebrick/https -e'WEBrick::HTTPServer.new(:SSLEnable => true, :SSLCertName => #{cert_name}, :Port => 4430,:DocumentRoot => \"/tmp\").start' > /tmp/mylogfile 2>&1 &"
+  on(default, http_cmd)
+  on(default, https_cmd)
+  dir = default.tmpdir('test_dir')
+  file = default.tmpfile('test_file')
+  dir.slice! "/tmp"
+  file.slice! "/tmp"
+  dst_dir = 'web_helpers'
 
   step '#port_open_within? can tell if a port is open' do
-    assert port_open_within?(host,80,1)
+    assert port_open_within?(default,80)
   end
 
   step '#link_exists? can tell if a basic link exists' do
-    assert link_exists?("http://#{host}")
+    assert link_exists?("http://#{default}")
   end
 
   step '#link_exists? can tell if a basic link does not exist' do
-    assert !link_exists?("http://#{host}/test")
+    assert !link_exists?("http://#{default}/test")
   end
 
   step '#link_exists? can use an ssl link' do
-    host = hosts[0]
-
-    cert_name = [
-        %w[CN localhost],
-    ]
-
-    http_cmd = "ruby -rwebrick/https -e'WEBrick::HTTPServer.new(:SSLEnable => true, :SSLCertName => #{cert_name}, :Port => 4430,
-:DocumentRoot => Dir.pwd).start' > /tmp/mylogfile 2>&1 &"
-
-    sleep(2)
-    on(host, http_cmd)
-    assert link_exists?("https://#{host}:4430")
+    assert link_exists?("https://#{default}:4430")
   end
 
   step '#fetch_http_dir can fetch a dir' do
-    assert_equal "#{dst_dir}tmp", fetch_http_dir("http://#{host}/tmp", dst_dir)
+    assert_equal "#{dst_dir}#{dir}", fetch_http_dir("http://#{default}/#{dir}", dst_dir)
   end
 
   step '#fetch_http_dir will raise an error if unable fetch a dir' do
-    begin
-      fetch_http_dir("http://#{host}/tmps", dst_dir)
-    rescue RuntimeError => e
-      assert_match /Failed to fetch_remote_dir.*/, e.message, "#fetch_http_dir raised an unexpected RuntimeError"
-    end
+    exception = assert_raises(RuntimeError) { fetch_http_dir("http://#{default}/tmps", dst_dir) }
+    assert_match /Failed to fetch_remote_dir.*/, exception.message, "#fetch_http_dir raised an unexpected RuntimeError"
   end
 
   step '#fetch_http_file can fetch a file' do
-    assert_equal "#{dst_dir}test.txt", fetch_http_file("http://#{host}/tmp", "test.txt", dst_dir)
+    assert_equal "#{dst_dir}#{file}", fetch_http_file("http://#{default}", file, dst_dir)
   end
 
   step '#fetch_http_file will raise an error if unable to fetch a file' do
-    begin
-      fetch_http_file("http://#{host}/tmp", "test2.txt", dst_dir)
-    rescue RuntimeError => e
-      assert_match /Failed to fetch_remote_file.*/, e.message, "#fetch_http_file raised an unexpected RuntimeError"
-    end
+    exception = assert_raises(RuntimeError) { fetch_http_file("http://#{default}", "test2.txt", dst_dir) }
+    assert_match /Failed to fetch_remote_file.*/, exception.message, "#fetch_http_dir raised an unexpected RuntimeError"
   end
 
 end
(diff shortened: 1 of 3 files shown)


acceptance/tests/base/dsl/helpers/web_helpers/web_helpers_test.rb
```diff
@@ -0,0 +1,66 @@
...
+require 'webrick'
+require 'webrick/https'
+
+test_name 'dsl::helpers::web_helpers #link_exists?' do
+
+  http_cmd = "ruby -rwebrick -e'WEBrick::HTTPServer.new(:Port => 80, :DocumentRoot => Dir.pwd).start' > /tmp/mylogfile 2>&1 &"
+  host = hosts[0]
+  on(host, http_cmd)
+  on(host, 'mkdir tmp')
+  create_remote_file(host, 'tmp/test.txt', 'test')
```
nwolfe: For all this temp file/dir stuff, there's convenience methods on `host` for doing it, something like `host.tmpdir('my-test')` that results in `on(host, "mkdir <proper temp directory for OS>/my-test")`.  

There might be a couple of methods for working with temp files/dirs in beaker, probably worth a look.

lib/beaker/dsl/helpers/web_helpers.rb
```diff
@@ -52,9 +53,17 @@ def fetch_http_file(base_url, file_name, dst_dir)
...
-                FileUtils.copy_stream(remote, file)
+            begin
+              open(src, :allow_redirections => :all) do |remote|
+                File.open(dst, "w") do |file|
+                  FileUtils.copy_stream(remote, file)
+                end
+              end
+            rescue OpenURI::HTTPError => e
+              if e.message =~ /404.*/
+                raise "Failed to fetch_remote_file '#{src}' (exit code #{$?}"
```
nwolfe: Whoa, I'm having trouble figuring out what `$?` is going to get.  Typically I see `exit_code` being called on something to get the exit code.   This looks like bash but I don't see any bash executed nearby?

Also, the message is missing the trailing `)` after the exit code.

lib/beaker/dsl/helpers/web_helpers.rb
```diff
@@ -52,9 +53,17 @@ def fetch_http_file(base_url, file_name, dst_dir)
...
-              File.open(dst, "w") do |file|
-                FileUtils.copy_stream(remote, file)
+            begin
+              open(src, :allow_redirections => :all) do |remote|
+                File.open(dst, "w") do |file|
+                  FileUtils.copy_stream(remote, file)
+                end
+              end
+            rescue OpenURI::HTTPError => e
+              if e.message =~ /404.*/
```
nwolfe: Is the status code available at `e.code`?  I can't quite tell from ruby docs, it's not exactly clear; didn't dig into the source code either.

acceptance/tests/base/dsl/helpers/web_helpers/web_helpers_test.rb
```diff
@@ -0,0 +1,66 @@
+require 'helpers/test_helper'
+require 'webrick'
+require 'webrick/https'
+
+test_name 'dsl::helpers::web_helpers #link_exists?' do
+
+  http_cmd = "ruby -rwebrick -e'WEBrick::HTTPServer.new(:Port => 80, :DocumentRoot => Dir.pwd).start' > /tmp/mylogfile 2>&1 &"
+  host = hosts[0]
```
nwolfe: Is there any expectation that `hosts[0]` is a particular role, or just any host will work?

acceptance/tests/base/dsl/helpers/web_helpers/web_helpers_test.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  step '#fetch_http_dir can fetch a dir' do
+    assert_equal "#{dst_dir}tmp", fetch_http_dir("http://#{host}/tmp", dst_dir)
+  end
+
+  step '#fetch_http_dir will raise an error if unable fetch a dir' do
+    begin
+      fetch_http_dir("http://#{host}/tmps", dst_dir)
+    rescue RuntimeError => e
```
nwolfe: Could we use `assert_raise` here? http://ruby-doc.org/stdlib-2.1.0/libdoc/test/unit/rdoc/Test/Unit/Assertions.html#method-i-assert_raise

acceptance/tests/base/dsl/helpers/web_helpers/web_helpers_test.rb
kevpl: [host `tmpdir` & `tmpfile`](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host/unix/file.rb#L4-L10)

pinkypie: I Initially didn't use them as didn't have the ability to pass the -p flag to the mktemp command to specify a different tmp directory (it defaults to /tmp and the tmpdir method is hardcoded to only pass the -dt flag).
I was serving the current directory with webrick which was /root.
Anyway, I changed it to serve the /tmp directory and switched to use those methods instead.

acceptance/tests/base/dsl/helpers/web_helpers/web_helpers_test.rb
pinkypie: Yes, switched it to use assert_raise and assert_match on the error message.

acceptance/tests/base/dsl/helpers/web_helpers/web_helpers_test.rb
pinkypie: It should probably use master for the webserver, have updated.

lib/beaker/dsl/helpers/web_helpers.rb
pinkypie: Changed it to e.message (404 not found).

acceptance/tests/base/dsl/helpers/web_helpers/web_helpers_test.rb
```diff
@@ -0,0 +1,60 @@
...
+    assert link_exists?("https://#{master}:4430")
+  end
+
+  step '#fetch_http_dir can fetch a dir' do
+    assert_equal "#{dst_dir}#{dir}", fetch_http_dir("http://#{master}/#{dir}", dst_dir)
+  end
+
+  step '#fetch_http_dir will raise an error if unable fetch a dir' do
+    exception = assert_raises(RuntimeError) { fetch_http_dir("http://#{master}/tmps", dst_dir) }
+    assert_match /Failed to fetch_remote_dir.*/, exception.message, "#fetch_http_dir raised an unexpected RuntimeError"
```
nwolfe: I believe `assert_raises` can take additional parameters, like the expected exception message.
So we could probably collapse these two assert calls into just the `assert_raises`

lib/beaker/dsl/helpers/web_helpers.rb
nwolfe: But what about looking for the status code directly, rather than the message?  The `e.message` is built from other attributes, like `e.code`, so we can just reference that directly.

The only problem is the docs suck for this and it's not clear from just googling, so we might just need to try it or look at the methods available at runtime.

I've seen plenty of Ruby code around here that does this exact thing, so I think it's pursuing for a few minutes.

pinkypie: Does such a thing exist? Runtime:
[1] pry(#<Beaker::TestCase>)> ls e -m
Exception#methods: 
  ==  backtrace  backtrace_locations  cause  exception  inspect  message  respond_to?  set_backtrace  to_s
OpenURI::HTTPError#methods: io

acceptance/tests/base/dsl/helpers/web_helpers/web_helpers_test.rb
pinkypie: Can it? You can pass an error message as the last arg, but it is the assertion error that is displayed when the assertion fails, not the expected error message.

nwolfe: I'll just say it's a very common way of writing exception assertion methods, but I admit I can't tell where these `assert_XXX` methods are coming from so I can't say for sure that these support that.

But if they don't, they should 😈  Oh well thanks for looking.

lib/beaker/dsl/helpers/web_helpers.rb
nwolfe: Yeah, I think I saw some rubydocs for this class that only had those methods, but then I saw another that had a bunch, and I couldn't tell which we had here.  This is totally fine, I was just thinking it was better to use the exit code field directly since that what you're after (not the message).

nwolfe:
👍 

Not sure what the acceptance failures mean...looks like it's not finding a host with the `master` role?

tvpartytonight:
The configuration for some of the jobs have no role for master, so that is why the job is failing...though, if memory serves, `default` should mean `master` if `master` isn't found? That's at least the issue that is failing [here](https://jenkins-beaker.delivery.puppetlabs.net/job/qe_beaker_intn-sys_beaker-acceptance-base-vpool/1828/agent=centos7/consoleFull).

acceptance/tests/base/dsl/helpers/web_helpers/web_helpers_test.rb
```diff
@@ -0,0 +1,60 @@
+require 'helpers/test_helper'
+require 'webrick'
+require 'webrick/https'
+
+test_name 'dsl::helpers::web_helpers #link_exists?' do
+
+  http_cmd = "ruby -rwebrick -e'WEBrick::HTTPServer.new(:Port => 80, :DocumentRoot => \"/tmp\").start' > /tmp/mylogfile 2>&1 &"
+  master = only_host_with_role(hosts, 'master')
```
tvpartytonight: Is this master assignment necessary? The `master` method just calls `only_host_with_role`, so I'm not sure if this is necessary.

pinkypie:
retest this please

pinkypie:
retest this please

pinkypie:
retest this please

pinkypie:
retest this please

pinkypie:
@tvpartytonight @kevpl the failure now was to do with a template for centos not being available on vmpooler.

pinkypie:
retest this please

kevpl:

PR #1143, opened 2016-06-14
(MAINT) improved error output from beaker-hostgenerator workflow

kevpl:
If you provide bad beaker-hostgenerator input (such as `poobah-baily` in my example below), you'll get this output from running beaker with this change:

```
beaker --log-level trace --hosts poobah-baily --tests ../my_beaker_files/tests/smoke_simple.rb

Hosts file 'poobah-baily' does not exist.
Trying as beaker-hostgenerator input.

WARNING: Starting with beaker-hostgenerator 1.x platform strings for "el" hosts
will correspond to the actual linux distribution name. ie, the platform string
corresponding to a host specified as "centos4_64a" will be "centos-4-x86_64"
rather than "el-4-x86_64". It is recommended that you update your project's test
suites ASAP or be forced to do so when beaker-hostgenerator development moves on
to the 1.x series. We don't intend to backport features or platforms to 0.x.

beaker-hostgenerator was not able to use this value as input.
Exiting with an Error.

beaker/.bundle/gems/gems/beaker-hostgenerator-0.7.0/lib/beaker-hostgenerator/parser.rb:161:in `parse_node_info_token': Invalid node_info token: poobah (BeakerHostGenerator::Exceptions::InvalidNodeSpecError)
	from beaker/.bundle/gems/gems/beaker-hostgenerator-0.7.0/lib/beaker-hostgenerator/generator.rb:46:in `block in generate'
	from beaker/.bundle/gems/gems/beaker-hostgenerator-0.7.0/lib/beaker-hostgenerator/generator.rb:37:in `each'
	from beaker/.bundle/gems/gems/beaker-hostgenerator-0.7.0/lib/beaker-hostgenerator/generator.rb:37:in `generate'
	from beaker/.bundle/gems/gems/beaker-hostgenerator-0.7.0/lib/beaker-hostgenerator/cli.rb:149:in `execute'
	from beaker/lib/beaker/options/parser.rb:239:in `rescue in parse_hosts_options'
	from beaker/lib/beaker/options/parser.rb:228:in `parse_hosts_options'
	from beaker/lib/beaker/options/parser.rb:197:in `parse_args'
	from beaker/lib/beaker/cli.rb:15:in `initialize'
	from beaker/bin/beaker:6:in `new'
	from beaker/bin/beaker:6:in `<top (required)>'
	from -e:1:in `load'
	from -e:1:in `<main>'

Process finished with exit code 1
```

### DIFF
--- a/lib/beaker/options/parser.rb
+++ b/lib/beaker/options/parser.rb
@@ -224,17 +224,29 @@ module Beaker
       # @raise [ArgumentError] if a hosts file is generated, but it can't
       #   be read by the HostsFileParser
       def parse_hosts_options
-        #read the hosts file that contains the node configuration and hypervisor info
-        hosts_options = Beaker::Options::HostsFileParser.parse_hosts_file(@options[:hosts_file])
-        return hosts_options
-      rescue Errno::ENOENT
-        require 'tempfile'
+        if File.exists?(@options[:hosts_file])
+          #read the hosts file that contains the node configuration and hypervisor info
+          return Beaker::Options::HostsFileParser.parse_hosts_file(@options[:hosts_file])
+        end
+
+        dne_message = "\nHosts file '#{@options[:hosts_file]}' does not exist."
+        dne_message << "\nTrying as beaker-hostgenerator input.\n\n"
+        $stdout.puts dne_message
         require 'beaker-hostgenerator'
-        bhg_cli = BeakerHostGenerator::CLI.new( [ @options[:hosts_file] ] )
-        hosts_file_content = bhg_cli.execute
-        hosts_options = Beaker::Options::HostsFileParser.parse_hosts_string( hosts_file_content )
+
+        hosts_file_content = begin
+          bhg_cli = BeakerHostGenerator::CLI.new( [ @options[:hosts_file] ] )
+          bhg_cli.execute
+        rescue BeakerHostGenerator::Exceptions::Error,
+          BeakerHostGenerator::Exceptions::InvalidNodeSpecError => error
+          error_message = "\nbeaker-hostgenerator was not able to use this value as input."
+          error_message << "\nExiting with an Error.\n\n"
+          $stderr.puts error_message
+          raise error
+        end
+
         @options[:hosts_file_generated] = true
-        return hosts_options
+        Beaker::Options::HostsFileParser.parse_hosts_string( hosts_file_content )
       end
 
       #Validate all merged options values for correctness
--- a/spec/beaker/options/parser_spec.rb
+++ b/spec/beaker/options/parser_spec.rb
@@ -153,7 +153,7 @@ module Beaker
             parser.instance_variable_set(:@command_line_parser, command_line_parser_obj)
 
             allow(OptionsFileParser).to receive(:parse_options_file).and_return(opt_file)
-            allow(HostsFileParser).to receive(:parse_hosts_file).and_return(host_file)
+            allow(parser).to receive(:parse_hosts_options).and_return(host_file)
           end
 
           it 'presets have the lowest priority' do
@@ -224,14 +224,11 @@ module Beaker
       describe '#parse_hosts_options' do
 
         context 'Hosts file exists' do
-          it 'returns nil as the second value' do
-            parser.instance_variable_set( :@options, {} )
-            allow( Beaker::Options::HostsFileParser ).to receive( :parse_hosts_file )
-            _, val2 = parser.parse_hosts_options
-            expect( val2 ).to be_nil
+          before :each do
+            allow(File).to receive(:exists?).and_return(true)
           end
 
-          it 'returns the parser\'s output as the first value' do
+          it 'returns the parser\'s output' do
             parser.instance_variable_set( :@options, {} )
             test_value = 'blaqwetjijl,emikfuj1235'
             allow( Beaker::Options::HostsFileParser ).to receive(
@@ -243,8 +240,10 @@ module Beaker
         end
 
         context 'Hosts file does not exist' do
-          require 'tempfile'
           require 'beaker-hostgenerator'
+          before :each do
+            allow(File).to receive(:exists?).and_return(false)
+          end
 
           it 'calls beaker-hostgenerator to get hosts information' do
             parser.instance_variable_set( :@options, {} )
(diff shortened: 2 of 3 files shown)


spec/beaker/options/parser_spec.rb
```diff
@@ -283,6 +283,31 @@ def mock_out_parsing
...
+            parser.instance_variable_set( :@options, options_test )
+            allow( Beaker::Options::HostsFileParser ).to receive(
+              :parse_hosts_file
+            ).and_raise( Errno::ENOENT )
+
+            mock_beaker_hostgenerator_cli = Object.new
+            allow( mock_beaker_hostgenerator_cli ).to receive( :execute )
+            allow( BeakerHostGenerator::CLI ).to receive(
+              :new
+            ).and_raise( BeakerHostGenerator::Exceptions::InvalidNodeSpecError )
```
nwolfe: Does `BeakerHostGenerator::CLI.new` raise errors?  I don't think it should; `execute` is when the parsing occurs.

lib/beaker/options/parser.rb
```diff
@@ -228,10 +228,23 @@ def parse_hosts_options
         hosts_options = Beaker::Options::HostsFileParser.parse_hosts_file(@options[:hosts_file])
         return hosts_options
       rescue Errno::ENOENT
+        error_message = "\nHosts file '#{@options[:hosts_file]}' does not exist."
+        error_message << "\nTrying as beaker-hostgenerator input.\n\n"
+        $stderr.puts error_message
         require 'tempfile'
```
nwolfe: I know you didn't change this line in this PR, but is this require still necessary if we're not writing to disk anymore?

tvpartytonight:
👍 Is there any use in introducing the logger object earlier so we don't have to just put to the iostreams? Doesn't block merging this ticket at all, but, it might be worth doing in the future?

lib/beaker/options/parser.rb
```diff
@@ -228,10 +228,22 @@ def parse_hosts_options
...
+
+        hosts_file_content = begin
+          bhg_cli = BeakerHostGenerator::CLI.new( [ @options[:hosts_file] ] )
+          bhg_cli.execute
+        rescue BeakerHostGenerator::Exceptions::Error,
+          BeakerHostGenerator::Exceptions::InvalidNodeSpecError => error
+          error_message = "\nbeaker-hostgenerator was not able to use this value as input."
+          error_message << "\nExiting with an Error.\n\n"
+          $stderr.puts error_message
+          raise error
```
nwolfe: Is it common in Beaker code to print to STDERR *and* raise an exception?

How did you decide on doing both?  Not a problem, just curious if there's some funky reason for this.  As a library you usually take greater care over printing/logging/errors because you want to cede control to the application.

lib/beaker/options/parser.rb
```diff
@@ -228,10 +228,22 @@ def parse_hosts_options
         hosts_options = Beaker::Options::HostsFileParser.parse_hosts_file(@options[:hosts_file])
         return hosts_options
       rescue Errno::ENOENT
-        require 'tempfile'
+        error_message = "\nHosts file '#{@options[:hosts_file]}' does not exist."
+        error_message << "\nTrying as beaker-hostgenerator input.\n\n"
+        $stderr.puts error_message
```
nwolfe: This isn't an error that it's not a hosts file, so printing to stderr seems wrong.  This is a supported feature, and normal behavior, so stdout seems appropriate.

lib/beaker/options/parser.rb
nwolfe: I like @tvpartytonight's comment about getting the logger available in here.

We should look into the dependencies between logger and options and see if we can't tease these things apart.

kevpl: Looked into the [dependencies between these two](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/cli.rb#L14-L16). Logger directly depends on the options to determine how it should be setup.

The option that comes to mind would be to create a logger specifically for the phase before the "real logger" is configured post options parsing. However, that doesn't seem like a good option to me, as it's an unconfigurable logger. At least this way, the user has the standard control mechanisms for the basic OS output streams.

For your comment about both printing to standard error and raising an exception, I decided to have it that way to give the most information possible. I'm not familiar with anyone using beaker as a library, in my experience it's just an application. The printing is to help someone get the context that they need to be able to deal with the error message. This is particularly helpful if they weren't intending to send the string to beaker-hostgenerator, as I've seen happen already with people using the master branch.

lib/beaker/options/parser.rb
kevpl: At this point, we can't know if this is going to be an error or not. There's only one code path that runs through these print statements where no error has occurred (correct beaker-hostgenerator syntax that was intended). Since there are a number of other pathways through here, and those are all error cases (examples below), I believe using standard error here is correct.

- host file path improperly specified
- beaker-hostgenerator syntax incorrect

One other option we could use here is to wait and not print until we know a little more. Since if the beaker-hostgenerator use succeeds, we don't typically print out at all, it seems that just deleting this might be a better option. However, my concerns with this are that:

1. If you checkout the logging/printing that happens above in my snippet in the first comment, you'll see that a beaker-hostgenerator warning is issued. Without the printing done here, there's no context for that warning, and beaker users will be confused about where this is coming from and why it's there without the additional printing.
2. If there's either something wrong with the importing mechanism (doubtful) or with the way beaker initializes and uses beaker-hostgenerator (not as surprising), then this will fail with a potentially uncaught exception type, with no output that even let the beaker user know they were using a different application to generate their hosts file at all.

I'm of the opinion that this should be kept. @nwolfe what do you think?

nwolfe: If you think about the feature we're introducing and not how the code is written, it's not correct to say that using BHG is an error.  It's not.  The feature we're introducing is the ability to have `--hosts` be a file or a BHG string, so why would we print a stdout message for one and a stderr message for the other?  

They're both equally valid and formally supported features, so they should just be logged like normal behavior; not an error.

Regarding the implementation, why wouldn't we do a `File.exists(@options[:hosts_file])` first to determine which code path we're going down?  That would be better than using exception for normal flow-of-control.

If we're forced to keep it wrapped in a try/catch like this, I still don't see why we would print to stderr yet.  It makes sense to print to stderr if the subsequent call to BHG fails.

lib/beaker/options/parser.rb
nwolfe: > I'm not familiar with anyone using beaker as a library, in my experience it's just an application.

Interesting. I consider it a library, and I feel like others do to.  We should probably be on the same page about this 😄

tvpartytonight:

tvpartytonight:
I'm a 👍  on this improved implementation.

PR #1148, opened 2016-06-16
(BKR-847) Support puppet.com in forge stubs

johnduarte:
This commit adds an `aliases` parameter to the `stub_hosts_on`
and the `with_host_stubbed_on` to facilitate adding multiple host aliases when
stubbing hosts.

This commit also updates the `stub_forge_on` and `with_forge_stubbed_on`
methods to add the `puppet.com` and `puppetlabs.com` variants of `forge`
and `forgeapi` in order to support both domain names.

### DIFF
--- a/lib/beaker/dsl/helpers/puppet_helpers.rb
+++ b/lib/beaker/dsl/helpers/puppet_helpers.rb
@@ -505,13 +505,22 @@ module Beaker
         #                            or a role (String or Symbol) that identifies one or more hosts.
         # @param ip_spec [Hash{String=>String}] a hash containing the host to ip
         #   mappings
-        # @example Stub puppetlabs.com on the master to 127.0.0.1
-        #   stub_hosts_on(master, 'puppetlabs.com' => '127.0.0.1')
-        def stub_hosts_on(machine, ip_spec)
+        # @param alias_spec [Hash{String=>Array[String]] an hash containing the host to alias(es) mappings to apply
+        # @example Stub puppetlabs.com on the master to 127.0.0.1 with an alias example.com
+        #   stub_hosts_on(master, {'puppetlabs.com' => '127.0.0.1'}, {'puppetlabs.com' => ['example.com']})
+        def stub_hosts_on(machine, ip_spec, alias_spec={})
           block_on machine do | host |
             ip_spec.each do |address, ip|
+              aliases = alias_spec[address] || []
+              manifest =<<-EOS.gsub /^\s+/, ""
+                host { '#{address}':
+                  \tensure       => present,
+                  \tip           => '#{ip}',
+                  \thost_aliases => #{aliases},
+                }
+              EOS
               logger.notify("Stubbing address #{address} to IP #{ip} on machine #{host}")
-              on( host, puppet('resource', 'host', address, 'ensure=present', "ip=#{ip}") )
+              apply_manifest_on( host, manifest )
             end
 
             teardown do
@@ -530,17 +539,15 @@ module Beaker
         #                            or a role (String or Symbol) that identifies one or more hosts.
         # @param ip_spec [Hash{String=>String}] a hash containing the host to ip
         #   mappings
-        # @example Stub puppetlabs.com on the master to 127.0.0.1
-        #   with_host_stubbed_on(master, 'forgeapi.puppetlabs.com' => '127.0.0.1') do
+        # @param alias_spec [Hash{String=>Array[String]] an hash containing the host to alias(es) mappings to apply
+        # @example Stub forgeapi.puppetlabs.com on the master to 127.0.0.1 with an alias forgeapi.example.com
+        #   with_host_stubbed_on(master, {'forgeapi.puppetlabs.com' => '127.0.0.1'}, {'forgeapi.puppetlabs.com' => ['forgeapi.example.com']}) do
         #     puppet( "module install puppetlabs-stdlib" )
         #   end
-        def with_host_stubbed_on(host, ip_spec, &block)
+        def with_host_stubbed_on(host, ip_spec, alias_spec={}, &block)
           begin
             block_on host do |host|
-              ip_spec.each_pair do |address, ip|
-                logger.notify("Stubbing address #{address} to IP #{ip} on machine #{host}")
-                on( host, puppet('resource', 'host', address, 'ensure=present', "ip=#{ip}") )
-              end
+              stub_hosts_on(host, ip_spec, alias_spec)
             end
 
             block.call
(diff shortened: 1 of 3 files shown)


kevpl:

johnduarte:
Thanks @kevpl . I was able to grok rspec for this without pestering you. :-)

lib/beaker/dsl/helpers/puppet_helpers.rb
```diff
@@ -504,13 +504,21 @@ def run_agent_on(host, arg='--no-daemonize --verbose --onetime --test',
         #                            or a role (String or Symbol) that identifies one or more hosts.
         # @param ip_spec [Hash{String=>String}] a hash containing the host to ip
         #   mappings
+        # @param aliases Array[String] an array containing the host alias to apply ip mappings
```
tvpartytonight: Nit: this should say aliases instead of alias?

johnduarte: Good eye @tvpartytonight ! I will submit an update.

lib/beaker/dsl/helpers/puppet_helpers.rb
```diff
@@ -504,13 +504,21 @@ def run_agent_on(host, arg='--no-daemonize --verbose --onetime --test',
...
         #   stub_hosts_on(master, 'puppetlabs.com' => '127.0.0.1')
-        def stub_hosts_on(machine, ip_spec)
+        def stub_hosts_on(machine, ip_spec, aliases=[])
           block_on machine do | host |
             ip_spec.each do |address, ip|
+              manifest =<<-EOS.gsub /^\s+/, ""
+                host { '#{address}':
+                  \tensure       => present,
+                  \tip           => '#{ip}',
+                  \thost_aliases => #{aliases},
```
tvpartytonight: Will this add the same aliases for each `address` and `ip` for each pair? That doesn't seem right...

johnduarte: Hmm. I am not clear on how to resolve this. If I am interpreting the declared interface in the doc correctly, it is only expecting a single host/ip map. Obviously, however the method supports a longer hash.
In order to map the aliases to a specific host in the hash we would need to change the signature of the hash argument. This would be a braking change to these methods.

Any suggestions?

johnduarte: @tvpartytonight take a look now. I have reimplemented the `aliases` array as a `alias_spec` hash to map to the `address` key in the `ip_spec` hash in order to allow for submitting multiple mappings to this method.

lib/beaker/dsl/helpers/puppet_helpers.rb
```diff
@@ -504,13 +504,22 @@ def run_agent_on(host, arg='--no-daemonize --verbose --onetime --test',
...
+                }
+              EOS
               logger.notify("Stubbing address #{address} to IP #{ip} on machine #{host}")
-              on( host, puppet('resource', 'host', address, 'ensure=present', "ip=#{ip}") )
+              apply_manifest_on( host, manifest )
```
kevpl: It looks like this `apply_manifest_on` call is happening once per `ip_spec`, rather than once per host as below.

It looks like this work of creating & applying the manifest is duplicate code. Could you make a helper method that could be called from both of these locations?  Could this happen with the teardown too?

johnduarte: @kevpl it seems like the `stub_hosts_on` is method that should be wrapped in `with_host_stubbed_on`. I am updating accordingly.

lib/beaker/dsl/helpers/puppet_helpers.rb
```diff
@@ -529,17 +538,15 @@ def stub_hosts_on(machine, ip_spec)
...
           begin
             block_on host do |host|
-              ip_spec.each_pair do |address, ip|
-                logger.notify("Stubbing address #{address} to IP #{ip} on machine #{host}")
-                on( host, puppet('resource', 'host', address, 'ensure=present', "ip=#{ip}") )
-              end
+              stub_hosts_on(host, ip_spec, aliases_spec)
             end
 
             block.call
```
kevpl: @johnduarte I have one question about teardown behavior here. Now that `with_host_stubbed_on` is calling `stub_hosts_on`, this makes me wonder about the teardown behavior.

When this method calls `stub_hosts_on`, a teardown will be added to the test, so this is an operation that happens over the scope of one entire test. However, `with_host_stubbed_on` is for passing a block to be executed, which assumes that these host stubbings will be cleaned up at the end of the block, and you can run code after the block that won't be affected by these stubs. Does the code below clean up the aliases correctly, so that block cleanup happens correctly, or do we have to wait for the test teardown procedure to fully remove these?

johnduarte: @kevpl this is a great question.

It looks like the `with_host_stubbed_on` is only used by `with_forge_stubbed_on` and `with_forge_stubbed`. This should be tested against a repo using these methods, but at a cursory glance I am unable to find and example of these being used. @phongdly I have searched a few module repos and not found an example of these methods being used (firewall, postfix, puppetdb, vcsrepo). Are you aware of an example where these methods are used?

phongdly: @johnduarte I don't think module code still have it anymore, _stub_forge_on(master)_ had been used before and then was replaced by _install_dev_puppet_module_on_ as in the last line of this commit: https://github.com/puppetlabs/puppetlabs-vsphere/commit/de2b3bc11ecfec7c26564dd33bcc503ff7dc603c

kevpl: @johnduarte it seems like this would be a good place to put an acceptance test in to verify this behavior.

kevpl:
last comment is current

kevpl:
last comment still current

johnduarte:
@kevpl I have added an acceptance test to verify the teardown behavior of the `stub_host_on` and `with_stub_host_on` methods. Thanks.

kevpl:
@johnduarte looks like this is failing in escaping the windows path separator enough. Adding `\\` to the windows hosts file path fixes the issue in my local testing of this.

johnduarte:
Thanks @kevpl . Not sure why it worked for me locally before. Took me a bit to discover how to run this test as per the Jenkins job. It is now passing for me locally. I will keep my fingers crossed for the CI run.

kevpl:
spec failed, but it looks like something @tvpartytonight and I were talking about earlier. @tvpartytonight, should we just re-kick this?

tvpartytonight:
@johnduarte can rebase off of master and push? I suspect you will need the changes I made to master in order to successfully pass spec testing in 1.9.3.

johnduarte:
@tvpartytonight I have rebased this on master at 7df594b

tvpartytonight:

PR #1149, opened 2016-06-17
(BKR-843) Use hypervisor specific logic for ip

tvpartytonight:
When determining the ip address of a machine, beaker normally ssh'd into
the box and would run a command that would return the ip configured on
that box. However, this doesn't work for several cloud hypervisors, as
the public ip of the box may not match the internally configured ip
address. This change allows for hypervisor specific logic to be executed
instead of sshing into the box. Specifically, this commit changes all
ec2 instances to use the public ip address instead of the private one.

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -228,7 +228,15 @@ module Beaker
     def get_public_ip
       case host_hash[:hypervisor]
       when 'ec2'
-        host_hash[:instance].ip_address
+        if host_hash[:instance]
+          host_hash[:instance].ip_address
+        else
+          # In the case of using ec2 instances with the --no-provision flag, the ec2
+          # instance object does not exist and we should just use the curl endpoint
+          # specified here:
+          # http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-instance-addressing.html
+          execute("curl http://169.254.169.254/latest/meta-data/public-ipv4").strip
+        end
       end
     end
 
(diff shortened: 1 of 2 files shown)


tvpartytonight:
There may be an issue in this implementation because I think that the `aws` object won't exist if `--no-provision` is set to true, but it's still ready for review. I'll investigate the `--no-provision` issue and report back.

tvpartytonight:
This is now in a mergeable state. 📃

tvpartytonight:

tvpartytonight:

doug-rosser:
I ❤️  this so much!

pinkypie:
👍

lib/beaker/host.rb
```diff
@@ -224,10 +224,26 @@ def get_ip
...
+      case host_hash[:hypervisor]
+      when 'ec2'
+        if host_hash[:instance]
+          host_hash[:instance].ip_address
+        else
+          # In the case of using ec2 instances with the --no-provision flag, the ec2
+          # instance object does not exist and we should just use the curl endpoint
+          # specified here:
+          # http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-instance-addressing.html
+          execute("curl http://169.254.169.254/latest/meta-data/public-ipv4").strip
```
kevpl: @tvpartytonight this doesn't need to be parameterized at all, or configurable, even for outside users?

tvpartytonight: Not from what I can tell @kevpl, but I just realized that amazon documents using `wget` for windows boxes; I should add that in now too.

tvpartytonight: This has been added.

PR #1161, opened 2016-06-24
(BKR-856) Add el/sles support to remove_puppet_on

johnduarte:
This commit expands platform support for the `remove_puppet_on` install
helper method to include `sles` and all `el` derivatives.

This allows hosts declared with a `hypervisor` of `none` on these
platforms to have puppet uninstalled as part of a pre-suite.

### DIFF
--- a/spec/beaker/dsl/install_utils/foss_utils_spec.rb
+++ b/spec/beaker/dsl/install_utils/foss_utils_spec.rb
@@ -1110,68 +1110,45 @@ describe ClassMixedWithDSLInstallUtils do
   end
 
   describe '#remove_puppet_on' do
-    let(:aixhost) { make_host('aix', :platform => 'aix-53-power') }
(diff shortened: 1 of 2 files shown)


spec/beaker/dsl/install_utils/foss_utils_spec.rb
```diff
@@ -1131,6 +1139,110 @@ def test_fetch_http_file_no_ending_slash(platform)
       subject.remove_puppet_on( aixhost )
     end
 
+    it 'uninstalls packages on el' do
```
tvpartytonight: @johnduarte can have these uninstall tests run through an array of hosts, rather than copying the test for each platform it now supports?

kevpl:
@tvpartytonight have all your concerns been addressed? looks good to me. 👍

tvpartytonight:
yup, looks good to me. 👍

PR #1162, opened 2016-06-24
Revert "(BKR-568) no longer accept PRs marked as "(MAINT)""

kevpl:
Reverts puppetlabs/beaker#974

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 ## Getting Started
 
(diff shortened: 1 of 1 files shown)


samuelson:
@kevpl how about adding a tag (DOC) or something for documentation improvements?

kevpl:
@samuelson I'd rather have something general and have us deal with the ambiguities rather than needing a DOC, and then others for any other kind of MAINT fixes we might need.

samuelson:
@kevpl that's a good point.

PR #1164, opened 2016-06-28
(BKR-40) Fix UTF8 buffers for SSH and SCP

Iristyle:
- Prior to this change, attempting to send UTF8 commands through
  SSH, or attempting to copy files with UTF8 filenames could fail.
  This was particularly easy to trigger by attempting to execute
  commands that were 128 bytes or longer.
- monkey patch net-ssh gem to allow UTF-8 strings >= 128 bytes
  
  The buffer @content is often built as a UTF-8 string, until the
  point at which it appends data that cannot be encoded as a UTF-8
  sequence.
  
  One case occurs when the call to write_string is made to append a
  string that exceeds 127 bytes in length.  The SSH2 format says
  that strings must be length prefixed, and when the value [128]
  has pack("N*") called against it, the resultant 4 byte network
  order representation does not have a valid UTF-8 equivalent,
  resulting in an ASCII-8BIT / BINARY string.
  
  [127].pack('N*').encode('utf-8')
  => "\u0000\u0000\u0000\u007F"
  
  [128].pack('N*').encode('utf-8')
  Encoding::UndefinedConversionError: "\x80" from ASCII-8BIT to UTF-8
  
  Ruby has a subtle behavior where appending a BINARY string to
  an existing UTF-8 string is allowed and the resultant string
  changes encoding to BINARY.  However, once this has happened,
  the string can no longer have UTF-8 encoded strings appended as
  Ruby will raise an Encoding:CompatibilityError
  
  Appending BINARY to UTF-8 always creates BINARY:
  "foo".encode('utf-8') << [128].pack('N*')
  => "foo\x00\x00\x00\x80"
  
  Appending UTF-8 representable strings to existing strings:
  
  Ruby 2.1.7 keeps the string as its default UTF-8
  "foo" << [127].pack('N*')
  => "foo\u0000\u0000\u0000\u007F"
  
  Ruby 1.9.3 keeps UTF-8 strings as UTF-8
  "foo".encode('utf-8') << [127].pack('N*')
  => "foo\u0000\u0000\u0000\u007F"
  
  Ruby 1.9.3 defaults to US-ASCII which changes it to BINARY
  pry(main)> "foo" << [127].pack('N*')
  => "foo\x00\x00\x00\x7F"
  
  The simple solution is to call force_encoding on UTF-8 strings
  prior to appending them to @content, given it's always OK to
  append ASCII-8BIT / BINARY strings to existing strings, but
  appending UTF-8 to BINARY raises errors.
  
  "\x80".force_encoding('ASCII-8BIT') << "\u16A0"
  Encoding::CompatibilityError: incompatible character encodings: ASCII-8BIT and UTF-8
  
  force_encoding in this case, will simply translate a valid UTF-8
  string to its BINARY equivalent
  
  "\u16A0".force_encoding('BINARY')
  => "\xE1\x9A\xA0"
  Correct conversion per http://www.fileformat.info/info/unicode/char/16a0/index.htm

### DIFF
--- a/lib/beaker/ssh_connection.rb
+++ b/lib/beaker/ssh_connection.rb
@@ -1,6 +1,53 @@
 require 'socket'
 require 'timeout'
 require 'net/scp'
+require 'net/ssh'
+
+module Net
+  module SSH
+    class Buffer
+      # @private
+      # monkey patch net-ssh gem to allow UTF-8 strings >= 128 bytes
+      # \@content is often built as a UTF-8 string, until the point at
+      # which it appends data that cannot be encoded as a UTF-8 sequence.
+      #
+      # One case occurs when the call to write_string is made to append a
+      # string that exceeds 127 bytes in length.  The SSH2 format says
+      # that strings must be length prefixed, and when the value [128]
+      # has pack("N*") called against it, the resultant 4 byte network
+      # order representation does not have a valid UTF-8 equivalent,
+      # resulting in an ASCII-8BIT / BINARY string.
+      #
+      # [127].pack('N*').encode('utf-8')
+      # => "\u0000\u0000\u0000\u007F"
+      #
+      # [128].pack('N*').encode('utf-8')
+      # Encoding::UndefinedConversionError: "\x80" from ASCII-8BIT to UTF-8
+      #
+      # Ruby has a subtle behavior where appending a BINARY string to
+      # an existing UTF-8 string is allowed and the resultant string
+      # changes encoding to BINARY.  However, once this has happened,
+      # the string can no longer have UTF-8 encoded strings appended as
+      # Ruby will raise an Encoding:CompatibilityError
+      #
+      # The simple solution is to call force_encoding on UTF-8 strings
+      # prior to appending them to @content, given it's always OK to
+      # append ASCII-8BIT / BINARY strings to existing strings, but
+      # appending UTF-8 to BINARY raises errors.
+      #
+      # force_encoding in this case, will simply translate a valid UTF-8
+      # string to its BINARY equivalent
+      #
+      # "\u16A0".force_encoding('BINARY')
+      # => "\xE1\x9A\xA0"
+      # Correct conversion per http://www.fileformat.info/info/unicode/char/16a0/index.htm
+      def write(*data)
+        data.each { |datum| @content << datum.frozen? ? datum.encode('BINARY') : datum.force_encoding('BINARY') }
+        self
+      end
+    end
+  end
+end
 
 module Beaker
   class SshConnection
@@ -108,7 +155,7 @@ module Beaker
 
     # Wait for the ssh connection to fail, returns true on connection failure and false otherwise
     # @param [Hash{Symbol=>String}] options Options hash to control method conditionals
-    # @option options [Boolean] :pty Should we request a terminal when attempting 
+    # @option options [Boolean] :pty Should we request a terminal when attempting
     #                                to send a command over this connection?
     # @option options [String] :stdin Any input to be sent along with the command
     # @param [IO] stdout_callback An IO stream to send connection stdout to, defaults to nil
--- a/spec/beaker/ssh_connection_spec.rb
+++ b/spec/beaker/ssh_connection_spec.rb
@@ -1,6 +1,22 @@
 require 'spec_helper'
 require 'net/ssh'
 
+module Net
+  module SSH
+    describe Buffer do
+      it 'has a monkey patch tested on 2.9.4' do
+        # simple verification that the patch is understood for this particular version
+        # and might be removable in future versions
+        expect(Gem.loaded_specs['net-ssh'].version).to eq(Gem::Version.new('2.9.4'))
+      end
+
+      it 'can build up a buffer properly with a UTF8 string of 128 bytes or longer' do
+        expect { described_class.from(:string, ("\u16A0" * 128)) }.to_not raise_error
+      end
+    end
+  end
+end
+
 module Beaker
   describe SshConnection do
     let( :user )      { 'root'    }


spec/beaker/ssh_connection_spec.rb
```diff
@@ -1,6 +1,22 @@
 require 'spec_helper'
 require 'net/ssh'
 
+module Net
+  module SSH
+    describe Buffer do
+      it 'has a monkey patch tested on 2.9.4' do
```
Iristyle: This is a guard to make sure we notice when a new version releases - requiring us to verify the patch still makes sense.

kevpl:
@Iristyle it looks like the tests are failing yard doc generation

kevpl:
last comment still current

kevpl:
last comment still current

Iristyle:
@kevpl / @tvpartytonight - 

Updated with a silly escape of the `@content` in the comments as `\@content`.  Also added `@private`, even though we're not passing `--no-private` to YARD doc generation from what I can see.

Iristyle:
FYI - I have opened 4 PRs against `net-ssh` itself to address this problem - 

* 4.x alpha branch - https://github.com/net-ssh/net-ssh/pull/394
* 3.1 branch - https://github.com/net-ssh/net-ssh/pull/397
* 2.10 branch - https://github.com/net-ssh/net-ssh/pull/396
* 2.9 branch - https://github.com/net-ssh/net-ssh/pull/395

lib/beaker/ssh_connection.rb
```diff
@@ -1,6 +1,53 @@
...
+      # appending UTF-8 to BINARY raises errors.
+      #
+      # force_encoding in this case, will simply translate a valid UTF-8
+      # string to its BINARY equivalent
+      #
+      # "\u16A0".force_encoding('BINARY')
+      # => "\xE1\x9A\xA0"
+      # Correct conversion per http://www.fileformat.info/info/unicode/char/16a0/index.htm
+      def write(*data)
+        data.each { |datum| @content << datum.frozen? ? datum.encode('BINARY') : datum.force_encoding('BINARY') }
```
Iristyle: This will copy strings when frozen, or change their encoding metadata when not frozen.

kevpl:
last comment is current

kevpl:
last comment is current.

kevpl:
last comment is current. If this hasn't moved forward in a week, we'll close this due to inactivity.

kevpl:
Still failing spec tests:
```

Failures:

  1) Net::SSH::Buffer can build up a buffer properly with a UTF8 string of 128 bytes or longer
     Failure/Error: expect { described_class.from(:string, ("\u16A0" * 128)) }.to_not raise_error
     
       expected no Exception, got #<TypeError: no implicit conversion of false into String> with backtrace:
         # ./lib/beaker/ssh_connection.rb:45:in `block in write'
         # ./lib/beaker/ssh_connection.rb:45:in `each'
         # ./lib/beaker/ssh_connection.rb:45:in `write'
         # ./vendor/bundle/ruby/2.2.0/gems/net-ssh-2.9.4/lib/net/ssh/buffer.rb:322:in `block in write_string'
         # ./vendor/bundle/ruby/2.2.0/gems/net-ssh-2.9.4/lib/net/ssh/buffer.rb:319:in `each'
         # ./vendor/bundle/ruby/2.2.0/gems/net-ssh-2.9.4/lib/net/ssh/buffer.rb:319:in `write_string'
         # ./vendor/bundle/ruby/2.2.0/gems/net-ssh-2.9.4/lib/net/ssh/buffer.rb:55:in `block in from'
         # ./vendor/bundle/ruby/2.2.0/gems/net-ssh-2.9.4/lib/net/ssh/buffer.rb:47:in `step'
         # ./vendor/bundle/ruby/2.2.0/gems/net-ssh-2.9.4/lib/net/ssh/buffer.rb:47:in `from'
         # ./spec/beaker/ssh_connection_spec.rb:14:in `block (3 levels) in <module:SSH>'
         # ./spec/beaker/ssh_connection_spec.rb:14:in `block (2 levels) in <module:SSH>'
     # ./spec/beaker/ssh_connection_spec.rb:14:in `block (2 levels) in <module:SSH>'

Finished in 24.16 seconds (files took 3.27 seconds to load)
1707 examples, 1 failure
```

lib/beaker/ssh_connection.rb
samwoods1: I'm a little bit confused here... what causes the data here to be frozen or not frozen?  Why is that used to determine whether to change the byte representation of the string or not?

spec/beaker/ssh_connection_spec.rb
```diff
@@ -1,6 +1,22 @@
...
+        # simple verification that the patch is understood for this particular version
+        # and might be removable in future versions
+        expect(Gem.loaded_specs['net-ssh'].version).to eq(Gem::Version.new('2.9.4'))
+      end
+
+      it 'can build up a buffer properly with a UTF8 string of 128 bytes or longer' do
+        expect { described_class.from(:string, ("\u16A0" * 128)) }.to_not raise_error
+      end
```
samwoods1: I think a test that validates frozen vs not frozen data would be useful.

kevpl:
last comment is still current

lib/beaker/ssh_connection.rb
Iristyle: If the values are Ruby constants, then they might be frozen.  `.encode` creates a copy of the original string with the specified  encoding, while `.force_encoding` changes the current strings metadata so that the string is treated as a `BINARY` in this case.

Iristyle:
I haven't updated this yet, because my patch + some extra code landed in the upstream `net-ssh` project at https://github.com/net-ssh/net-ssh/pull/407 that would make this monkey patch unnecessary (which is preferred).

Unfortunately, the patch landed on the master branch of `net-ssh`, which only targets the 4.0 (currently alpha series).  I opened up another PR at https://github.com/net-ssh/net-ssh/pull/434 to backport those commits to the `3.x` branch, so that Beaker should automatically get the updated gem when it's next published (assuming it includes that fix)

Iristyle:
Of note that one way to trigger this bug easily is to modify the test at https://github.com/puppetlabs/puppet/blob/master/acceptance/tests/resource/scheduled_task/should_modify.rb and change:

```ruby
name = "pl#{rand(999999).to_i}"
```

To include some UTF-8 characters in the name (or anywhere else in the string)

```ruby
name = "plᚠᛇᚻ#{rand(999999).to_i}"
```

If the task has the correct name, then the bug has been fixed.  If it does not, the bug is still present.

kevpl:
@Iristyle if we're going to wait on determining if you're going to do work on this PR to make it ready based on the outcome of your https://github.com/net-ssh/net-ssh/pull/434 PR, is there a reason to keep this PR open in the meantime? It seems to me that that open PR should be enough to track this effort, and that keeping this open is redundant.

If there's no comment made about this in a week, I'll close this PR for now.

PR #1166, opened 2016-07-04
Fix links to hypervisors

traylenator:
All the particular hypervisor docs are now in
a sub directory.


kevpl:
Thanks @traylenator for the fix! In the future, if you prefix your git commit message with `(MAINT)` (for doc improvements) then we'll be able to merge the commit as-is. For now, I've fixed it in a squash-merge, so no problem. Thanks again!

PR #1168, opened 2016-07-05
(BKR-797) timesync helper should handle ntpd already running

james-stocks:
This commit fixes a bug where timesync causes beaker to error if the host has ntpd already running.
The fix is simply to use the -u option for ntpdate to use an alternate network port.

Tested on Redhat and Ubuntu.


james-stocks:
Closing as tests need fixed

PR #1175, opened 2016-07-09
BKR-874 PUPPET_MODULE_INSTALL_IGNORE

apollo-io:
Updated the filtering on PUPPET_MODULE_INSTALL_IGNORE to exclude .svn,
junit, pkg and example folders

Without this change, unnecessary files are copied to the SUT which slows down test execution time.  

The changes filters out build generated files as well, created when using puppetlabs_spec_helper.


kevpl:

PR #1176, opened 2016-07-11
(BKR-859) Add rake quick start tasks

pinkypie:
For generating hosts file, test suite and running test (on vmpooler).
Needed to make beaker-hostgenerator a runtime dependency in order for gen_hosts task to work.

As it stands, I think it's useful to be able to generate default host file and test files for a new user. But not sure how useful the run task would be as they would probably need to edit the files before they kicked off a test run.

Not sure about the tasks as they stand. Some questions (@kevpl @tvpartytonight) :
Should we make the gen_hosts tasks configurable so a different host config string can be passed in?
Should we make the run task configurable so that it can be used to run different pre-suite/tests?

Or do we keep this as simple as possible, as a way to show the workflow to a new user? And address the other issues if we decide to work on the subcommand functionality at a later stage.

### DIFF
--- a/lib/beaker/tasks/quick_start.rb
+++ b/lib/beaker/tasks/quick_start.rb
@@ -1,0 +1,49 @@
+require 'beaker-hostgenerator'
+
+namespace :beaker_quickstart do
+
+  desc 'Generate Default Beaker Host Config File'
+  task :gen_hosts do
+    cli = BeakerHostGenerator::CLI.new(['redhat7-64default.mdcal-redhat7-64af'])
+    FileUtils.mkdir_p('acceptance/config') # -p ignores when dir already exists
+    File.open("acceptance/config/default_hosts.yaml", 'w') do |fh|
+      fh.print(cli.execute)
+    end
+  end
+
+  desc 'Generate Default Pre-Suite'
+  task :gen_pre_suite do
+    FileUtils.mkdir_p('acceptance/setup') # -p ignores when dir already exists
+    File.open("acceptance/setup/default_pre_suite.rb", 'w') do |fh|
+      fh.print('install_pe')
+    end
+  end
+
+  desc 'Generate Default Smoke Test'
+  task :gen_smoke_test do
+    FileUtils.mkdir_p('acceptance/tests') # -p ignores when dir already exists
+    File.open("acceptance/tests/default_smoke_test.rb", 'w') do |fh|
+      fh.print("test_name 'puppet install smoketest'
+step 'puppet install smoketest: verify \\'puppet help\\' can be successfully called on
+all hosts'
+    hosts.each do |host|
+      on host, puppet('help')
+    end")
+    end
+  end
+
+  desc 'Run Default Smoke Test'
+  task :run => ['beaker_quickstart:gen_hosts', 'beaker_quickstart:gen_pre_suite', 'beaker_quickstart:gen_smoke_test'] do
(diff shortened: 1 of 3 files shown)


lib/beaker/tasks/quick_start.rb
```diff
@@ -0,0 +1,49 @@
+require 'beaker-hostgenerator'
+
+namespace :beaker_quickstart do
```
kevpl: Does this need to be prefixed with `beaker_`? I suspect that this is redundant.

lib/beaker/tasks/quick_start.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  desc 'Generate Default Beaker Host Config File'
+  task :gen_hosts do
```
kevpl: maybe generation tasks should have their own namespace, since they're separate from running beaker itself?

kevpl:
My opinion is that keeping things as simple as possible is the best route here to get people going with the bare minimum. Once we have that, then we get into the harder question: what comes next? I think that's where the doc improvements that come along with the next task come in.

@tvpartytonight and I were discussing this change, and he raised the question of how a new user would use these rake tasks. I think I'd spaced on the nature of raketasks in dependent gems when specifying the ticket, but it looks like you're using a workflow similar to [this article](http://andyatkinson.com/blog/2014/06/23/sharing-rake-tasks-in-gems). Is this right?

lib/beaker/tasks/quick_start.rb
kevpl: nvm this comment. In light of the rake task sharing discussion below, I was mistaken about this being redundant.

pinkypie:
@kevpl it seemed the easiest way was how some people were already doing it in their project files in order have access to the existing rake tasks to run Beaker.

Basically you just:

require 'beaker/tasks/test' (name of file containing the tasks)

at the top of your local project rake file.

So that you inherit all the tasks in that file as well as any in the required file.

I have another doc PR to open today that explains how to use the current rake tasks, and will add to it, or create a new doc filed for the quickstart work.

lib/beaker/tasks/quick_start.rb
pinkypie: Ok, I was thinking that since they are all part of the quickstart workflow, it would make sense to have them in the same namespace? Maybe I should create a secondary level of namespaces to differentiate the generation tasks from the run task?

kevpl: The secondary level of namespaces would be nice to have, considering that there will be variants on some of the generation tasks in the follow-up work. But this is a nice-to-have, so it's not PR merge blocking.

kevpl:
cool, thanks for the explanation. I'd like to delay merging this until after the release, because there are changes I'd like for the docs PR, and I'd rather have them merge together than have the functionality released without docs. In the future, we want all work to include its docs, so it'd be better to put docs in the same PR in the future.

pinkypie:
Ok, I can include the docs for running rake tasks in this PR, but it was more for running the existing set of rake tasks (not the ones for the quickstart) so thought it should be separate (https://github.com/puppetlabs/beaker/pull/1180) I can do a separate docs file for the quickstart tasks and add it to this PR?

kevpl:
Ah, ok, I thought they were more closely related. Since that's not the case, I can merge this separately, and I can do that now.

PR #1171, opened 2016-07-06
(BKR-846) re-organized docs

kevpl:
[skip ci]

PR #1180, opened 2016-07-13
(BKR-883) Add docs for Beaker rake tasks

pinkypie:
Was really difficult to work out how to use the existing beaker rake tasks that are available to run Beaker tests via rake.
So added a doc with basic instructions.


kevpl:
Please update the doc location to work with the new folder structure going forward.

Since this is a "how to" for running our rake tasks, the appropriate path would be `docs/how_to/rake_tasks.md`

pinkypie:
@kevpl ah ok, I thought we were putting the docs in the same folder structure as what was in beaker/lib, to make them easier to find?

kevpl:
yeah, that's what we were doing some time ago, but that's really only easier to find if you know the code, so we've moved on to a different structure. If you check out the Tool Time trello board, there's a 'docs re-org' card with more information about the re-org. Your doc itself doesn't have to update for this, it just needs to match the new path naming. You can see the re-org itself in #1171.

pinkypie:
@kevpl cool, should be in right location now.

kevpl:
for future doc changes, you can just add 

    [skip ci]

in your git commits to make sure they don't waste CI time.  Also, yay for squashing! Have a good weekend!

PR #1181, opened 2016-07-13
(maint) Pin gems that removed ruby 1.9.3 support

tvpartytonight:
This PR pins gems that are transitive dependencies that dropped ruby
1.9.3 support. This should be removed eventually, as these gems are not
actually needed by beaker itself, but by gems beaker relies on.


kevpl:

PR #1183, opened 2016-07-14
(BKR-884) Fix issue in run_block_on

pinkypie:
Need to check if opts[:run_in_parallel] == true.
If it evaluates to [], it will go into the if statement as it stands.

### DIFF
--- a/lib/beaker/shared/host_manager.rb
+++ b/lib/beaker/shared/host_manager.rb
@@ -102,7 +102,7 @@ module Beaker
         end
         if block_hosts.is_a? Array
           if block_hosts.length > 0
-            if opts[:run_in_parallel]
+            if opts[:run_in_parallel] == true
               # Pass caller[1] - the line that called block_on - for logging purposes.
               result = block_hosts.map.each_in_parallel(caller[1]) do |h|
                 run_block_on h, &block
--- a/spec/beaker/shared/host_manager_spec.rb
+++ b/spec/beaker/shared/host_manager_spec.rb
@@ -163,6 +163,14 @@ module Beaker
           }
         end
 
+        it "will ignore run_in_parallel global option" do
+          myhosts = host_handler.run_block_on( hosts, nil, { :run_in_parallel => [] } ) do |host|
+            host
+          end
+          expect( InParallel::InParallelExecutor ).not_to receive(:_execute_in_parallel).with(any_args)
+          expect(myhosts).to eq(hosts)
+        end
+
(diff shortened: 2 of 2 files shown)


lib/beaker/shared/host_manager.rb
```diff
@@ -102,7 +102,7 @@ def run_block_on hosts = [], filter = nil, opts = {}, &block
         end
         if block_hosts.is_a? Array
           if block_hosts.length > 0
-            if opts[:run_in_parallel]
+            if opts[:run_in_parallel] == true
```
nwolfe: Can we instead ensure that `opts[:run_in_parellel]` will always be a boolean?  Otherwise we'll have to account for the case when it's a non-boolean value in multiple places, such as any/every other `if` statement.

samwoods1: The only place where it won't be a boolean is if it is used as a global option where I'm already checking to ensure that it is an array and contains specific values... the option for run_block_on is the only place anywhere in Beaker where we are checking for a true/false value, so this should be ok.

samwoods1:
👍 If you want an associated unit test, you can steal from here: https://github.com/puppetlabs/beaker/pull/1184

pinkypie:
Thanks @samwoods1, added the test.

lib/beaker/shared/host_manager.rb
nwolfe: This value can be an array?  I was thinking it was a simple true/false toggle.

samwoods1: @nwolfe My original PR had 2 different options, one called install_in_parallel which was a global option, and one called run_in_parallel which was an option that could be passed into run_block_on.  Based on feedback, I changed the global install_in_parallel option to run_in_parallel with an array of what 'things' to run in parallel.  The 2 current options are configure and install.  In retrospect, naming the global option the same thing as the run_block_on option is confusing and can possibly have collisions like this...

samwoods1:
Hmm.... so, I had a chance to dig in here, I think there is actually a defect further up the chain... This change will make it so that you can't install FOSS in parallel...  I think this fix is still perfectly reasonable, but the full fix would require a change here: https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/install_utils/foss_utils.rb#L343-L345

tvpartytonight:
After talking it over with @samwoods1 it doesn't look there is actually an issue further upstream. The fix here in the `run_block_on` method is correct.

PR #1187, opened 2016-07-20
(BKR-890) Surface and provide more visibility into where failures occured

samwoods1:
BKR-890 - Print failure message and stack trace inline when a test failure occurs

### DIFF
--- a/lib/beaker/test_case.rb
+++ b/lib/beaker/test_case.rb
@@ -132,8 +132,7 @@ module Beaker
               test = File.read(path)
               eval test,nil,path,1
             rescue FailTest, TEST_EXCEPTION_CLASS => e
-              @test_status = :fail
-              @exception   = e
+              log_and_fail_test(e, :fail)
             rescue PendingTest
               @test_status = :pending
             rescue SkipTest
@@ -166,13 +165,14 @@ module Beaker
         # individually as well.
         #
         # @param exception [Exception] exception to fail with
-        def log_and_fail_test(exception)
+        # @param exception [Symbol] the test status
+        def log_and_fail_test(exception, status=:error)
           logger.error("#{exception.class}: #{exception.message}")
           bt = exception.backtrace
           logger.pretty_backtrace(bt).each_line do |line|
             logger.error(line)
           end
-          @test_status = :error
+          @test_status = status
           @exception   = exception
         end
       end
--- a/spec/beaker/test_case_spec.rb
+++ b/spec/beaker/test_case_spec.rb
@@ -49,7 +49,7 @@ module Beaker
           f.write "raise FailTest"
         end
         @path = path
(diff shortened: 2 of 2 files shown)


lib/beaker/test_suite.rb
```diff
@@ -141,7 +141,14 @@ def summarize(summary_logger)
       #@param [TestCase] test_case The {TestCase} to examine and print results for
       def print_test_result(test_case)
         test_reported = if test_case.exception
-                          "reported: #{test_case.exception.inspect}"
+                          test_file_trace = ""
```
kevpl: this block needs indent fixes. A block should only be indented 2 spaces from its parent code.

kevpl:
@samwoods1 is there any way to split these into smaller more focused PRs? It would be much cleaner if each ticket had its own PR.

For instance, it looks like these commits are pretty great at being their own "logically complete units" of implementation, but when I look closer, it appears that the commit for BKR-556 doesn't have a test included with it. This would be much easier to spot if it was its own PR.

With something as fundamental to beaker as the logger, I would much rather err on the side of over-testing than combining a bunch of work so that it goes through shorter test cycles.

samwoods1:
@kevpl Yeah, I'll split them up... As for the lack of unit test, I didn't find any existing tests for summarizing the results and wasn't quite sure how to go about that.

samwoods1:

samwoods1:

lib/beaker/shared/error_handler.rb
```diff
@@ -9,7 +9,7 @@ def report_and_raise(logger, e, msg)
       logger.pretty_backtrace(bt).each_line do |line|
         logger.error(line)
       end
-      raise e
+      raise e unless e.is_a? Beaker::DSL::FailTest
```
kevpl: @samwoods1 what are the consequences of this line of change? At first look, the rest of the PR seems unrelated to the `report_and_raise` method.

kevpl:
last comment is current

samwoods1:

lib/beaker/shared/error_handler.rb
samwoods1: You're right, I was playing with this, but thought I had removed it.. updated PR.

samwoods1:

er0ck:
needs tests

samwoods1:
@er0ck I believe the existing test that was modified covers it.  The behavior changed, and it now checks for the new behavior.  The existing :error test also ensures that I don't break the existing error logic.

kevpl:
@er0ck is there something in particular you're looking to be tested that's not covered?

er0ck:
fair.  i'd hoped this would increase some acceptance testing coverage as well, but upon further inspection i suppose this covers that case well enough.  we can increase logger testing elsewhere.

PR #1188, opened 2016-07-20
(maint) remove TODO from lets_write_a_test doc

er0ck:
[skip ci]

### DIFF
--- a/docs/Lets-Write-a-Test.md
+++ b/docs/Lets-Write-a-Test.md
@@ -1,8 +1,8 @@
-###The Task
+##The Task
 
 Consider if mcollectived incorrectly spawned a new process with every puppet agent run on Ubuntu 10.04.  We need an acceptance test to check that a new process is not spawned and to ensure that this issue does not regress in new builds.
 
-###Figure Out Test Steps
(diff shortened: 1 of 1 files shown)


docs/Lets-Write-a-Test.md
```diff
@@ -38,36 +42,38 @@ Here's our magic command that restarts mcollective:
...
-  # Restart once
-  on(host, restart_command) { assert_equal(0, exit_code) }
-  # Restart again
-  on(host, restart_command) { assert_equal(0, exit_code) }
-  # Check to make sure only one process is running
-  on(host, process_count_check) { assert_equal(0, exit_code) }
-end
-```
+
+    test_name "/etc/init.d/pe-mcollective restart check"
```
kevpl: there's one advantage to using the three backticks method for this code block, and it's that you get ruby-specific highlighting, which I find really nice.  You can get that here by specifying

    ```ruby

as the opening mark on the code block. I thought about mentioning it above, but that line won't get as much from doing something like this since it's really ruby quoting a bash command, but here it could be nice for readers.

kevpl:
looks good 👍

PR #1190, opened 2016-07-20
(BKR-556) Print line number from test file in stack in summary on fail

samwoods1:
(BKR-556) Print line number from test file in stack in summary on fail


samwoods1:
@puppetlabs-jenkins  retest this please

PR #1192, opened 2016-07-20
(MAINT) add generic dmg and msi installers

johnduarte:
This commit adds a generic dmg installer to the OS X package facility.
This allows either a either a local file location or a URL to be specified
to install a dmg on an OS X host. It is based on the existing `install_dmg`
method, but requires the Volume location and package name to be specified by
the caller. These values cannot be introspected without making assumptions
about the packaging.

It also adds a generic msi installer as a helper. It is based on the
`install_msi_on` helper. The additional helper requirements make it a larger
effort to port to the Windows package facility which is not pursued in this
commit. The interface of the `generic_install_msi_on` method is the same as
the `install_msi_on`, it simply removes the validations that are puppet
specific.

### DIFF
--- a/acceptance/tests/install/from_file.rb
+++ b/acceptance/tests/install/from_file.rb
@@ -1,0 +1,18 @@
+test_name 'test generic installers'
+
+step 'install arbitrary msi via url' do
+  hosts.each do |host|
+    if host['platform'] =~ /win/
+      # this should be implemented at the host/win/pkg.rb level someday
+      generic_install_msi_on(host, 'https://releases.hashicorp.com/vagrant/1.8.4/vagrant_1.8.4.msi', {}, {:debug => true})
+    end
+  end
+end
+
+step 'install arbitrary dmg via url' do
+  hosts.each do |host|
+    if host['platform'] =~ /osx/
+      host.generic_install_dmg('https://releases.hashicorp.com/vagrant/1.8.4/vagrant_1.8.4.dmg', 'Vagrant', 'Vagrant.pkg')
+    end
+  end
+end
--- a/lib/beaker/host/mac/pkg.rb
+++ b/lib/beaker/host/mac/pkg.rb
@@ -6,8 +6,26 @@ module Mac::Pkg
   end
 
   def install_package(name, cmdline_args = '', version = nil)
-    execute("hdiutil attach #{name}.dmg")
-    execute("installer -pkg /Volumes/#{name}/#{name}.pkg -target /")
+    generic_install_dmg("#{name}.dmg", name, "#{name}.pkg")
+  end
+
+  # Install a package from a specified dmg
+  #
+  # @param [String] dmg_file      The dmg file, including path if not
+  #                               relative. Can be a URL.
+  # @param [String] pkg_base      The base name of the directory that the dmg
+  #                               attaches to under `/Volumes`
+  # @param [String] pkg_name      The name of the package file that should be
+  #                               used by the installer
+  # @example: Install vagrant from URL
+  #   mymachost.generic_install_dmg('https://releases.hashicorp.com/vagrant/1.8.4/vagrant_1.8.4.dmg', 'Vagrant', 'Vagrant.pkg')
+  def generic_install_dmg(dmg_file, pkg_base, pkg_name)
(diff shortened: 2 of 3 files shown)


lib/beaker/host/mac/pkg.rb
```diff
@@ -10,6 +10,25 @@ def install_package(name, cmdline_args = '', version = nil)
...
+  #
+  # @param [String] dmg_file      The dmg file, including path if not
+  #                               relative. Can be a URL.
+  # @param [String] pkg_base      The base name of the directory that the dmg
+  #                               attaches to under `/Volumes`
+  # @param [String] pkg_name      The name of the package file that should be
+  #                               used by the installer
+  # @example: Install vagrant from URL
+  #   mymachost.generic_install_dmg('https://releases.hashicorp.com/vagrant/1.8.4/vagrant_1.8.4.dmg', 'Vagrant', 'Vagrant.pkg')
+  def generic_install_dmg(dmg_file, pkg_base, pkg_name)
```
kevpl: can we refactor the [install_package](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host/mac/pkg.rb#L8) method to use this one, so that we don't have the duplicate `hdiutil` and `installer` commands?

acceptance/tests/generic/generic_test.rb
```diff
@@ -0,0 +1,18 @@
+test_name 'test generic installers'
```
kevpl: for the file path for this test, can we put it in a folder that specifies that we're testing installers? Maybe `acceptance/tests/install/generic_test.rb`?

lib/beaker/host/mac/pkg.rb
johnduarte: Hmm. l don't see `install_package` and `install_dmg` being equivalent.

The interface for `install_package` takes the name of a package to install, which I see as equivalent to using a package management system (e.g `apt-get install foo`, `yum install foo`). Whereas, `install_dmg` takes a file identifier, which I see as equivalent to using a lower level package manager (e.g `dpkg -i foo.deb`, `rpm -i foo.rpm`). Combining this functionality would require a considerable amount of work across the `install_package` implementations for all of host operating systems.

I will defer to your judgment, @kevpl , but this seems like a large can of worms to open.

acceptance/tests/generic/generic_test.rb
johnduarte: :+1:

lib/beaker/host/mac/pkg.rb
kevpl: I see what you mean from a conceptual level, but I'm just talking about duplicate code in the mac implementation of both methods. If we replaced the current `install_package` (just above, L8) with this, would it not function exactly the same as before, except that you'd be able to use `install_package` with URLs as well?
```ruby
# current version:
def install_package(name, cmdline_args = '', version = nil)
  execute("hdiutil attach #{name}.dmg")
  execute("installer -pkg /Volumes/#{name}/#{name}.pkg -target /")
end

# refactored:
def install_package(name, cmdline_args = '', version = nil)
  generic_install_dmg("#{name}.dmg", name, "#{name}.pkg")
end
```
I'll defer to your knowledge here, especially from your recent testing for this change. I just wanted to bring that up as a way to potentially remove duplicate code. Let me know what you decide looking at this, and hopefully we'll get this merged & released soon.

johnduarte: Ah, that makes perfect sense, @kevpl . My apologies, I misunderstood the nature of your refactoring request. I will amend my commit to remove the duplicate code as per your suggestion.

johnduarte:
The CI failure looks unrelated to this change to me. Unable to checkout designated SHA.

kevpl:

acceptance/tests/install/from_file.rb
```diff
@@ -0,0 +1,18 @@
+test_name 'test generic installers'
+
```
tvpartytonight: @johnduarte I was curious about the case when the operating system was neither `osx` or `win`; in this case, the test will indeed pass, but nothing will have actually really been tested. Should this have a `skip_test` condition at the top?

johnduarte: @tvpartytonight good catch. I will submit a maint patch to add the `skip_test`

PR #1193, opened 2016-07-21
(QENG-3198) Mechanism for archiving SUT files

nwolfe:
This commit adds a new host helpers method called `archive_file_from`
that is intended to be used in post-suites for grabbing arbitrary files
from the SUTs and puts them in a known location where the Jenkins
Archive Artifact plugin can find them and present them with the build
results.

The implementation is basically thin wrapper around `scp_from` that sets
up a particular directory structure with a hardcoded root directory.
This hardcoded root directory will need to be added to the Jenkins
acceptance job configurations so they can be published along with the
rest of the build results.


nwolfe:
@waynr @kevpl for review

lib/beaker/dsl/helpers/host_helpers.rb
```diff
@@ -224,6 +224,45 @@ def rsync_to host, from_path, to_path, opts = {}
...
+        # @param [Host] host A host object (or some object that can be passed to
+        #                    #scp_from)
+        # @param [String] from_path A remote absolute path on the host to copy.
+        # @!macro common_opts
+        # @options [String] archive_root The local directory to store the copied
+        #                                file under. Defaults to
+        #                                'archive/sut-files'.
+        #
+        # @return [Result] Returns the result of the #scp_from operation.
+        def archive_file_from(host, from_path, opts = {}, archive_root = 'archive/sut-files')
```
nwolfe: We don't _need_ to make the archive root configurable, but I thought it might be useful when running tests locally if you wanted to route them to a different location.

But, this means we have two arguments with default values: `opts` and `archive_root`.
Assuming users aren't on newer Ruby versions that support keyword arguments...I decided to put go against the grain and make `archive_root` the last argument rather than `opts`.
This is to prevent users having to repeat the sensitive string `archive/sut-files` in the case when they want to provide some `opts` (which just go to the `scp_from` method).  This default value is important because that's what all our CI job-templates will be hardcoded to look for.

But maybe we don't need to make the archive root configurable?  Is it worth it?
But maybe we don't need to have an `opts` parameter either? I added it because I could imagine users wanting to be able to control the underlying `scp_from` for things like logging.

nwolfe: Also, suggestions welcome for method name.

And should this method support an array of hosts rather than just a single host?
`scp_from` takes one or more.

nwolfe:
There will be an accompanying PR to ci-job-configs where I add the `archive/sut-files/**` path to the defaults/job-templates.

@waynr @kevpl Do you guys think a dedicated Confluence page for how to use this feature is still worthwhile?  Or maybe we just write some docs in this repo?

nwolfe:
Other half: https://github.com/puppetlabs/ci-job-configs/pull/1520

lib/beaker/dsl/helpers/host_helpers.rb
```diff
@@ -224,6 +224,45 @@ def rsync_to host, from_path, to_path, opts = {}
...
+        # @param [Host] host A host object (or some object that can be passed to
+        #                    #scp_from)
+        # @param [String] from_path A remote absolute path on the host to copy.
+        # @!macro common_opts
+        # @option [String] archive_root The local directory to store the copied
+        #                               file under. Defaults to
+        #                               'archive/sut-files'.
+        #
+        # @return [Result] Returns the result of the #scp_from operation.
+        def archive_file_from(host, from_path, opts = {}, archive_root = 'archive/sut-files')
```
nwolfe: (see my original comments in the Outdated section on GitHub)

nwolfe: Another thought....what about making `archive_root` a key in the `opts`?

Or do we typically keep method-specific options out of the `opts` parameter (this is assuming that `opts` is the global Beaker options)?

waynr: In my opinion for something like this it's better to have users "configure" the behavior where it is used rather than someplace far away (is that what "action at a distance means"?...when some configuration option set somewhere else configures the behavior for a localized piece of code?). But I'm definitely not a Beaker expert so I'll defer to pretty much anyone else's opinion.

waynr:
@nwolfe does this work for file globs? Would it be possible to have this new method take an array of filenames to batch the SCP requests? Maybe it wouldn't be that useful.

nwolfe:
> Does this work for file globs?

No.  It doesn't expand the glob, it thinks `*` is the actual name of the file.

> Would it be possible to have this new method take an array of filenames to batch the SCP requests?

No, since neither `scp_from` nor `Host#do_scp_from` support multiple files.  That doesn't mean it's impossible to do, but it's not just done for us already.

@wayne But, I do think the code is written in a way that would make it possible for us to add these behaviors later in a backwards-compatible manner.

lib/beaker/dsl/helpers/host_helpers.rb
waynr: To be more clear about what I mean here, users will be writing the code that calls this method...since that is the case, they should be responsible for any options values that modify its behavior (if that's what they want). Since all we are adding is the method, let's let them decide at the point where they call the method whether or not to override the (quite reasonable) default.

waynr:
:+1:

nwolfe:
jenkins retest this please

lib/beaker/dsl/helpers/host_helpers.rb
kevpl: much rather have local options just be local options. I think the pattern we have of passing the global options hash into method-specific options hashes is a bad one.

PR #1197, opened 2016-07-25
(BKR-80) created style guide for beaker tests

kevpl:
[skip ci]

### DIFF
--- a/docs/concepts/style_guide.md
+++ b/docs/concepts/style_guide.md
@@ -1,0 +1,216 @@
+# Beaker Style Guide
+
+## Scope of this guide
+
+The purpose of this guide is to provide definitions for best practices when
+writing Beaker tests, both syntactically and stylistically. This guide will
+define and provide examples for preferred test layout and conventions. Common
+patterns that are recommended as well as patterns that should be avoided will be
+described.
+
+No style manual can cover every possible circumstance. When a judgement call
+becomes necessary, keep in mind the following general ideas:
+
+1. **Readability matters**. If you have to choose between two equally effective
+alternatives, pick the more readable one. This is, of course, subjective, but if
+you can read your own code three months from now, that's a great start. Don't be
+clever over readable, unless you have a documented purpose. Use object oriented
+programming when things get complex and
+[DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).
+2. **Inherit upstream conventions**. Beaker is still ruby, so use the
+[ruby community style guide](https://github.com/bbatsov/ruby-style-guide). When
+not called out here, use the ruby style guide.
+
+## Test Naming
+
+Tests should test what they say they test. Test names, both the name of the test
+file and the value given to the
+[`test_name`](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Structure#test_name-instance_method)
+function, should provide an accurate indication about the purpose of the test.
+
+The `test_name` function should be the first line in the Beaker test file.
+
+**Good:**
+```ruby
+# head -1 puppet/acceptance/tests/resource/file/should_default_mode.rb
+test_name "file resource: set default modes" do
+```
+
+## Structure Methods Should Use Explicit Blocks
+
+These methods aid in self-documenting your tests, including indention in the
+logs. If you don't use explicit blocks, beaker does not know how to properly
+indent your test's output.
+
+The most common
+[structure methods](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Structure)
+are
+[`#test_name`](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Structure#test_name-instance_method),
+[`#step`](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Structure#step-instance_method),
+and
+[`#teardown`](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Structure#teardown-instance_method).
+
+**Good:**
+```ruby
+step 'do this thang' do
+  on(host, "echo 'do this thang'")
+end
+```
+**Bad:**
+```ruby
+step 'do this thang'
+on(host, "echo 'do this thang'")
+```
+
+## Teardowns
+
+- Return the state of the system to the way it was prior to test execution
+- Put the teardown as early in the test as possible
+
+Teardowns must be used to return the system to the state it was in prior to the
+execution or attempted execution of the test. Beaker will gather all teardowns
+encountered throughout the execution path of the test. These teardowns will all
+be executed when the test exits, even if the test exits early due to a failure
+or error.
+
+### Place Teardowns Early
+
+Teardowns can be placed anywhere in the test file or its helpers. The preferred
+style is to have a teardown step near the beginning of the test file to show the
+reader that the system state will be restored.
+
+**Good:**
+```ruby
+test_name 'The source attribute' do
+
+  target_file_on_nix = '/tmp/source_attr_test'
+  teardown do
+    hosts.each do |host|
+      on(host, "rm #{target_file_on_nix}", :accept_all_exit_codes => true) unless host['platform'] =~ /^win/
+    end
+  end
+  ...
+end
+```
+
+Teardowns are at the mercy of the scoping of the variables necessary to perform
+the restoration of the system. This fact means that additional teardown steps
+will need to be added within the scope necessary to do their job. Effort should
+be taken to make the teardown steps prominent and readable so that it can be
+confirmed, via the logs, that the system has been restored.
+
+Teardown steps registered outside of tests should use
+[`#step`](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Structure#step-instance_method)
+to document and log what they are doing.
+
+## Acceptable Exit Codes
+
+When using the Beaker
+[`on`](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Helpers/HostHelpers#on-instance_method)
+method, the default setting is that only an exit code of 0 (zero) will not
+trigger an error. When other exit codes are acceptable, the
+`:acceptable_exit_codes` key with an array of exit codes should be passed in the
+options hash to `#on`. If 0 (zero) is the only acceptable exit code, then the
+`:acceptable_exit_codes` symbol must not be used.
+
+**Good:**
+
+- Single 0 exit code allowed
+```ruby
+on(host, "rm #{file_to_rm}")
+```
+- Single non-0 exit code allowed
+```ruby
+on(host, "rm #{file_to_rm}", :acceptable_exit_codes => 1)
+```
+- Multiple exit codes allowed
+```ruby
+on(host, "rm #{file_to_rm}", :acceptable_exit_codes => [0,1])
+```
+- Any exit code allowed
+```ruby
+on(host, "rm #{file_to_rm}", :accept_all_exit_codes => true)
+```
+
+In the last case, when any exit code is allowed, one must follow-up with a valid
+assertion test.
+
+If an exit_code outside of 0 is expected, one must use acceptable_exit_codes so
+the test will fail on the proper assertion and not error at that command.  Allow
+only the minimum expected set of exit codes unless coverage is provided by
+subsequent assertions.
+
+## Test Outcomes
+
+When to use each, and how to format the message:
+
+### Expecting Failure
+
+If your tests are failing due to an "expected failure", you should wrap your
+failing assertion in an
+[`expect_failure`](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Structure#expect_failure-instance_method)
+block with an explanatory logging message:
+
+```ruby
+expect_failure('expected to fail due to PE-1234') do
+  assert_equal(400, response.code, 'bad response code from API call')
+end
+```
+
+Note that `expect_failure` will only trigger from failed assertions. It won't
+take care of failed host or `on` commands. To deal with expected failure from an
+`on` invocation, you'd want something more like this:
+
+```ruby
+on(blah, 'blah', :allow_all_exit_codes => true) do |result|
+  expect_failure 'known issue: TIK-1234' do
+    assert_equal(4,result.exit_code,'did not receive expected exit code for blah')
+  end
+end
+```
+
+### `fail_test`, `pass_test`, & `skip_test`
+
+These can be used anywhere in a test to exit early. A
+[`skip_test`](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Outcomes#fail_test-instance_method)
+between two assertions, for instance, will run the first assertion, raise an
+exception for `skip_test`, run teardown, and then exit as we expect. The same is
+true for
+[fail_test](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Outcomes#fail_test-instance_method).
+
+[`pass_test`](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Outcomes#fail_test-instance_method)
+is typically not required. When the end of a test is reached without causing an
+error (due to bad test code execution, or an unhandled exit code or exception)
+or failure (due to assertions), then it passes.
+
+`pass_test` can be used in a situation where one knows a test has passed before
+the end of a test under certain circumstances, such as during a loop that has
+not yet completed.
+
+### Skipping Tests
+
+Skipping tests can be used, for instance, when they are temporarily failing or
+not yet complete.
+
+**Good:**
+```ruby
+skip_test 'requires puppet and mcollective service scripts from AIO agent package' if @options[:type] != 'aio'
+```
+**Bad:**
+```ruby
+confine :to, :platform => 'solaris:pending'
+```
+
+## Confining
+
+Another way that you can skip or manipulate tests is by confining them to apply
+to a subset of the SUTs available for testing. Confining is a complex topic, and
+one we don't have the length to get into in the style guide. For an explanation
+of confining, as well as the best practices in using it, check out our
+[confine doc](../how_to/confine.md).
+
+## Assertions
+
+Always include a _unique_ error message in your assertion statement. Use strict
+asserts whenever possible (e.g. assert_equal, assert_match.
+[More info](http://danwin.com/2013/03/ruby-minitest-cheat-sheet/)).


kevpl:
@puppetlabs/beaker, @er0ck, @cowofevil, @johnduarte, this is ready for a look!

docs/concepts/style_guide.md
```diff
@@ -0,0 +1,178 @@
+# Beaker Style Guide
+
+## Scope of this guide
+
+The purpose of this guide is to provide definitions for preferred practices when
+writing Beaker tests, both syntactically and stylistically. This guide will
+define and provide examples for preferred test layout and conventions. Common
+patterns that are recommend as well as patterns that should be avoided will be
```
er0ck: prolly my fault here: "recommend" -> "recommended"

docs/concepts/style_guide.md
```diff
@@ -0,0 +1,178 @@
...
+described.
+
+No style manual can cover every possible circumstance. When a judgement call
+becomes necessary, keep in mind the following general ideas:
+
+1. **Readability matters**. If you have to choose between two equally effective
+alternatives, pick the more readable one. This is, of course, subjective, but if
+you can read your own code three months from now, that’s a great start. Don’t be
```
johnduarte: nit pick, replace smart quote in `that's` and `don't` with simple ascii single quote

docs/concepts/style_guide.md
```diff
@@ -0,0 +1,178 @@
...
+**Good:**
+```ruby
+# head -1 puppet/acceptance/tests/resource/file/should_default_mode.rb
+test_name "file resource: set default modes"
+```
+
+## Structure Methods Should Use Explicit Blocks
+
+These methods aid in self-documenting your tests, including indention in the
+logs. If you don’t use explicit blocks, beaker does not know how to properly
```
johnduarte: nit pick, replace smart quote in `don't` with simple ascii single quote

docs/concepts/style_guide.md
```diff
@@ -0,0 +1,178 @@
...
+indent your test’s output.
```
johnduarte: nit pick, replace smart quote in `test's` with simple ascii single quote

docs/concepts/style_guide.md
```diff
@@ -0,0 +1,178 @@
...
+If an exit_code outside of 0 is expected, one must use acceptable_exit_codes so
+the test will fail on the proper assertion and not error at that command.  Allow
+only the minimum expected set of exit codes unless coverage is provided by
+subsequent assertions.
+
+## Outcomes
+
+When to use each, how to format the message:
+
+### `fail_test`, `pass_test`, & `skip_test`
```
johnduarte: trailing whitespace on this line

docs/concepts/style_guide.md
```diff
@@ -0,0 +1,178 @@
...
+only the minimum expected set of exit codes unless coverage is provided by
+subsequent assertions.
+
+## Outcomes
+
+When to use each, how to format the message:
+
+### `fail_test`, `pass_test`, & `skip_test` 
+
+These can be used anywhere in a test to exit early. A
```
johnduarte: trailing whitespace on this line

docs/concepts/style_guide.md
```diff
@@ -0,0 +1,178 @@
...
+**Good:**
+```ruby
+skip_test 'should never run on solaris' if agent['platform'] =~ /solaris/
+```
+**Bad:**
+```ruby
+confine :to, :platform => 'solaris:pending'
+```
+
+If they are failing due to an “expected failure” you should wrap your failing
```
johnduarte: nit pick, replace smart quotes wrapping `"expected failure"` with simple ascii double quotes

docs/concepts/style_guide.md
```diff
@@ -0,0 +1,178 @@
...
+file and the value given to the
+[`test_name`](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Structure#test_name-instance_method)
+function should provide an accurate indication about the purpose of the test.
+
+The `test_name` function should be the first line in the Beaker test file.
+
+**Good:**
+```ruby
+# head -1 puppet/acceptance/tests/resource/file/should_default_mode.rb
+test_name "file resource: set default modes"
```
johnduarte: Since we are advocating for structure methods to use blocks, the good example should include a `do` block opener.

docs/concepts/style_guide.md
```diff
@@ -0,0 +1,178 @@
...
+step 'do this thang' do
+  on(host, "echo 'do this thang'")
+end
+```
+**Bad:**
+```ruby
+step 'do this thang'
+on(host, "echo 'do this thang'")
+```
+
```
johnduarte: In the future, we should include opinionated examples of dealing with variable scoping in the context of using blocked structure methods.

docs/concepts/style_guide.md
```diff
@@ -0,0 +1,178 @@
...
+When using the Beaker
+[`on`](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Helpers/HostHelpers#on-instance_method)
+method, the default setting is that only an exit code of 0 (zero) will not
+trigger an error. When other exit codes are acceptable, the
+`:acceptable_exit_codes` key with an array of exit codes should be passed in the
+options hash to `#on`. If 0 (zero) is the only acceptable exit code, then the
+`:acceptable_exit_codes` symbol must not be used.
+
+Each example can be considered a bad example for the situation below it:
```
johnduarte: This line is not clear to me.

I would also prefer to not see the use cases lumped in a single source listing.

Alternative suggestion:

`
**Good:**
* Single 0 exit code allowed
```ruby
on(host, "rm #{file_to_rm}")
```
* Single non-0 exit code allowed
```ruby
on(host, "rm #{file_to_rm}", :acceptable_exit_codes => 1)
```
* Multiple exit codes allowed
```ruby
on(host, "rm #{file_to_rm}", :acceptable_exit_codes => [0,1])
```
* Any exit code allowed
```ruby
on(host, "rm #{file_to_rm}", :accept_all_exit_codes => true)
```
`

docs/concepts/style_guide.md
```diff
@@ -0,0 +1,178 @@
...
+should not be used unless fast-passing in a conditional branch.
+
+### Skipping Tests
+
+Skipping tests can be used, for instance, when they are temporarily failing or
+not yet complete.
+
+**Good:**
+```ruby
+skip_test 'should never run on solaris' if agent['platform'] =~ /solaris/
```
johnduarte: Hrm, given the definition above, I would not consider this a good example. These examples seem fit for explaining the use of the `confine` facility rather than for `skip_test`.

I would suggest that a better use of `skip_test` if for a known issue or condition that is broader than can be expressed using `expect_failure`.

Alternate suggestion:
```
skip_test 'requires puppet and mcollective service scripts from AIO agent package' if @options[:type] != 'aio'
```

I think most of our usage for `skip_test` is associated with specific platforms. However, I would consider that an anti-pattern that should be replaced with one or more calls to the `confine` facility.

er0ck: agreed.  i struggled with this last week.  thanks john!

docs/concepts/style_guide.md
```diff
@@ -0,0 +1,178 @@
+# Beaker Style Guide
+
+## Scope of this guide
+
+The purpose of this guide is to provide definitions for preferred practices when
```
cowofevil: preferred -> best

We should be confident in our recommendations.

docs/concepts/style_guide.md
```diff
@@ -0,0 +1,178 @@
...
+described.
+
+No style manual can cover every possible circumstance. When a judgement call
+becomes necessary, keep in mind the following general ideas:
+
+1. **Readability matters**. If you have to choose between two equally effective
+alternatives, pick the more readable one. This is, of course, subjective, but if
+you can read your own code three months from now, that’s a great start. Don’t be
+clever over readable, unless you have a documented purpose. Use object oriented
+programming when things get complex and DRY.
```
cowofevil: Link to DRY definition.

docs/concepts/style_guide.md
cowofevil: +1

docs/concepts/style_guide.md
cowofevil: Agreed. We should emphasize ```expect_failure``` over skip_test.

docs/concepts/style_guide.md
```diff
@@ -0,0 +1,178 @@
...
+If they are failing due to an “expected failure” you should wrap your failing
+assertion in an
+[`expect_failure`](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Structure#expect_failure-instance_method)
+block with an explanatory logging message:
+
+```ruby
+expect_failure('expected to fail due to PE-1234') do
+  assert_equal(400, response.code, 'bad response code from API call')
+end
+```
```
cowofevil: You should make it really clear that ```expect_failure``` will only trigger off of failed assertions and not failed ```on``` outcomes. I would suggest that you give a short demonstration on how to deal with an ```on``` statement that has an expected failure and how to incorporate that with ```expect_failure```.

cowofevil:
I think you need to add a section about confining. It is one of the areas that causes the most confusion on top of being incredibly useful.

johnduarte:
Suggestion
`
## Confining hosts
The `confine` method  is used to narrow the `hosts` array based on criteria for either filtering out or only including hosts.

The primary use for `confine` is to `exclude` host platforms that are not applicable for the test in question. This usage should be used rather than constructing your own logic to call `skip_test`. This will prevent redundant code from creeping into the test suite. The `confine` method can also be called multiple times, so there is no need to combine disparate criteria together.

**Good:**
* Omit operating systems that are not applicable for a test
```
confine :except, :platform => 'windows'
confine :except, :platform => 'osx'
```
* Include only hosts that meet a desired criteria
```
confine :to, {}, agents.select { |agent| supports_systemd?(agent) }
````
`

File: docs/concepts/style_guide.md (hunk 14 above)
Comment (er0ck) [inline docs/concepts/style_guide.md]: +1
something like
```
on(blah, 'blah', :allow_all_exit_codes => true) do |result|
  expect_failure 'known issue: TIK-1234' do
    assert_equal(4,result.exit_code,'did not receive expected exit code for blah')
  end
end
```

Comment (er0ck) [inline docs/concepts/style_guide.md]: the above also illustrates why we might need a quick TL;DR on when to (not) use parens in method calls, even though it's spelled out in the ruby community style guide.

**Comment by er0ck** [comment]
IMO, above might be long for a style guide.  maybe put some of the verbiage from there into the confine docs, and reference back to them, retaining the good.

**Comment by pinkypie** [comment]
I wrote this a while ago after lots of tickets due to the confusion (and optional hosts array): https://github.com/puppetlabs/beaker/blob/master/docs/how_to/confine.md

**Comment by johnduarte** [comment]
+1 @pinkypie

File: docs/concepts/style_guide.md (hunk 8 above)
Comment (ericwilliamson) [inline docs/concepts/style_guide.md]: +1, I am very much an avid supporter of wrapping both `test_name` and `step`'s in `do...end` blocks. It improves the readability a lot.

File: docs/concepts/style_guide.md (hunk 14 above)
Comment (kevpl) [inline docs/concepts/style_guide.md]: added below this

**Comment by kevpl** [comment]
added a confine section to the bottom, linking to @pinkypie's doc on it.

@puppetlabs/beaker, @johnduarte, @er0ck, @cowofevil, ready for another look!

File: docs/concepts/style_guide.md (hunk 11 above)
Comment (kevpl) [inline docs/concepts/style_guide.md]: @johnduarte, @er0ck, @cowofevil: I've included the better example here that was suggested, but I didn't re-word any of the text around it. I suspect that I'm missing part of the message sent in the feedback since that's the case. Would you mind elaborating further, or suggest a text workaround that would be better as well?

Feel free to come over and beat me on the head with sticks until understanding strikes as well, I hear that works.

File: docs/concepts/style_guide.md [hunk 15]
```diff
@@ -0,0 +1,206 @@
...
+If an exit_code outside of 0 is expected, one must use acceptable_exit_codes so
+the test will fail on the proper assertion and not error at that command.  Allow
+only the minimum expected set of exit codes unless coverage is provided by
+subsequent assertions.
+
+## Outcomes
+
+When to use each, how to format the message:
+
+### `fail_test`, `pass_test`, & `skip_test`
```
cowofevil: Rename to `fail_test`, `pass_test`, `expect_failure` & `skip_test`

docs/concepts/style_guide.md
```diff
@@ -0,0 +1,206 @@
...
+exception for `skip_test`, run teardown, and then exit as we expect. The same is
+true for
+[fail_test](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Outcomes#fail_test-instance_method).
+
+[pass_test](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Outcomes#fail_test-instance_method)
+should not be used unless fast-passing in a conditional branch.
+
+### Skipping Tests
+
+Skipping tests can be used, for instance, when they are temporarily failing or
```
cowofevil: I would change the order to have ```expect_failure``` explained first for skipping followed by ```skip_test```. I think this is important to demonstrate the preference for ```expect_failure``` over ```skip_test```.

docs/concepts/style_guide.md
```diff
@@ -0,0 +1,206 @@
...
+
+These can be used anywhere in a test to exit early. A
+[`skip_test`](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Outcomes#fail_test-instance_method)
+between two assertions, for instance, will run the first assertion, raise an
+exception for `skip_test`, run teardown, and then exit as we expect. The same is
+true for
+[fail_test](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Outcomes#fail_test-instance_method).
+
+[pass_test](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Outcomes#fail_test-instance_method)
+should not be used unless fast-passing in a conditional branch.
```
cowofevil: You'll want to expand on what you mean by "fast-passing".

cowofevil:
@kevpl I made two comments otherwise I'm :+1: at this point.

docs/concepts/style_guide.md
```diff
@@ -0,0 +1,206 @@
...
+2. **Inherit upstream conventions**. Beaker is still ruby, so use the
+[ruby community style guide](https://github.com/bbatsov/ruby-style-guide). When
+not called out here, use the ruby style guide.
+
+## Test Naming
+
+Tests should test what they say they test. Test names, both the name of the test
+file and the value given to the
+[`test_name`](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Structure#test_name-instance_method)
+function should provide an accurate indication about the purpose of the test.
```
johnduarte: add comma `function, should` to complete the clause.

docs/concepts/style_guide.md
```diff
@@ -0,0 +1,206 @@
...
+style is to have a teardown step near the beginning of the test file to show the
+reader that the system state will be restored.
+
+**Good:**
+```ruby
+test_name 'The source attribute' do
+
+  target_file_on_nix = '/tmp/source_attr_test'
+  teardown do
+    on(hosts[0], "rm #{target_file_on_nix}", :accept_all_exit_codes => true)
```
johnduarte: Using an indexed array reference for host does not seem like a great practice to me. If this is the example we want to use, it would seem safer to wrap it in a conditional test for platform type.

Suggestion
```
test_name 'The source attribute' do

  target_file_on_nix = '/tmp/source_attr_test'
  teardown do
    hosts.each do |host|
      on(host, "rm #{target_file_on_nix}", :accept_all_exit_codes => true) unless host['platform'] =~ /^win/
    end
  end
  ...
end
```

docs/concepts/style_guide.md
```diff
@@ -0,0 +1,206 @@
...
+```ruby
+on(host, "rm #{file_to_rm}", :acceptable_exit_codes => [0,1])
+```
+- Any exit code allowed
+```ruby
+on(host, "rm #{file_to_rm}", :accept_all_exit_codes => true)
+```
+
+In the last case, when any exit code is allowed, one must follow-up with a valid
+assertion test.
```
johnduarte: +1

docs/concepts/style_guide.md
kevpl: I'm not really sure what "fast-passing in a conditional branch" is, that's something I just took from the doc draft. To be honest, I was surprised we had a `pass_test` method. @er0ck, any clarification on this text?

docs/concepts/style_guide.md
kevpl: @cowofevil I've changed the ordering to put the `expect_failure` section first. I've actually just put it up above the "`fail_test`, `pass_test`, ..." section to show this preference.

I didn't add it to that same section though, because it didn't seem to fit the text in the section as well as the others. Specifically, the methods mentioned are used solely to exit early with a certain status. They don't take blocks, or do anything complex, they just quit, whereas the `expect_failure` method is more complex and depends on the body you pass it. Does this make sense?

cowofevil: @kevpl Yeah, that looks good to me.

docs/concepts/style_guide.md
er0ck: how about:

`pass_test` is typically not required.  When the end of a test is reached without causing an error (due to bad test code execution, or an unhandled exit code or exception) or failure (due to assertions), then it passes.
`pass_test` can be used in a situation where one knows a test has passed before the end of a test under certain circumstances, such as during a loop that has not yet completed.

johnduarte: This sounds good to me @er0ck . :+1:

kevpl:
added in @er0ck's `pass_test` text. 

@johnduarte, @cowofevil, @puppetlabs/beaker, what do you think?

johnduarte:
:+1:  @kevpl . I say ship it! Nice work everyone.

cowofevil:
:shipit:

docs/concepts/style_guide.md
```diff
@@ -0,0 +1,216 @@
...
+confine :to, :platform => 'solaris:pending'
+```
+
+## Assertions
+
+Always include a _unique_ error message in your assertion statement. Use strict
+asserts whenever possible (e.g. assert_equal, assert_match.
+[More info](http://danwin.com/2013/03/ruby-minitest-cheat-sheet/)).
+
+## Confining
```
tvpartytonight: It seems to me that this confine section should just be moved to the `Skipping Tests` section, as there is code that is using `confine` there, but not real explanation given on why it is "bad".

kevpl:
@tvpartytonight I rearranged the sections. I agree that the flow is better if skipping leads right into confining. I don't think they should go together completely though. I think it's misleading to say confining is a way of skipping, since you can confine to constrain your hosts list as well as to skip a test.

tvpartytonight:
👍

PR #1199, opened 2016-07-26
(BKR-860) - Adding support for vagrant to quick start tasks.

pinkypie:


lib/beaker/tasks/quick_start.rb
```diff
@@ -1,49 +1,85 @@
...
+  desc 'Generate Default Beaker Host Config File, valid options are: vmpooler or vagrant.'
+  task :gen_hosts, [:provisioner] do |t, args|
+    hosts_file = "acceptance/config/default__#{args[:provisioner]}_hosts.yaml"
+    if args[:provisioner] == 'vagrant'
+      cli = VAGRANT
+    elsif args[:provisioner] == 'vmpooler'
+      cli = VMPOOLER
+    else
+      puts "No provisioner provided, defaulting to vmpooler."
+      cli = VMPOOLER
```
er0ck: i think we might want to move to defaulting to what users outside of puppet need and use; especially in the docs/guides.  those of us inside puppet have plenty of resources to learn how to use vmpooler with beaker.  

thoughts?

lib/beaker/tasks/quick_start.rb
```diff
@@ -1,49 +1,85 @@
...
+    end
     FileUtils.mkdir_p('acceptance/config') # -p ignores when dir already exists
-    File.open("acceptance/config/default_hosts.yaml", 'w') do |fh|
-      fh.print(cli.execute)
+    if File.exist?(hosts_file) then
+      File.open(hosts_file, 'w') do |fh|
+        fh.print(cli.execute)
+      end
+    else
+      puts "Host Config File: #{hosts_file} already exists so not overwriting"
```
er0ck: this 'warning' message is a bit difficult for me to read.

lib/beaker/tasks/quick_start.rb
```diff
@@ -1,49 +1,85 @@
...
+    system_args[:tests]     = 'acceptance/tests/default_smoke_test.rb'
+    system(beaker_command(system_args))
   end
 
 end
 
-def beaker_command
+
+def beaker_command(system_args)
+  binding.pry
```
er0ck: i know this is a WIP, but this should be removed later.

lib/beaker/tasks/quick_start.rb
```diff
@@ -1,49 +1,85 @@
...
   cmd_parts = []
   cmd_parts << "beaker"
-  cmd_parts << "--hosts acceptance/config/default_hosts.yaml"
-  cmd_parts << "--pre-suite acceptance/setup/default_pre_suite.rb"
-  cmd_parts << "--tests acceptance/tests/default_smoke_test.rb"
+  cmd_parts << "--hosts #{system_args[:hosts]}"
+  cmd_parts << "--pre-suite #{system_args[:pre_suite]}"
+  cmd_parts << "--tests #{system_args[:tests]}"
   cmd_parts.flatten.join(" ")
```
er0ck: should this beaker_command have an ENV setup so one can add arbitrary arguments?  rototiller will largely move us away from this, but i've found it useful in the past.  it would allow users to retain this Rakefile and expand functionality without edits.

er0ck:
since this is part of a quick-start type guide, should this Rakefile perhaps be overly verbose about what it's doing? 
e.g.: {{i'm about to run command `beaker-hostgenerator blah`}}
and/or 
should it have tasks for running with preserved-hosts and args for preserving hosts in some tasks?
and/or
should it spit out some info about where the preserved-hosts, hosts.yaml and log files end up?

docs/tutorials/quick_start_rake_tasks.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Summary
+
+### Gen tasks
+
+These tasks are standalone and can be run independently from each other to generate the desired files.
+
+beaker_quickstart:gen_hosts  (generates default host config)
+beaker_quickstart:gen_pre_suite  (generates default pre-suite)
+beaker_quickstart:gen_smoke_test  (generates default smoke test)
```
kevpl: @pinkypie when I view these rendered, they don't show up in a list. Was this intended?

docs/tutorials/quick_start_rake_tasks.md
```diff
@@ -0,0 +1,205 @@
...
+
+If the file already exists, it will not be overwritten. This will allow you to edit the pre_suite file yourself if required.
+
+pre_suite file:
+
+    install_puppet
+  
+See the other tutorial docs in this directory for more information on pre-suites:
+
+* [Test Suites](test_suites.md)
```
kevpl: If we're not going to have more than one reference in this list, perhaps it would be better to just have the sentence be:

    See the [Test Suites doc](test_suites.md) in this directory for more information on pre-suites.

docs/tutorials/quick_start_rake_tasks.md
```diff
@@ -0,0 +1,205 @@
...
+If the file already exists, it will not be overwritten. This will allow you to edit the test file yourself if required.
+
+smoke test file:
+
+    test_name 'puppet install smoketest'
+    step 'puppet install smoketest: verify \'puppet help\' can be successfully called on
+    all hosts'
+        hosts.each do |host|
+          on host, puppet('help')
+        end
```
kevpl: In the new style guide, we're pushing that all invocations of `step` & `test_case` should provide bodies, rather than just running on one line. In order to be consistent with that, let's surround the tests with `do...end` blocks on those calls.

docs/tutorials/quick_start_rake_tasks.md
```diff
@@ -0,0 +1,205 @@
...
+This task will run the above 3 tasks in sequential order and then execute a Beaker test run using all 3 files.
+
+    beaker --hosts acceptance/config/default_vmpooler_hosts.yaml --pre-suite acceptance/setup/default_pre_suite.rb --tests 
+    acceptance/tests/default_smoke_test.rb
+
+You will end up with provisioned hosts, with puppet installed and a test check executed to make sure that puppet was indeed installed.
+
+For more information on running Beaker tests see:
+
+* [Test run](test_run.md)
```
kevpl: same as above: if we're not going to have a list of multiple references, I think it would be better to include the link in the sentence itself.

docs/tutorials/quick_start_rake_tasks.md
```diff
@@ -0,0 +1,205 @@
...
+    CONFIG:
+      nfs_server: none
+      consoleport: 443
+      box_url: https://vagrantcloud.com/puppetlabs/boxes/ubuntu-14.04-64-nocm
+      box: puppetlabs/ubuntu-14.04-64-nocm
+
+
+For more info on host generation and what these configs represent see:
+
+* [Creating A Test Environment](creating_a_test_environment.md)
```
kevpl: same as below about having a list of references

kevpl:
my intention for this doc is to use it as the immediate quick-start doc, to use just after installation is complete. With that in mind, I've got some general feedback about the doc:

- We should start with the introduction "How to Use Them" (HUT) section first, as people don't really need a listing of the tasks until they've got the system up and running
- In the HUT section, we should explain how to use them for both installation methods. This can be as simple as saying "if you've installed from gem, put this in the top of your project's rake file" vs "if you've cloned beaker, you can start by running `rake --tasks`"
- In the HUT section, we say that installing Beaker is a prerequisite, but we don't talk about hypervisors or hypervisor pre-reqs. We shouldn't go too deep into the rabbit hole, but I think we should give a glossary introduction to hypervisors, something just like "Hypervisors are services that provision SUTs for Beaker", and then a similar one-line introduction to the two hypervisors available to start with: vmpooler & vagrant. For both, you have to set things up (unless you're internal). We can link to the hypervisor docs for them, and just reference the steps to follow in those docs to get setup.

lib/beaker/tasks/quick_start.rb
kevpl: @pinkypie is `puts` the best option for writing out in these files? Does Rake itself provide any functionality that we could hook into to manage this better?

kevpl:
Something that I found a little confusing is that it's not clear to me the first time you mention it how you're supposed to surround the hypervisor name in the CLI from these docs:

    rake beaker_quickstart:gen_hosts[provisioner]

Am I actually supposed to say `rake beaker_quickstart:gen_hosts[vmpooler]`, or `rake beaker_quickstart:gen_hosts['vmpooler']`, or `rake beaker_quickstart:gen_hostsvmpooler`? An example with a value here would clear that up. I can understand not wanting to hand-hold that much, but I think it's worth it in tutorial docs to over-communicate.

lib/beaker/tasks/quick_start.rb
pinkypie: Yep, changed the default to vagrant.

lib/beaker/tasks/quick_start.rb
pinkypie: @kevpl - puts is just for the output message, using File methods to open and print to file. Seems to be the common practice even in Rake tasks from what I could gather.

lib/beaker/tasks/quick_start.rb
pinkypie: @er0ck we wanted to keep this initial quick start as simple and easy as possible for new users to get up and running. I think the main rake tasks in [beaker/docs/how_to/rake_tasks.md](url) are the ones we would to expand on possibly with ENV and extras to make the tasks more versatile.

docs/tutorials/quick_start_rake_tasks.md
pinkypie: That should be fixed now.

pinkypie:
@kevpl I've made changes re your comments. For the second point above, I found I needed to require that path at the top of the rakefile whether I had cloned the repo or was using the gem?

pinkypie:
@er0ck I have added more verbose output. I didn't add the ability to preserve hosts or use preserved hosts as we wanted to keep these quick start tasks as simple as possible. @kevpl @tvpartytonight @ferglor unless you think we should?

lib/beaker/tasks/quick_start.rb
```diff
@@ -1,49 +1,93 @@
 require 'beaker-hostgenerator'
 
+VAGRANT  = ['ubuntu1404-64default.mdcal-ubuntu1404-64af', '--hypervisor=vagrant',
+            '--global-config=box_url=https://vagrantcloud.com/puppetlabs/boxes/ubuntu-14.04-64-nocm,box=puppetlabs/ubuntu-14.04-64-nocm']
```
nwolfe: Does this work? 

This is going to need to follow the custom `{key=value,...}` syntax, so I think we need to wrap it in braces like so:
```
--global-config={box_url=https://vagrantcloud.com/puppetlabs/boxes/ubuntu-14.04-64-nocm,box=puppetlabs/ubuntu-14.04-64-nocm}
```

pinkypie: @nwolfe Yeah both ways seem to work in code, from the command line though if i use the {} it only generates the last key value pair which is why I held off from doing this in code. I've updated it though as it makes sense to do it right.

FYI from the command line:

beaker-hostgenerator ubuntu1404-64default.mdcal-ubuntu1404-64af --hypervisor=vagrant --global-config={box_url=https://vagrantcloud.com/puppetlabs/boxes/ubuntu-14.04-64-nocm,box=puppetlabs/ubuntu-14.04-64-nocm}

produces:
CONFIG:
  nfs_server: none
  consoleport: 443
  box: puppetlabs/ubuntu-14.04-64-nocm

If I run: 

beaker-hostgenerator ubuntu1404-64default.mdcal-ubuntu1404-64af --hypervisor=vagrant --global-config=box_url=https://vagrantcloud.com/puppetlabs/boxes/ubuntu-14.04-64-nocm,box=puppetlabs/ubuntu-14.04-64-nocm

I get as expected:

CONFIG:
  nfs_server: none
  consoleport: 443
  box_url: https://vagrantcloud.com/puppetlabs/boxes/ubuntu-14.04-64-nocm
  box: puppetlabs/ubuntu-14.04-64-nocm

er0ck:
i think we should have preserve/run-against-preserved, but it's totally fine as a later PR

docs/tutorials/quick_start_rake_tasks.md
```diff
@@ -0,0 +1,213 @@
...
+
+* Hypervisors are services that provision SUTs for Beaker. We have made two available in this quick start guide to allow you to get up 
+and running [Vmpooler](../how_to/hypervisors/vsphere.md) and [Vagrant](../how_to/hypervisors/vagrant.md).
+
+
+## How to use them
+
+To use the tasks, you need to put the following line at the top of your project's rake file:
+
+require 'beaker/tasks/quick_start'
```
kevpl: this line isn't highlighted or set apart as code snippets should be

lib/beaker/tasks/quick_start.rb
nwolfe: > beaker-hostgenerator ubuntu1404-64default.mdcal-ubuntu1404-64af --hypervisor=vagrant --global-config={box_url=https://vagrantcloud.com/puppetlabs/boxes/ubuntu-14.04-64-nocm,box=puppetlabs/ubuntu-14.04-64-nocm}

You need to escape the `,`'s when invoking on the CLI, which is why it didn't produce the right thing.

FYI the behavior differs between the CLI and CI due to `{` and `}` being special characters the shell interprets rather than passing to BHG.  You can get around this by escaping either the surrounding braces or each comma within the braces.

> beaker-hostgenerator ubuntu1404-64default.mdcal-ubuntu1404-64af --hypervisor=vagrant --global-config=box_url=https://vagrantcloud.com/puppetlabs/boxes/ubuntu-14.04-64-nocm,box=puppetlabs/ubuntu-14.04-64-nocm

By golly you're right!  This is just a coincidence due to how it's written currently and not an API guarantee.

Yeah we should include the braces though since that's the API.

docs/tutorials/quick_start_rake_tasks.md
```diff
@@ -0,0 +1,213 @@
...
+If you have the zsh shell then you will need to escape the square brackets:
+
+    rake beaker_quickstart:gen_hosts\[vmpooler\]
+
+The gen_hosts task will create a file 'default_provisioner_hosts.yaml' in acceptance/config.
+
+If the file already exists, it will not be overwritten. This will allow you to play around with the config yourself, either by manually 
+editing the file or by using beaker-hostgenerator to generate a new hosts config.
+
+Vmpooler file (Ubuntu 14 master and agent):
```
kevpl: I just noticed they're actually redhat hosts below. 😿

docs/tutorials/quick_start_rake_tasks.md
```diff
@@ -0,0 +1,213 @@
...
+For more information on the Beaker dsl methods available to you in your tests see - [Beaker dsl](../how_to/the_beaker_dsl.md)
+
+
+### Run task
+
+This task will run all the above tasks in sequential order, to generate a hosts file, pre-suite file, smoke test and then use these files
+to perform a Beaker test run. If the files already exist (see below for further info on file names and location) then they will not be
+overwritten.
+ 
+beaker_quickstart:run_test
```
kevpl: shouldn't this be code highlighted like the others? Or actually, since you have it highlighted as it should be run below, it'd probably be better just to erase this and include it in the text above.

One way would be to say at the beginning of that text:

The `beaker_quickstart:run_test` will run all the above...

docs/tutorials/quick_start_rake_tasks.md
```diff
@@ -0,0 +1,213 @@
...
+To run:
+  
+    rake beaker_quickstart:run_test[provisioner]
+    
+This task takes an optional parameter (provisioner), if you leave it empty it will default to 'vagrant'. You can pass in either 'vmpooler' 
+or 'vagrant'.
+    
+If you have the zsh shell then you will need to escape the square brackets:
+    
+    rake beaker_quickstart:run_test\[vmpooler\]
```
kevpl: this is repeated from the section above. Maybe there should just be a "Provisioning Argument" section, so that there's one spot where you explain this? Then down here, you could just say that this task also supports the provisioning argument, rather than needing to explain how that argument works again.

Thanks for putting up with all of this @pinkypie! I know docs can be particularly painful since they go through the most scrutiny, but I think it'll be worth it!

pinkypie: No probs @kevpl, just thinking maybe I should change this param to 'hypervisor' instead of provisioner?

kevpl: I think that makes sense, given that in the "acceptance testing"(/beaker) domain, we usually use hypervisor as the name for these types of services.

pinkypie:
@kevpl I think I've made all the necessary edits.

PR #1202, opened 2016-07-29
(BKR-896) Fix SELinux context on root's SSH keys

trevor-vaughan:
This runs fixfiles on root's home directory after copying in the user
keys so that systems with SELinux enabled by default will allow root to
login afterwards.

### DIFF
--- a/lib/beaker/host_prebuilt_steps.rb
+++ b/lib/beaker/host_prebuilt_steps.rb
@@ -360,6 +360,10 @@ module Beaker
         else
           host.exec(Command.new('sudo su -c "cp -r .ssh /root/."'), {:pty => true})
         end
+
+        if host.exec(Command.new('sudo selinuxenabled'), :accept_all_exit_codes => true).exit_code == 0
+          host.exec(Command.new('sudo fixfiles restore /root'))
+        end
       end
     end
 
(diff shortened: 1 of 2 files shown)


lib/beaker/host_prebuilt_steps.rb
```diff
@@ -360,6 +360,10 @@ def copy_ssh_to_root host, opts
         else
           host.exec(Command.new('sudo su -c "cp -r .ssh /root/."'), {:pty => true})
         end
+
+        if host.exec(Command.new('sudo selinuxenabled'), :accept_all_exit_codes => true).exit_code == 0
```
kevpl: @trevor-vaughan this seems like something that would be great to have as a utility method of the host object. Would you create something along the lines of `selinux_enabled?` in the [unix](https://github.com/puppetlabs/beaker/tree/master/lib/beaker/host/unix) object?

If you just create an override of that method in the [windows](https://github.com/puppetlabs/beaker/tree/master/lib/beaker/host/windows) space (probably the [mac](https://github.com/puppetlabs/beaker/tree/master/lib/beaker/host/mac) space as well, I'm not familiar with selinux myself) that returns false since it's not applicable, that should cover the possibilities. Just to be complete, however, our host hierarchy looks like so:

- Unix
  - AIX
  - Cisco
  - EOS
  - FreeBSD
  - Mac
  - Windows
    - PSWindows

Don't worry if you're not familiar with some of these platforms. Having the main utility method in the Unix space should cover most of the possibilities needed here.

kevpl:
last comment still current

lib/beaker/host_prebuilt_steps.rb
trevor-vaughan: I haven't had time to get to this. If someone wants to just get it in correctly and abandon my PR, that's fine. Otherwise, It might not be too soon.

PR #1209, opened 2016-08-03
(MAINT) Output informational message at debug level.

smortex:
The previous behavior output this as a warning message, causing it to appear in red with the default configuration which is probably not desired since this is expected in normal conditions.

The new behavior use the same debug level to indicate the connection attempt and the connection terminaison.


pinkypie:
👍

ferglor:
👍

PR #1210, opened 2016-08-03
BKR-892 Generalize how beaker determines run_in_parallel option

ferglor:


lib/beaker/shared/options_resolver.rb
```diff
@@ -0,0 +1,21 @@
+module Beaker
+  module Shared
+    # Methods for parsing options.
+    # - determine if a mode should be run in parallel
```
tvpartytonight: This particular comment seems out of place; I think this needs to be moved to the `run_in_parallel?` method, not `OptionsResolver` module.

lib/beaker/shared/options_resolver.rb
```diff
@@ -0,0 +1,21 @@
...
+    module OptionsResolver
+      def run_in_parallel?(opts, options, mode)
```
tvpartytonight: We should probably have some yard docs in repo for this method; checkout [scp example](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host.rb#L358) for a pretty good yarddoc precedent.

tvpartytonight: @puppetlabs/beaker did we ever come up with an agreement of what we really mean in the code base when we name variables `opts` or `options`? If I remember correctly, `opts` are local options, and `options` refer to the global options map we pass around. Or, is that even too limited of a definition?

kevpl: yeah, `opts` is definitely used as our local options. I think a better standard would be to actually call it `local_opt(s/ions)`, however.

lib/beaker/shared/host_manager.rb
```diff
@@ -102,7 +102,7 @@ def run_block_on hosts = [], filter = nil, opts = {}, &block
         end
         if block_hosts.is_a? Array
           if block_hosts.length > 0
-            if opts[:run_in_parallel] == true
+            if run_in_parallel? opts, nil, nil
```
kevpl: Is there a reason we'd rather have users pass `nil` arguments instead of having `nil` default arguments?

ferglor:

tvpartytonight:
👍

PR #1213, opened 2016-08-05
(BKR-904) Provide A Way To Wait For A Host Reboot

spjmurray:
Fixes in issue in Host.port_open? where an exception Errno::EHOSTUNREACH is
raised but not rescued.  Creates a wait_for_port wrapper around this function
to wait upon a specific port for a specific time.  Also allows the interval
between checks to be specified to prevent spamming the host.

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -92,9 +92,24 @@ module Beaker
           TCPSocket.new(reachable_name, port).close
           return true
         end
-      rescue Errno::ECONNREFUSED, Timeout::Error, Errno::ETIMEDOUT
+      rescue Errno::ECONNREFUSED, Timeout::Error, Errno::ETIMEDOUT, Errno::EHOSTUNREACH
         return false
       end
+    end
+
+    # Wait for a port on the host.  Useful for those occasions when you've called
+    # host.reboot and want to avoid spam from subsequent SSH connections retrying
+    # to connect from say retry_on()
+    def wait_for_port(port, attempts=15)
+      @logger.debug("  Waiting for port #{port} ... ", false)
+      start = Time.now
+      done = repeat_fibonacci_style_for(attempts) { port_open?(port) }
+      if done
+        @logger.debug('connected in %0.2f seconds' % (Time.now - start))
+      else
+        @logger.debug('timeout')
+      end
+      done
     end
 
     def up?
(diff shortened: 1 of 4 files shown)


lib/beaker/host.rb
```diff
@@ -92,11 +92,32 @@ def port_open? port
...
-      rescue Errno::ECONNREFUSED, Timeout::Error, Errno::ETIMEDOUT
+      rescue Errno::ECONNREFUSED, Timeout::Error, Errno::ETIMEDOUT, Errno::EHOSTUNREACH
         return false
       end
     end
 
+    # Wait for a port on the host.  Useful for those occasions when you've called
+    # host.reboot and want to avoid spam from subsequent SSH connections retrying
+    # to connect from say retry_on()
+    def wait_for_port(port, timeout=60, interval=1)
```
tvpartytonight: @spjmurray thanks for the PR! For methods that are waiting, we would like to see the `repeat_fibonacci_style_for` method used to exponentially backoff retry attempts. You can see the definition of the method [here](https://github.com/puppetlabs/beaker/blob/a6c2297b7dc5959a55f1db538a967f15bfe9ab2a/lib/beaker/shared/repetition.rb#L14) and an example of it being used [here](https://github.com/puppetlabs/beaker/blob/a6c2297b7dc5959a55f1db538a967f15bfe9ab2a/lib/beaker/host/windows/exec.rb#L59). Instead of a seconds based timeout, I think that a number of retry attempts should be the parameter configurable here.

Also, since this is a new method, we would to see some spec coverage; please take a look at the contributing doc, specifically [here](https://github.com/puppetlabs/beaker/blob/master/CONTRIBUTING.md#testing) on what we are looking for.

Please respond in this PR with any questions you have. Thanks!

spjmurray:
@tvpartytonight haven't got a bloody clue what I'm doing with RSpec so feel free to advise if I'm doing it wrong :)  As a special bonus I've enabled `wait_for_port` on the OpenStack hypervisor which now looks like:
```
OpenStack host puppet (dpce4hr8oympms8) assigned ip: 185.98.148.96
  Waiting for port 22 ... connected in 24.02 seconds
Give root a copy of current user's keys, on puppet
```
Rather than a load of "cannot connect to SSH" spam.

tvpartytonight:

kevpl:

kevpl:

kevpl:

kevpl:

PR #675, opened 2015-02-09
Provide a means of supplying a custom Vagrantfile

mcallaway:
This branch adds vagrant_custom to the hypervisor list, reading the path to a Vagrantfile from CONFIG: vagrantfile:

```
> cat spec/acceptance/nodesets/default.yml
HOSTS:
  trusty:
    roles:
      - apache
    platform: ubuntu-1404-x86_64
    hypervisor: vagrant_custom
    box: trusty64
CONFIG:
  vagrantfile: ./spec/acceptance/nodesets/Vagrantfile
```

With this ability to set your own Vagrantfile in its entirety, you can make use of all of Vagrant's features and plugins with no added code. Thus, Openstack support comes for free with the vagrant-openstack-plugin, and similarly for vsphere, etc.

https://tickets.puppetlabs.com/browse/PUP-3984

PR #1215, opened 2016-08-08
(BKR-907) Add vagrant_custom hypervisor

petems:
* Gives the ability to specify users own Vagrantfile.
* Useful if there's a bug in beaker, you have a custom Vagrant provider missing from Beaker...
* or you just want to be super explicit about your Vagrantfile

### DIFF
--- a/lib/beaker/hypervisor.rb
+++ b/lib/beaker/hypervisor.rb
@@ -41,6 +41,8 @@ module Beaker
           end
         when /^vagrant$/
           Beaker::Vagrant
+        when /^vagrant_custom$/
+          Beaker::VagrantCustom
         when /^vagrant_libvirt$/
           Beaker::VagrantLibvirt
         when /^vagrant_virtualbox$/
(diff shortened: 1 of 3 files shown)


petems:
Resurrected from https://github.com/puppetlabs/beaker/pull/675/ and rebased

@mcallaway are you still using beaker? Do you give this a seal of approval? 👍

petems:
Might need a hand with the specs for this, would a fixture be the best way of handling this?

ferglor:

kevpl:
@petems yeah, I think either a fixture of an example vagrant file with stubbed out values to verify, or I think better, a HEREDOC string at the top of the spec file itself specifying either the entire doc or a template that you can add onto for each of your tests.

kevpl:
last commit still current

kevpl:
last comment is still current. If no motion is made on this PR in another week, it'll be closed due to inactivity.

petems:
@kevpl Is there anyone with some slack time available who can help me with writing tests for this?

kevpl:
@petems, I think you'll want something like this:

```ruby
vagrant_file_contents = <-EOF
  <vagrant file content>
EOF
File.open(vagrant_file_path, 'w') { |file| file.write(vagrant_file_contents) }

it 'writes the vagrant file to the correct location' do
  subject.provision
  vagrant_copy_file = File.open(vagrant_copy_location, 'r')
  expect(File.exists(vagrant_copy_file)).to be === true
end

it "doesn't break the file content when copying" do
  subject.provision
  vagrant_copy_file = File.open(vagrant_copy_location, 'r')
  expect(vagrant_copy_file.read).to be === vagrant_file_contents
end
```

You'll probably have to mock up a few more things, or potentially just test the `make_vhost` method. Now that I think about it, since that's the only thing changed, it would be better to test just that, which you should be able to do with tests like these. If you look at other hypervisor spec files, you should see examples of how we test them.

If you still want to pair on this, or just have any more questions, let us know, and we'll setup a session to finish this with you.

petems:
@kevpl Awesome thanks! I think that'll be enough to give a stab at the unit tests tonight, if I need help I'll ping in hipchat 👍

petems:
Added tests when pairing with @kevpl! 😄

PR #1219, opened 2016-08-10
(BKR-896) Fix SELinux context on root's SSH keys

ferglor:


lib/beaker/host/mac/exec.rb
```diff
@@ -27,4 +27,11 @@ def ssh_permit_user_environment
     ssh_service_restart()
   end
 
+  # Checks if selinux is enabled
+  #
+  # @return [Boolean] true if selinux is enabled, false otherwise
```
ferglor: @tvpartytonight @kevpl @pinkypie should i change the doc here to say this method will only return false? (and also in the exec.rb for the windows host?) would there be any value in creating a selinux_available? method, to be called before querying if selinux is enabled? or would that be superfluous?

kevpl: @ferglor I would say that changing the doc is a good idea, so that people don't think it's possible from reading the docs when it's not.

I don't think there's value in creating an `selinux_available?` method. I'm assuming that usage of that method would mostly look like:
```ruby
other_method_call() if selinux_available? && selinux_enabled?
```
when you could just use the `selinux_enabled?`. You could counter that people would probably do platform checks then add an `selinux_enabled?` method call, but I think that just using `selinux_enabled?` is better than either of those options.

ferglor:

PR #1224, opened 2016-08-17
(BKR-914) Fix loose checking in add, get and clear env_var

glennsarti:
Previously the regular expression to add, get or clear an environment variable
were too loose, matching lines which were not intended e.g. `get_env_var("abc")`
would match `abc=123`, `xx_abc_xx=123` and `123=abc` however the intent of the
function would be to only match `abc=123` i.e. Get the env var called 'abc'.

This commit modifies the `add_env_var`, `get_env_var` and `clear_env_var` grep
statements to prefix `#{key}` with `^` and a trailing `=` to ensure that the
matches are as intended.  This commit also includes additional acceptance tests
for this behavior.

### DIFF
--- a/lib/beaker/host/unix/exec.rb
+++ b/lib/beaker/host/unix/exec.rb
@@ -106,7 +106,7 @@ module Unix::Exec
     if exec(Beaker::Command.new("grep ^#{key}=.*#{escaped_val} #{env_file}"), :accept_all_exit_codes => true ).exit_code == 0
       return #nothing to do here, key value pair already exists
     #see if the key already exists
-    elsif exec(Beaker::Command.new("grep ^#{key} #{env_file}"), :accept_all_exit_codes => true ).exit_code == 0
+    elsif exec(Beaker::Command.new("grep ^#{key}= #{env_file}"), :accept_all_exit_codes => true ).exit_code == 0
       exec(Beaker::SedCommand.new(self['platform'], "s/^#{key}=/#{key}=#{escaped_val}:/", env_file))
     else
       exec(Beaker::Command.new("echo \"#{key}=#{val}\" >> #{env_file}"))
@@ -142,7 +142,7 @@ module Unix::Exec
   #  host.get_env_var('path')
   def get_env_var key
     key = key.to_s
-    exec(Beaker::Command.new("env | grep #{key}"), :accept_all_exit_codes => true).stdout.chomp
+    exec(Beaker::Command.new("env | grep ^#{key}="), :accept_all_exit_codes => true).stdout.chomp
   end
 
   #Delete the environment variable from the current ssh environment
@@ -153,7 +153,7 @@ module Unix::Exec
     key = key.to_s
     env_file = self[:ssh_env_file]
     #remove entire line
-    exec(Beaker::SedCommand.new(self['platform'], "/#{key}=.*$/d", env_file))
+    exec(Beaker::SedCommand.new(self['platform'], "/^#{key}=.*$/d", env_file))
     #update the profile.d to current state
     #match it to the contents of ssh_env_file
     mirror_env_to_profile_d(env_file)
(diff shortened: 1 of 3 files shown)


lib/beaker/host/unix/exec.rb
```diff
@@ -142,7 +142,7 @@ def delete_env_var key, val
   #  host.get_env_var('path')
   def get_env_var key
     key = key.to_s
-    exec(Beaker::Command.new("env | grep #{key}"), :accept_all_exit_codes => true).stdout.chomp
+    exec(Beaker::Command.new("env | grep ^#{key}="), :accept_all_exit_codes => true).stdout.chomp
```
ferglor: would it be better to use printenv here?

glennsarti: It may well be, but it seemed less complicated to just modify the regexp used.  I don't know what the difference is between env and printenv on the various platforms (Linux and Cygwin)

tvpartytonight:
@glennsarti thanks for this PR; unfortunately, I am a bit hesitant to merge this at the moment because I am afraid that there are tests relying on this inaccurately doc'd method. However, the good news is that we are releasing a new major version of beaker in 2+ weeks, and we plan on including this change into that version. 

In the meantime, could you add some spec tests to this method? Thanks again!

Iristyle:
I would be a bit surprised if any tests relied on the broken behavior.  It seems more likely that tests could be generating false positives by returning inaccurate values based on the presence of other environment variables.

I would lobby for getting this into the next release if possible.  (I also agree some tests would be helpful)

lib/beaker/host/unix/exec.rb
rick: Sooo... having never looked at this code before, this appears to be doing a command execution with an arbitrary unescaped+unquoted parameter embedded.

In the original, if I call `get_env_key("whatever ; rm -rf /")`, or in the modified version, if I call `get_env_key("whatever ; rm -rf / ; # ")` then presumably the caller can execute whatever nefarious command they choose.

I don't know all the places where this is called, and what the threat model might be, but looks kinda wide open.  Am I missing something?

cc @puppetlabs/beaker

kevpl: For @ferglor & @glennsarti:
I agree with @glennsarti that it could definitely be better, but is outside the scope of this PR.

For @rick:
This is wide open, but only leaves a particular SUT open to arbitrary command execution, since it is a host command. Since for the most part SUTs are virtualized, no one tends to care about hosing one very much.

With PR testing, I could see someone submitting code to demolish a SUT, but the effect of that is minimal. If we were to hunt down a command that runs on the beaker coordinator itself, I could see that doing potentially much more damage in today's world, since our Jenkins slaves tend to be static hosts. Note that I'm not pretending that any of those commands are secured, I don't think this would be a difficult task.

My recommendation here is to have a whitelist like beaker does on when Jenkins will run your acceptance tests, and use discretion before signalling that it's OK for Jenkins to pick up new code to run.

Also in the next iteration of CI, the Jenkins slaves themselves will by dynamic and throw-away. Since that's the case, and that's how we would recommend someone setup their Jenkins to run beaker testing, this danger will be mitigated soon.

Iristyle: I also wouldn't worry about the injection in this case.  If you have ability to exec arbitrary commands already on a host that you have root / Administrator access to - is there really any increased risk?

kevpl:
last comment is still current: spec tests are needed, regardless of which release this goes into.

glennsarti:
Will add tests when I get time

glennsarti:
@kevpl Spec tests could not be written to test this behaviour.  I've added additional acceptance tests instead.

McdonaldSeanp:
:+1:

glennsarti:
@tvpartytonight @kevpl Acceptance tests have been added.  Ok for merge?

PR #1225, opened 2016-08-19
(BKR-916) Use dns_name for ec2 hostnames

tvpartytonight:
Prior to this change, provisioning in ec2 would set each host's hostname
to the `name` attribute of the host. This change removes that behavior
and sets each host's hostname to be the public facing `dns_name`
provided by ec2 to allow traffic from outside ec2 to route correctly to
that host.

### DIFF
--- a/lib/beaker/hypervisor/aws_sdk.rb
+++ b/lib/beaker/hypervisor/aws_sdk.rb
@@ -630,19 +630,21 @@ module Beaker
       @logger.notify("netscaler: nsroot password is #{host['instance'].id}")
     end
 
-    # Set the hostname of all instances to be the hostname defined in the
-    # beaker configuration.
+    # Set the :vmhostname for each host object to be the dns_name, which is accessible
+    # publicly. Then configure each ec2 machine to that dns_name, so that when facter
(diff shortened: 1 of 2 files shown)


tvpartytonight:

tvpartytonight:

doug-rosser:
Is there a way to override this behavior? Classification for scale testing is almost always set by hostname/certname, which I need to be able to make unique for each role.

tvpartytonight:
@doug-rosser I am a little confused on your requirement to 

> make (hostname/certname) unique for each role.

This change doesn't alter that aspect; all machines will still have unique certnames, but they will differ as the certnames will be based on their externally available names instead of internally available names.

For instance, before this change, after beaker provisioned the host, it would have a hostname like `aws2.us-west-2.compute.internal`, where `aws2` is the name of the host from your beaker host configuration. With this change, the hostname would be something like `ec2-52-43-22-235.us-west-2.compute.amazonaws.com`.

I'm unclear how that change breaks your requirements; can you elaborate?

doug-rosser:
Ah, in that case 👍 ! I was thinking this would set the hostnames to the garbage ec2 auto-assigned.

PR #1227, opened 2016-08-23
(BKR-221) Beaker assumes apache2ctl for graceful restarting

ferglor:


ferglor:

kevpl:
@ferglor shouldn't it be the case that with the code change as you made it, this should only work for redhat-7? I would assume that other platforms need the name to remain the same. Is this not the case?

If I'm right about the above, then it looks like our current testing doesn't cover this scenario. If that's the case, would you add some acceptance testing to expose this?

kevpl:
last comment is still current.

ferglor:
Yeah so before I made this change I was checking a few different distros for the presence of apachectl vs apache2ctl. On the likes of Debian and Ubuntu, apachectl is symlinked to apache2ctl. On other systems like Fedora, Oracle and Cent OS, apache2ctl isn't available, but apachectl is. So I decided to push up this PR to test the changes against a bunch of OSes. I'll look into the acceptance tests and see what coverage we're missing. Thanks!

ferglor:
@kevpl @tvpartytonight @pinkypie the pe-acceptance tests for 3.3.2 don't seem to setup passenger to be used in place of puppet server. Can we merge this for now and hope it doesn't break anything down the line?

lib/beaker/dsl/helpers/puppet_helpers.rb
```diff
@@ -323,7 +323,7 @@ def bounce_service host, service, curl_retries = nil, port = nil
           curl_retries = 120 if curl_retries.nil?
           port = options[:puppetserver_port] if port.nil?
           if host.graceful_restarts?
-            apachectl_path = host.is_pe? ? "#{host['puppetsbindir']}/apache2ctl" : 'apache2ctl'
+            apachectl_path = host.is_pe? ? "#{host['puppetsbindir']}/apachectl" : 'apachectl'
```
joshcooper: Opensource puppet tests against passenger packages here but only for the Debian wheezy and trusty cells: https://jenkins.puppetlabs.com/view/legacy_Puppet/view/3.x/job/legacy_puppet_integration-system_linux_3.x/ We'd want to make sure `apachectl` is present on those systems before merging this.

That said I'd be more concerned about module pipelines that might be testing against a larger set of supported Debian platforms and whether `apachectl` is always present?

A safer change might be to test if the `apache2ctl` exists on the host, and if not, fall back to `apachectl`.

kevpl: @ferglor I'm sorry for delaying this further, but this check does sound like a better idea. There's a [`host.check_for_command`](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host/unix/pkg.rb#L10-L22) method that would be useful for this.

ferglor:
@puppetlabs-jenkins retest this cholo

ferglor:

PR #1228, opened 2016-08-23
(PE-17093) Fix noask directory path

highb:
Prior to this commit I attempted to use a variable that doesn't exist
`noask_directory`.
This commit uses the `opts[:copy_dir_external]` varible instead, which
should be provided by beaker, and default to `/root`

### DIFF
--- a/lib/beaker/host/unix/pkg.rb
+++ b/lib/beaker/host/unix/pkg.rb
@@ -95,8 +95,9 @@ module Unix::Pkg
         if ! check_for_command('pkgutil')
           # https://www.opencsw.org/package/pkgutil/
           noask_text = self.noask_file_text
-          create_remote_file self, File.join(noask_directory, 'noask'), noask_text
-          execute('pkgadd -d http://get.opencsw.org/now -a noask -n all', opts)
+          noask_file = File.join(opts[:copy_dir_external], 'noask')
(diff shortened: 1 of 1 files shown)


demophoon:
😠
👍
🔥

tvpartytonight:

lib/beaker/host/unix/pkg.rb
```diff
@@ -95,8 +95,9 @@ def install_package(name, cmdline_args = '', version = nil, opts = {})
         if ! check_for_command('pkgutil')
           # https://www.opencsw.org/package/pkgutil/
           noask_text = self.noask_file_text
-          create_remote_file self, File.join(noask_directory, 'noask'), noask_text
-          execute('pkgadd -d http://get.opencsw.org/now -a noask -n all', opts)
+          noask_file = File.join(opts[:copy_dir_external], 'noask')
+          create_remote_file(self, noask_file, noask_text)
+          execute("pkgadd -d http://get.opencsw.org/now -a #{noask_file} -n all", hpts)
```
kevpl: is this supposed to be `hpts`?

kevpl:
@highb would you add a spec test so that we can get this code path exercised & make sure we don't break it going forward?

demophoon:
I'm working on this in #1238 while Brandon is out.

PR #1229, opened 2016-08-24
(maint) Add security group option to openstack hypervisor initialization

lucywyman:
Because Puppet has it's own security group that is not default but is necessary to connect to an openstack instance, I'd like to add a security group option to the hosts config file so that the user can specify security group(s) to add to an instance on creation.

This is just an initial commit. I'm not sure how to test or verify the work I've done, so any help in that area would be excellent!

### DIFF
--- a/docs/how_to/hypervisors/openstack.md
+++ b/docs/how_to/hypervisors/openstack.md
@@ -16,7 +16,6 @@ Get openstack Access & Security credentials:
 - "openstack_tenant"
 - "openstack_network"
 - "openstack_keyname"
-- "security_group" (optional)
 
 If you are using [OpenStack Dashboard "Horizon"] (https://wiki.openstack.org/wiki/Horizon) 
 you can find these keys in next places:
@@ -28,8 +27,6 @@ you can find these keys in next places:
 2. "openstack_network": in "project" -> "Networks"
 3. "openstack_keyname": in "project" -> "Compute" -> "Access & Security" -> tab "Key Pairs"
 4. "openstack_api_key": Your user Password
-5. "security_group": in "project" -> "Compute" -> "Access & Security"
--> tab "Security Groups"
 
 ### Setup a Openstack Hosts File
 
@@ -60,7 +57,6 @@ in order for the Openstack hypervisor to provision hosts properly.
       openstack_tenant: testing
       openstack_network : testing
       openstack_keyname : nopass
-      security_group    : ['sg0', 'default']
 
 The `image` - image name.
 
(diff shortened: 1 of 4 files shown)


kevpl:
Hi @lucywyman! Thanks for submitting this new feature!

In terms of testing this, I would start by mocking `@compute_client.security_groups`. Once you've got that, then it looks like you can test different scenarios of having your `security_groups` list against different parameters listed, asserting against their returns.

If you would either create a ticket for this, or amend your git commit with the `(MAINT)` prefix, we'll be on our way to meeting the [contribution guidelines](https://github.com/puppetlabs/beaker/blob/master/CONTRIBUTING.md). Lastly, if you would provide a blurb about this functionality in the [openstack doc](https://github.com/puppetlabs/beaker/blob/master/docs/how_to/hypervisors/openstack.md), that would be great to help people see how they should use this without needing to read code.

lucywyman:
@kevpl Awesome, can do on the docs and testing.  I added a few more lines further down the file to deal with assigning the appropriate hostname to the created instance so that we can install PE on a SLICE VM in an automated way.  The biggest problem is that the VM can't be ssh'd into until it's been assigned a floating IP and the associated DNS record, which makes it hard to verify that the VM is actually fully created.  The current 2 sleeps work fine for the super-happy-path on my machine, but I've got a few more iterations to go to verifying that the machine is up by trying to assign it an IP.  Is that going to be acceptable?  Let me know if you're interested in pairing on this and I'll try to find a one-on-one time this week. 

Thanks for your patience!

lucywyman:
@kevpl  I added some spec tests, but do I need to fully mock the object like how images and flavors are?  If not, I think this PR is ready to go.

docs/how_to/hypervisors/openstack.md
```diff
@@ -57,6 +60,7 @@ in order for the Openstack hypervisor to provision hosts properly.
       openstack_tenant: testing
       openstack_network : testing
       openstack_keyname : nopass
+      security_group    : ['sg0', 'default']
```
kevpl: we would prefer it if we not clutter up the basic openstack hosts file with optional settings, because they're hard enough to grok as is.  Please include this as it's own heading under "Openstack-Specific Hosts File Settings." Rather than include a new copy of the hosts file, as done in the user-data section, I think it's sufficient to say that you include the setting in your `CONFIG` section, and it should look like so:

    security_group    : ['sg0', 'default']

lucywyman:
@kevpl  Thanks so much for looking at this!  Does that look a little more reasonable for the docs?

PR #1232, opened 2016-08-30
(BKR-888) install_puppet_agent_dev_repo_on prohibits passing different platforms

ferglor:


kevpl:
@ferglor I'm confused as to how this fixes the issue @johnduarte raises in the ticket.

It seems here that anything passed through the `opts` parameter still overrides anything in the host itself. @johnduarte, I'm assuming you want some kind of situation where the host overrides the `opts`? Is there every a situation where you would want the `opts` parameter to be used?

johnduarte:
@ferglor  this does not fix the issue. The constructed opts for first host still trashes the ability to install the repo on the next host. 

@kevpl The `opts` values passed into the method should override the host. I would never expect the host to override the passed in `opts` value.

The problem is that the `opts` values are constructed based on the first host and are not updated for the second host. Lets look at what happens internally to the `opts` hash when passed two hosts (1 windows, 1 osx) and an `opts` value of `{ :puppet_collection=>"PC1", :puppet_agent_sha=>"1.5.2", :puppet_agent_version=>"1.5.2"}`

During the first round of the `block_on hosts`:
```
host['platform'].to_array
=> ["windows", "2012r2", "64", nil]

opts => {:win_download_url=>"http://downloads.puppetlabs.com/windows",
 :mac_download_url=>"http://downloads.puppetlabs.com/mac",
 :pe_promoted_builds_url=>"http://pm.puppetlabs.com",
 :release_apt_repo_url=>"http://apt.puppetlabs.com",
 :release_yum_repo_url=>"http://yum.puppetlabs.com",
 :dev_builds_url=>"http://builds.delivery.puppetlabs.net",
 :puppet_collection=>"PC1",
 :puppet_agent_sha=>"1.5.2",
 :puppet_agent_version=>"1.5.2",
 :download_url=>"http://builds.delivery.puppetlabs.net/puppet-agent/1.5.2/repos/",
 :copy_base_local=>"tmp/repo_configs",
 :copy_dir_external=>"C:/PROGRA~3/"}

```
During the second round of the `block_on hosts`:
```
host['platform'].to_array
=> ["osx", "10.11", "x86_64", "elcapitan"]

opts => {:win_download_url=>"http://downloads.puppetlabs.com/windows",
 :mac_download_url=>"http://downloads.puppetlabs.com/mac",
 :pe_promoted_builds_url=>"http://pm.puppetlabs.com",
 :release_apt_repo_url=>"http://apt.puppetlabs.com",
 :release_yum_repo_url=>"http://yum.puppetlabs.com",
 :dev_builds_url=>"http://builds.delivery.puppetlabs.net",
 :puppet_collection=>"PC1",
 :puppet_agent_sha=>"1.5.2",
 :puppet_agent_version=>"1.5.2",
 :download_url=>"http://builds.delivery.puppetlabs.net/puppet-agent/1.5.2/repos/",
 :copy_base_local=>"tmp/repo_configs",
 :copy_dir_external=>"C:/PROGRA~3/"}
```
As you can see, the value of `opts[:copy_dir_external]` is set to `C:/PROGRA~3/` for the first host and is unable to be refreshed for the second host.

To fix this problem, the internally used `opts` should be reconstructed fresh for each host based on the passed in `opts`. I would suggest doing this with a deep_copy to an internal variable.

----
Fixing the above would allow BKR-888 to be closed.

@kevpl The remaining impedance mismatch for me is the design where I might want to pass in an array of hosts, but cannot pass in an `opts` hash with values keyed to the members of that array. I don't think there is enough evidence to support changing the API, but I wanted to raise it as a usability issue with the method call.

ferglor:
@kevpl @johnduarte sorry guys, I've been distracted with BKR-221 so I'm only replying now. I believe that this change does in fact fix the issue at hand. So, before, we modified the opts hash with the external_copy_base from the first host object in the list of hosts passed to the function:

opts[:copy_dir_external]  ||= host.external_copy_base

I.e., we only set copy_dir_external in the opts hash if it currently is nil, using the external_copy_base off the first host we iterate over. For subsequent iterations, this assignment won't be repeated, because copy_dir_external in the opts hash is no longer nil. Later in the function, we then read this value off as follows:

onhost_copy_base = opts[:copy_dir_external]

My modification solves this issue with the following line:

onhost_copy_base = opts[:copy_dir_external] || host.external_copy_base

I.e. for each iteration over the list of hosts, we first check the opts hash for copy_dir_external, and if it hasn't been defined, we read the external_copy_base off the host in the current iteration. This means that if copy_dir_external has been specified in the opts hash, this will always override the external_copy_base from the host objects. If, however, copy_dir_external has not been specified in the opts hash, we read the external_copy_base off the host for the current iteration - hence permitting subsequent hosts to use different external_copy_base values.

johnduarte:
Thanks @ferglor ! I have confirmed that this works for the stated case in the ticket. I must not have patched correctly before. My apologies.
:+1:

pinkypie:
👍

PR #1233, opened 2016-08-30
(BKR-712) changed default to not cache files locally

kevpl:


spec/beaker/dsl/helpers/web_helpers_spec.rb
```diff
@@ -36,21 +36,43 @@ def fetch_allows
       it "returns its second and third arguments concatenated." do
         concat_path = "#{destdir}/#{name}"
         create_files([concat_path])
-        expect( logger ).to receive( :notify ).with( /^Already\ fetched\ / )
```
tvpartytonight: It doesn't seem terribly important here, but why is the `with` rspec method removed here?

kvimber: The behavior flipped, so "Already fetched" changed code paths to the "Fetching and saving to" notifications.

This is not what this spec test is saying it tests, however, so I stopped it from testing that, since we do have tests for it below.

tvpartytonight: Yeah, I expected the `create_files` method here to touch the file and then cause the logger to receive that `:notify`, so it was a little bit confusing, but I agree that we don't need the expect to be that specific here.

PR #1236, opened 2016-09-07
(QA-2599) Allocate IP if none is available

lucywyman:
This PR:
* Moves the logic of getting or allocating a floating IP to it's own function
* Adds allocation functionality for if no floating IPs are available

I realized that in my last PR, I bypassed the error handling for when no floating IPs are available. This puts it back in.

### DIFF
--- a/docs/how_to/hypervisors/openstack.md
+++ b/docs/how_to/hypervisors/openstack.md
@@ -114,3 +114,12 @@ in the `CONFIG` section of your hosts file:
     security_group: ['my_sg', 'default']
 
 This is an optional config parameter.
+
+### Floating IP Pool
+
+The name of the floating IP pool that a VM can grab IPs from. This is useful
+if your organization doesn't have a public pool of floating IPs, or give each
+user their own pool.  It's used in allocating new IPs.  It's an options
+parameter in the CONFIG section of the host file:
+
+    floating_ip_pool: 'my_pool_name'
--- a/lib/beaker/hypervisor/openstack.rb
+++ b/lib/beaker/hypervisor/openstack.rb
@@ -171,15 +171,25 @@ module Beaker
       end
     end
 
+    # Get a floating IP address to associate with the instance, and
+    # allocate a new one if none are available
     def get_ip
+      @logger.debug "Finding IP"
       ip = @compute_client.addresses.find { |ip| ip.instance_id.nil? }
       if ip.nil?
-        @logger.debug "Creating IP"
-        ip = @compute_client.addresses.create
+        begin
+          @logger.debug "Creating IP"
+          ip = @compute_client.addresses.create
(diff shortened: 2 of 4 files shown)


lucywyman:
Sorry for all the noise, but this PR is definitely ready for review now. 

I didn't want to add *yet another option* to the CONFIG section, but I couldn't find a better way to allocate new addresses. The vm object isn't available in the get_ip function, and the addresses.create didn't seem to work at all :/

lib/beaker/hypervisor/openstack.rb
```diff
@@ -223,36 +233,9 @@ def provision
...
-            end
-          rescue NoMethodError
-            @logger.debug "No current address retrievable from OpenStack data"
-          end
-
-        end
-
-        raise 'Could not find or assign an address to the instance' unless address
-        host[:ip] = address
+        ip.ip.gsub! '-','.'
```
kevpl: It seems that just after the `get_ip` call on line 199, we're setting the ip to have dashes rather than dots (or vice versa), and then setting them back here to be set correctly as a host attribute. Since we're assigning the changed IP address to the `hostname` variable, couldn't we just take off the `!` above, and then we'd be able to delete this line?

docs/how_to/hypervisors/openstack.md
```diff
@@ -114,3 +114,12 @@ in the `CONFIG` section of your hosts file:
...
 This is an optional config parameter.
+
+### Floating IP Pool
+
+The name of the floating IP pool that a VM can grab IPs from. This is useful
+if your organization doesn't have a public pool of floating IPs, or give each
+user their own pool.  It's used in allocating new IPs.  It's an options
+parameter in the CONFIG section of the host file:
+
+    floating_ip_pool: 'my_pool_name'
```
kevpl: yay docs! 💯 thank you!

kevpl:
other than the question in-line, I'm 👍 on this change. Once that's resolved, we'll look at getting this merged. Thanks for the fix!

lucywyman:
M'k -- fixed the gsub!  I can't remember why I included it in the first place :P  thanks so much @kevpl!

lib/beaker/hypervisor/openstack.rb
```diff
@@ -171,23 +171,33 @@ def cleanup_storage vm
...
+      @logger.debug "Finding IP"
       ip = @compute_client.addresses.find { |ip| ip.instance_id.nil? }
       if ip.nil?
-        @logger.debug "Creating IP"
-        ip = @compute_client.addresses.create
+        begin
+          @logger.debug "Creating IP"
+          ip = @compute_client.addresses.create
+        rescue Fog::Compute::OpenStack::NotFound
+          # If there are no more floating IP addresses, the allocate a
```
tvpartytonight: should this be "then allocate" rather than "the allocate"?

PR #1239, opened 2016-09-09
(PA-514) Ensure windows reg value & startup type

MosesMendoza:
This commit adds two checks to make sure that after install our Windows agents
still have the appropriate registry value setting & location for
RememberedPuppetAgentStartupMode (which is Disabled) and that the service does
in fact have a startup type of Disabled. This is after a regression was
introduced in the MSI packaging which wrote these values to the wrong location
and then they were not retained/honored after upgrade.

The MSI flag for setting startup mode is PUPPET_AGENT_STARTUP_MODE, and it
accepts three values:

* Automatic
* Manual
* Disabled

Of these three, "Automatic" is the only on that does not correspond directly to
the equivalent value returned by `WMIC SERVICE` - for this case, `WMIC SERVICE`
returns "Auto" and not "Automatic. "Manual" and "Disabled" are both returned
as-is. For this reason, Automatic is treated as a special case in this commit.

This location (beaker) is chosen for this test because the puppet
agent testing pipeline does not include its own testing for package-based
upgrades, and at this time all systems that install MSIs will use this util. At
such time as package acceptance testing of upgrades is added to puppet-agent,
these checks should be moved there.

Paired with: Sean McDonald

Signed-off-by: Moses Mendoza moses@puppet.com

### DIFF
--- a/lib/beaker/dsl/install_utils/windows_utils.rb
+++ b/lib/beaker/dsl/install_utils/windows_utils.rb
@@ -141,6 +141,16 @@ exit /B %errorlevel%
             # therefore, this command will always exit 0 if either service is installed
             on host, Command.new("sc query puppet || sc query pe-puppet", [], { :cmdexe => true })
 
+            # (PA-514) value for PUPPET_AGENT_STARTUP_MODE should be present in
+            # registry and honored after install/upgrade.
+            reg_query_command = %Q(reg query "HKLM\\SOFTWARE\\Wow6432Node\\Puppet Labs\\PuppetInstaller" /v "RememberedPuppetAgentStartupMode" | findstr #{msi_opts['PUPPET_AGENT_STARTUP_MODE']})
+            on host, Command.new(reg_query_command, [], { :cmdexe => true })
+
+            start_mode = msi_opts['PUPPET_AGENT_STARTUP_MODE'] == "Automatic" ? "Auto" : msi_opts['PUPPET_AGENT_STARTUP_MODE']
+            service_query_command = %Q('WMIC SERVICE where (Name like "%Puppet" AND StartMode="#{start_mode}") | findstr Puppet')
+
+            on host, Command.new(service_query_command, [], { :cmdexe => true })
+
             # emit the misc/versions.txt file which contains component versions for
             # puppet, facter, hiera, pxp-agent, packaging and vendored Ruby
             [
--- a/spec/beaker/dsl/install_utils/windows_utils_spec.rb
+++ b/spec/beaker/dsl/install_utils/windows_utils_spec.rb
@@ -61,6 +61,16 @@ describe ClassMixedWithDSLInstallUtils do
     end
   end
 
+  def expect_reg_query_called(times = hosts.length)
+    expect( Beaker::Command ).to receive( :new )
+      .with(%r{reg query "HKLM\\SOFTWARE\\Wow6432Node\\Puppet Labs\\PuppetInstaller}, [], {:cmdexe => true})
+      .exactly(times).times
+  end
+
+  def expect_startup_query_called(times = hosts.length)
+    expect( Beaker::Command ).to receive( :new ).with(/WMIC SERVICE.*StartMode=/, [], {:cmdexe => true}).exactly(times).times
+  end
+
   describe "#install_msi_on" do
     let( :log_file )    { '/fake/log/file.log' }
     before :each do
@@ -71,6 +81,8 @@ describe ClassMixedWithDSLInstallUtils do
     it "will specify a PUPPET_AGENT_STARTUP_MODE of Manual (disabling the service) by default" do
       expect_install_called
       expect_status_called
+      expect_reg_query_called
+      expect_startup_query_called
       expect_version_log_called
       expect( subject ).to receive( :create_install_msi_batch_on ).with(
           anything, anything,
@@ -81,6 +93,8 @@ describe ClassMixedWithDSLInstallUtils do
     it "allows configuration of PUPPET_AGENT_STARTUP_MODE" do
       expect_install_called
       expect_status_called
+      expect_reg_query_called
+      expect_startup_query_called
       expect_version_log_called
       value = 'Automatic'
       expect( subject ).to receive( :create_install_msi_batch_on ).with(
@@ -109,6 +123,8 @@ describe ClassMixedWithDSLInstallUtils do
 
     it "will generate a command to emit a log file with the :debug option set" do
       expect_install_called
+      expect_reg_query_called
+      expect_startup_query_called
       expect_status_called
       expect_version_log_called
 
@@ -118,6 +134,8 @@ describe ClassMixedWithDSLInstallUtils do
 
     it 'will pass msi_path to #create_install_msi_batch_on as-is' do
       expect_install_called
+      expect_reg_query_called
+      expect_startup_query_called
       expect_status_called
       expect_version_log_called
       test_path = 'test/path'
@@ -125,6 +143,30 @@ describe ClassMixedWithDSLInstallUtils do
           anything, test_path, anything)
       subject.install_msi_on(hosts, test_path)
     end
+
+    it 'will confirm service startup type is honored via WMIC Service' do
+      expect_install_called
+      expect_reg_query_called
+      expect_status_called
+      expect_version_log_called
+
+      expect( Beaker::Command ).to receive( :new )
+        .with(/WMIC SERVICE.*StartMode="Foo"/, [], {:cmdexe => true})
+        .exactly(hosts.length).times
+      subject.install_msi_on(hosts, msi_path, {'PUPPET_AGENT_STARTUP_MODE' => "Foo"})
+    end
+
+    it 'will check the registry for remembered startup setting via reg query' do
+      expect_install_called
+      expect_status_called
+      expect_startup_query_called
+      expect_version_log_called
+
+      expect( Beaker::Command ).to receive( :new )
+        .with('reg query "HKLM\\SOFTWARE\\Wow6432Node\\Puppet Labs\\PuppetInstaller" /v "RememberedPuppetAgentStartupMode" | findstr Foo', [], {:cmdexe => true})
+        .exactly(hosts.length).times
+        subject.install_msi_on(hosts, msi_path, {'PUPPET_AGENT_STARTUP_MODE' => "Foo"})
+      end
   end
 
   describe '#create_install_msi_batch_on' do


MosesMendoza:
A checkout of this PR/version of beaker was successfully used with frankenbuilder to exercise the PE Windows upgrade tests, with this invocation:
```
be ./frankenbuilder 2016.3 --keyfile=~/.ssh/id_rsa-acceptance --agent windows-2008-64 --vmpooler --upgrade-from 2016.1.2 --install --agent-upgrade --pe-acceptance-tests ~/development/pe_acceptance_tests
```

The output from the run is here:
https://gist.github.com/MosesMendoza/85f3e219a2b6d6994858dc7f389e09bb

Directly relevant output is
```
  m3q70qx5yzlrarf.delivery.puppetlabs.net (windows-2008-64-2) 11:15:48$ cmd.exe /c reg query "HKLM\SOFTWARE\Wow6432Node\Puppet Labs\PuppetInstaller" /v "RememberedPuppetAgentStartupMode" | findstr Manual
        RememberedPuppetAgentStartupMode    REG_SZ    Manual

  m3q70qx5yzlrarf.delivery.puppetlabs.net (windows-2008-64-2) executed in 0.16 seconds

  m3q70qx5yzlrarf.delivery.puppetlabs.net (windows-2008-64-2) 11:15:48$ cmd.exe /c 'WMIC SERVICE where (Name="Puppet" AND StartMode="Manual") | findstr Puppet'
    FALSE        FALSE       Puppet Agent  0           Win32_Service      Periodically fetches and applies configurations from a Puppet master server.  FALSE            Puppet Agent  Normal        1077                   puppet  "C:\Program Files\Puppet Labs\Puppet\sys\ruby\bin\ruby.exe" -rubygems "C:\Program Files\Puppet Labs\Puppet\service\daemon.rb"  0          0                        Own Process  FALSE    Manual     LocalSystem  Stopped  OK      Win32_ComputerSystem     M3Q70QX5YZLRARF  0      0

  m3q70qx5yzlrarf.delivery.puppetlabs.net (windows-2008-64-2) executed in 0.60 seconds
```

MosesMendoza:
This was also run against a contrived case with `PUPPET_AGENT_STARTUP_MODE` set to `Automatic`, with this relevant output confirming conversation of `Automatic` to `Auto` in WMIC:
```
 kbkx9n0bnqk11wn.delivery.puppetlabs.net (windows-2008-64-2) 12:10:17$ cmd.exe /c "C:\Windows\TEMP\install-puppet-msi-2016-09-09_12.10.16.bat"

    C:\cygwin64\home\Administrator>start /w msiexec.exe /i "C:\PROGRA~3\puppet-agent-x64.msi" /qn /L*V C:\Windows\TEMP\install-puppet-2016-09-09_12.10.16.log PUPPET_AGENT_STARTUP_MODE=Automatic

    C:\cygwin64\home\Administrator>exit /B 0

  kbkx9n0bnqk11wn.delivery.puppetlabs.net (windows-2008-64-2) executed in 23.87 seconds

  kbkx9n0bnqk11wn.delivery.puppetlabs.net (windows-2008-64-2) 12:10:41$ cmd.exe /c sc query puppet || sc query pe-puppet

    SERVICE_NAME: puppet
            TYPE               : 10  WIN32_OWN_PROCESS
            STATE              : 4  RUNNING
                                    (STOPPABLE, PAUSABLE, ACCEPTS_SHUTDOWN)
            WIN32_EXIT_CODE    : 0  (0x0)
            SERVICE_EXIT_CODE  : 0  (0x0)
            CHECKPOINT         : 0x0
            WAIT_HINT          : 0x0

  kbkx9n0bnqk11wn.delivery.puppetlabs.net (windows-2008-64-2) executed in 0.50 seconds

  kbkx9n0bnqk11wn.delivery.puppetlabs.net (windows-2008-64-2) 12:10:41$ cmd.exe /c reg query "HKLM\SOFTWARE\Wow6432Node\Puppet Labs\PuppetInstaller" /v "RememberedPuppetAgentStartupMode" | findstr Automatic
        RememberedPuppetAgentStartupMode    REG_SZ    Automatic

  kbkx9n0bnqk11wn.delivery.puppetlabs.net (windows-2008-64-2) executed in 0.29 seconds

  kbkx9n0bnqk11wn.delivery.puppetlabs.net (windows-2008-64-2) 12:10:41$ cmd.exe /c 'WMIC SERVICE where (Name="Puppet" AND StartMode="Auto") | findstr Puppet'
    TRUE         TRUE        Puppet Agent  0           Win32_Service      Periodically fetches and applies configurations from a Puppet master server.  FALSE            Puppet Agent  Normal        0                      puppet  "C:\Program Files\Puppet Labs\Puppet\sys\ruby\bin\ruby.exe" -rubygems "C:\Program Files\Puppet Labs\Puppet\service\daemon.rb"  2388       0                        Own Process  TRUE     Auto       LocalSystem  Running  OK      Win32_ComputerSystem     KBKX9N0BNQK11WN  0      0

  kbkx9n0bnqk11wn.delivery.puppetlabs.net (windows-2008-64-2) executed in 1.51 seconds
```

MosesMendoza:
Ooh apparently I broke tests. Investigating.

MosesMendoza:

tvpartytonight:
Moses, this change looks good to me; the failure you see in CI is due to RE-8035. Whence that is resolved, we should kick off the tests again.

lib/beaker/dsl/install_utils/windows_utils.rb
```diff
@@ -141,6 +141,16 @@ def install_msi_on(hosts, msi_path, msi_opts = {}, opts = {})
             # therefore, this command will always exit 0 if either service is installed
             on host, Command.new("sc query puppet || sc query pe-puppet", [], { :cmdexe => true })
 
+            # (PA-514) value for PUPPET_AGENT_STARTUP_MODE should be present in
+            # registry and honored after install/upgrade.
+            reg_query_command = %Q(reg query "HKLM\\SOFTWARE\\Wow6432Node\\Puppet Labs\\PuppetInstaller" /v "RememberedPuppetAgentStartupMode" | findstr #{msi_opts['PUPPET_AGENT_STARTUP_MODE']})
+            on host, Command.new(reg_query_command, [], { :cmdexe => true })
+
+            start_mode = msi_opts['PUPPET_AGENT_STARTUP_MODE'] == "Automatic" ? "Auto" : msi_opts['PUPPET_AGENT_STARTUP_MODE']
```
tvpartytonight: Should this `PUPPET_AGENT_STARTUP_MODE` have a corresponding yardoc change?

MosesMendoza: @tvpartytonight I'm not quite sure what you mean. Could you elaborate?

tvpartytonight: @MosesMendoza I was confused about the valid key; I thought that the valid key in `msi_opts` should be `Auto` instead of `Automatic`, and thus the yardoc for the method would need a change. However, I see now that it is defined locally as `start_mode` and thus there is no need to for this change.

tvpartytonight:

kevpl:

MosesMendoza:
thanks to @tvpartytonight for pointing out the legit failure in that last run. On PE <= 3.8.x we install a service on windows called 'pe-puppet' but this PR only looks for service 'puppet'. Will update to look for either, assuming this version of beaker needs to support testing against PE 3.8.

MosesMendoza:
@tvpartytonight looks green!

tvpartytonight:
LGTM! 👍 
windowswindowswindows

lib/beaker/dsl/install_utils/windows_utils.rb
```diff
@@ -141,6 +141,16 @@ def install_msi_on(hosts, msi_path, msi_opts = {}, opts = {})
             # therefore, this command will always exit 0 if either service is installed
             on host, Command.new("sc query puppet || sc query pe-puppet", [], { :cmdexe => true })
 
+            # (PA-514) value for PUPPET_AGENT_STARTUP_MODE should be present in
+            # registry and honored after install/upgrade.
+            reg_query_command = %Q(reg query "HKLM\\SOFTWARE\\Wow6432Node\\Puppet Labs\\PuppetInstaller" /v "RememberedPuppetAgentStartupMode" | findstr #{msi_opts['PUPPET_AGENT_STARTUP_MODE']})
```
Iristyle: I don't think that this is quite right since we also need to look in a different location on a 32-bit OS

```
HKLM\Software\Puppet Labs\PuppetInstaller
```

https://github.com/puppetlabs/puppet_for_the_win/blob/ae1821b2ff7ab6c417bfb03b89ab3097be3e05fa/README.markdown#remembered-properties

Iristyle: After a bit of archaeology from @MosesMendoza, looks like our best bet is to call `is_x86_64?` on the host (https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host.rb#L269-L271), to then determine which reg key to query.

We'll make sure to run the new PR against 32-bit and 64-bit Windows hosts to verify.

lib/beaker/dsl/install_utils/windows_utils.rb
```diff
@@ -141,6 +141,16 @@ def install_msi_on(hosts, msi_path, msi_opts = {}, opts = {})
...
+            on host, Command.new(reg_query_command, [], { :cmdexe => true })
+
+            start_mode = msi_opts['PUPPET_AGENT_STARTUP_MODE'] == "Automatic" ? "Auto" : msi_opts['PUPPET_AGENT_STARTUP_MODE']
+            service_query_command = %Q('WMIC SERVICE where (Name like "%Puppet" AND StartMode="#{start_mode}") | findstr Puppet')
```
Iristyle: I think that we might be able to remove the wildcard checking for Puppet with WMIC... and just execute against `sc` like the service running / installed check.  Then we can be a bit more precise with matching the startup type like:

```
(sc qc puppet || sc qc pe-puppet) | findstr /C:"START_TYPE         : 4   DISABLED"
```

Sample results:

```
C:\Users\Administrator>(sc qc foo || sc qc puppet) | findstr /C:"START_TYPE    : 4   DISABLED"
        START_TYPE         : 4   DISABLED

C:\Users\Administrator>echo %ERRORLEVEL%
0

C:\Users\Administrator>(sc qc foo || sc qc bar) | findstr /C:"START_TYPE    : 4   DISABLED"

C:\Users\Administrator>echo %ERRORLEVEL%
1
```

That said, I think it might be better just to use the command output from above to regex search inside of Ruby (I would advocate for using `findstr` at command line, except when you do that, it eats all the output normally logged to Beaker):

So I think you could remove the above, and instead modify `sc query` to be `sc qc` a few lines above, and then use that output to string parse in Ruby code:

```ruby
output = on host, Command.new("sc qc puppet || sc qc pe-puppet", [], { :cmdexe => true })
case msi_opts['PUPPET_AGENT_STARTUP_MODE'].upcase
when 'AUTOMATIC'
  # there is also a value 1 / SYSTEM_START
  search = { :code => 2, :name => 'AUTO_START' }
when 'MANUAL'
  search = { :code => 3, :name => 'DEMAND_START' }
when 'DISABLED'
  search = { :code => 4, :name => 'DISABLED' }
end
output ~= /^\s+START_TYPE\s+:\s+#{search[:code]}\s+#{search[:name]}/
```

That avoids a round-trip to the host, and allows for some more flexible regex parsing in Ruby.

MosesMendoza: @Iristyle just to confirm - is our evaluation of the user supplied value for `PUPPET_AGENT_STARTUP_MODE` case-insensitive? i.e. user can supply Disabled or DISABLED ?

Iristyle:
Hah, i now see you already had commented on this.

PR #1241, opened 2016-09-12
(BKR-931) Ensure that :type = 'aio' works in EL

trevor-vaughan:
EL and Fedora systems were not installing the pc1 version of Puppet
when :type = aio. This ensures that the case is properly addressed.

BKR-931 #close

### DIFF
--- a/lib/beaker/dsl/install_utils/foss_utils.rb
+++ b/lib/beaker/dsl/install_utils/foss_utils.rb
@@ -436,7 +436,11 @@ module Beaker
         # @api private
         def install_puppet_from_rpm_on( hosts, opts )
           block_on hosts do |host|
-            install_puppetlabs_release_repo(host)
+            if host[:type] == 'aio'
+              install_puppetlabs_release_repo(host,'pc1',opts)
+            else
+              install_puppetlabs_release_repo(host,nil,opts)
+            end
 
             if opts[:facter_version]
               host.install_package("facter-#{opts[:facter_version]}")
@@ -922,6 +926,7 @@ module Beaker
                 # cisco ios xr requires using yum to localinstall the repo
                 on host, "yum -y localinstall #{remote}"
               else
+                opts[:package_proxy] ||= false
                 host.install_package_with_rpm( remote, '--replacepkgs',
                   { :package_proxy => opts[:package_proxy] } )
               end
--- a/spec/beaker/dsl/install_utils/foss_utils_spec.rb
+++ b/spec/beaker/dsl/install_utils/foss_utils_spec.rb
@@ -57,6 +57,17 @@ describe ClassMixedWithDSLInstallUtils do
                                                 :working_dir => '/tmp',
                                                 :type => 'foss',
                                                 :dist => 'puppet-enterprise-3.7.1-rc0-78-gffc958f-eos-4-i386' } ) }
+  let(:el6hostaio)    { make_host( 'el6hostaio', { :platform => Beaker::Platform.new('el-6-i386'),
+                                                 :pe_ver => '3.0',
+                                                 :working_dir => '/tmp',
+                                                :type => 'aio',
+                                                 :dist => 'puppet-enterprise-3.1.0-rc0-230-g36c9e5c-centos-6-i386' } ) }
+  let(:el6hostfoss)   { make_host( 'el6hostfoss', { :platform => Beaker::Platform.new('el-6-i386'),
+                                                 :pe_ver => '3.0',
+                                                 :working_dir => '/tmp',
+                                                :type => 'foss',
+                                                 :dist => 'puppet-enterprise-3.1.0-rc0-230-g36c9e5c-centos-6-i386' } ) }
+
   let(:win_temp)      { 'C:\\Windows\\Temp' }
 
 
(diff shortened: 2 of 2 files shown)


ferglor:
@trevor-vaughan can you update the spec tests for this change?

trevor-vaughan:
Not any time soon. Took a look and there are *no* tests for this capability at all so I can't just do a minor update, it's going to actually take work.

It certainly works in anger :-D

kevpl:
last comment is still current.

kevpl:
last comment is still current.

kevpl:
last comment is still current. If there's no activity on this in a week, we'll close this PR.

trevor-vaughan:
@kevpl You close this test and AIO does *not* work in Beaker *as advertised*.

I *will* add tests after you add the plethora of missing tests around that bit of code. I can 100% guarantee that this works in production since you can't test AIO any other way.

kevpl:
@trevor-vaughan without either:

1. effort on your part to raise this PR to our standards
1. some other signals that this isn't working and needs fixing

I'm afraid that this doesn't have the priority level required for us to take it over and get this issue closed. Since that's the case, and it doesn't look like the recent push made any changes, I'm going to close this PR as inactive.

wyardley:
@kevpl:

https://tickets.puppetlabs.com/browse/BKR-970
https://github.com/puppetlabs/beaker-puppet_install_helper/issues/19
may partially be related.

Basically, on Ubuntu 16, it seems to default to trying to install Puppet 3; setting type 'aio' didn't seem to work for me, though admittedly I may have still been using puppet_install_helper at the time, but either way, I believe `install_puppet_on` fails on 16.04 with Vagrant as the provisioner, because it's trying to install Puppet 3, and there's no pc1 repo for Ubuntu 16.

So, I don't know _what_ the right fix(es) are, but I do think that something should be improved, even if it's changed defaults. It seems to me that a reworked version of this might also be appropriate.

trevor-vaughan:
@kevpl Can you point me to the original tests for host[:type] = 'aio' so that I can enhance them?

kevpl:
@wyardley [`install_puppet_on`](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/install_utils/foss_utils.rb#L256-L257) defaults to installing Puppet 3.x. You have to specify the `:version` option to get different behavior. I've created [BKR-977](https://tickets.puppetlabs.com/browse/BKR-977) to track changing this default in the future, hopefully at our next major version boundary.

@trevor-vaughan iirc all of our puppet agent install code is in `foss_utils.rb`, and all of the corresponding testing for that is in [`foss_utils_spec.rb`](https://github.com/puppetlabs/beaker/blob/master/spec/beaker/dsl/install_utils/foss_utils_spec.rb). Searching there for `aio` shows all of the tests for that that I'm aware of.

wyardley:
@kevpl Right, it works if you use`install_puppet_agent_on()` and / or use `install_puppet_on()` and set the install type or version appropriately. FYI, I also opened
https://tickets.puppetlabs.com/browse/BKR-970, and https://github.com/puppetlabs/beaker-puppet_install_helper/issues/19 about related issues (because the module I was working on was using puppet_install_helper vs. calling beaker directly).

I think the problem was, switching to agent install directly in the spec file won't work in the cases where Dockerized acceptance tests need to test Puppet v3, and I couldn't figure out how to set the env variables or otherwise configure the appropriate type / version for that nodeset in YAML vs. in an env variable (so that the default rake tasks would work even when the version isn't explicitly set). The code seems to suggest that CONFIG: type: agent should work.

Do you have a quick pointer to the specifications for :version (i.e., what it can accept vs. just an exact version to install, say, the latest)? http://www.rubydoc.info/github/puppetlabs/beaker/Beaker%2FDSL%2FInstallUtils%2FFOSSUtils%3Ainstall_puppet_on shows an example with just '4', will that install the latest v4 Puppet (the version specification seems different from some other puppet related tools)? Can that version be set in the acceptance nodeset when using `install_puppet_on()` in CONFIG: version?

I tried switching my module to use `install_puppet_on()` directly again and then set 
```
CONFIG:
  version: '4'
```
in `spec/acceptance/nodesets/ubuntu-server-1604-x64.yml`, and still get the same problem.

trevor-vaughan:
@kevpl I just took a look at the code in Beaker to test https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/install_utils/foss_utils.rb#L268-L283. Unfortunately, I couldn't find anything besides tests for {{install_puppet_from_freebsd_ports_on}} and {{install_puppet_from_msi}}.

Am I missing the area where I should be adding my tests to the existing {{install_puppet_from_rpm_on}} test?

petems:
@trevor-vaughan can you add collaborators as editable for this PR so I can add the tests? 👍

trevor-vaughan:
@petems It's already set. Are you pushing to my personal branch? That'll update it here.

petems:
@trevor-vaughan Cool, updated! 👍

trevor-vaughan:
@petems You rock! Thanks :beer:

kevpl:
👍

PR #1243, opened 2016-09-13
(BKR-923) Pin all gems for final 2.x gem

tvpartytonight:
The final push to the last of the beaker 2.x version will be this
commit, which pins all the gems exactly. This makes it so that there
will be no breakage from dependencies, since every library is specified
exactly.


kevpl:
Since we agreed that this was the wrong move here, I'm closing this PR. Please close the ticket for it, and we should talk about what needs to happen for the 2.y part of the 3.0 release today so it gets ticketed & done.

PR #1247, opened 2016-09-15
(BKR-854) Stop hardcoding EPEL package names

tvpartytonight:
Now that EPEL has a `release-latest` file available from source, we
should start using that and stop hard coding the value of the package
name in the beaker defaults. This change allows beaker to pick up the
newest package available, rather than having to update the hardcoded
value for each release.


tvpartytonight:

tvpartytonight:

PR #1248, opened 2016-09-19
(maint) Add upgrade docs for epel and hypervisor changes

tvpartytonight:


kevpl:
I cancelled this manually. Should be skipped since it's a doc change.

PR #1250, opened 2016-09-20
(BKR-942) let beaker use either old pkg or new pkgng tools on FreeBSD

ptomulik:
This PR enables beaker to install and query packages on new (>=9.3) and old (<9.3) versions of FreeBSD. The proposed code was used by me as [a workaround](https://github.com/ptomulik/puppet-portsng/blob/master/spec/patches/lib/beaker/host/freebsd/pkg.rb) in [ptomulik/puppet-portsng](https://github.com/ptomulik/puppet-portsng/) puppet module and worked for all FreeBSD versions between 9.0 and 11.0, where the older versions (9.0 .. 9.2) used the old pkg toolstack and newer (9.3 ... 11.0) the new pkgng suite. A similar patch was proposed for specinfra as  PR [specinfra/581](https://github.com/mizzy/specinfra/pull/581).

### DIFF
--- a/lib/beaker/host/freebsd/pkg.rb
+++ b/lib/beaker/host/freebsd/pkg.rb
@@ -1,13 +1,37 @@
 module FreeBSD::Pkg
   include Beaker::CommandFactory
 
-  def install_package(name, cmdline_args = nil, opts = {})
-    cmdline_args ||= '-y'
-    execute("pkg install #{cmdline_args} #{name}", opts) { |result| result }
+  def pkg_info_pattern(package)
+    # This seemingly restrictive pattern prevents false positives...
+    "^#{package}-[0-9][0-9a-zA-Z_\\.,]*$"
   end
 
-  def check_for_package(name, opts = {})
-    execute("pkg info #{name}", opts) { |result| result }
+  def check_pkgng_sh
+    'TMPDIR=/dev/null ASSUME_ALWAYS_YES=1 PACKAGESITE=file:///nonexist ' \
+    'pkg info -x "pkg(-devel)?\\$" > /dev/null 2>&1'
   end
 
+  def pkgng_active?(opts = {})
+    opts = {:accept_all_exit_codes => true}.merge(opts)
+    execute("/bin/sh -c '#{check_pkgng_sh}'", opts) { |r| r }.exit_code == 0
+  end
+
+  def install_package(package, cmdline_args = nil, opts = {})
+    cmd = if pkgng_active?
+            "pkg install #{cmdline_args || '-y'} #{package}"
+          else
+            "pkg_add #{cmdline_args || '-r'} #{package}"
+          end
+    execute(cmd, opts) { |result| result }
+  end
+
+  def check_for_package(package, opts = {})
+    opts = {:accept_all_exit_codes => true}.merge(opts)
+    cmd = if pkgng_active?
+            "pkg info #{package}"
+          else
+            "pkg_info -Ix '#{pkg_info_pattern(package)}'"
+          end
+    execute(cmd, opts) { |result| result }.exit_code == 0
+  end
 end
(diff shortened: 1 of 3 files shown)


lib/beaker/host/freebsd/pkg.rb
```diff
@@ -1,13 +1,28 @@
...
-    execute("pkg info #{name}", opts) { |result| result }
+  def shell_ifelse(cond, stmt_t, stmt_f)
+    "if #{cond}; then #{stmt_t}; else #{stmt_f}; fi"
   end
 
+  def install_package(package, cmdline_args = nil, opts = {})
+    cmd = shell_ifelse shell_check_pkgng,
+                       "pkg install #{cmdline_args || '-y'} #{package}",
+                       "pkg_add #{cmdline_args || '-r'} #{package}"
+    execute("/bin/sh -c '#{cmd}'", opts) { |result| result }
```
ptomulik: Well, by default FreeBSD uses `csh` shell, which has different syntax than `sh`. This line "makes the code to work" but it eventually changes shell. If there are some environment variables set only for the `csh` shell and are not set for `sh`, then some problems may appear here.

lib/beaker/host/freebsd/pkg.rb
```diff
@@ -1,13 +1,28 @@
...
+  end
+
+  def check_for_package(package, opts = {})
+    cmd = shell_ifelse shell_check_pkgng,
+                       "pkg info #{package}",
+                       "pkg_info -Iqx #{package} 2> /dev/null || true"
```
ptomulik: ``pkg_info -Iqx #{package} 2> /dev/null || true`` is just copied from the original code, but I wonder, how it was supposed to test for anything? It always returns exit code 0.

kevpl:
@ptomulik thanks for submitting these fixes!

@petems would you mind taking a look at these FreeBSD changes?

lib/beaker/host/freebsd/pkg.rb
ptomulik: Oh my, I don't actually remember where it comes from, thought I've found it somewhere in beakers code. The problem is, that in case ``#{package}`` is not installed, ``pkg_info`` exits with exit code ``1`` and beaker exits with failure. Looks like I have to elaborate this a little bit.

kevpl:
@petems do you have any time to look at these changes specific to FreeBSD?

petems:
@kevpl LGTM! 👍

lib/beaker/host_prebuilt_steps.rb
```diff
@@ -13,7 +13,7 @@ module HostPrebuiltSteps
     SLEEPWAIT = 5
     TRIES = 5
     UNIX_PACKAGES = ['curl', 'ntpdate']
-    FREEBSD_PACKAGES = ['curl', 'perl5']
+    FREEBSD_PACKAGES = ['curl', 'perl5|perl']
```
ptomulik: This will need attention if someone is going to implement e.g. docker hypervisor for freebsd, as these constants are used in ``beaker/hypervisor/docker.rb``. Currently they are used simply as:
```ruby
dockerfile += <<-EOF
...
apt-get install -y openssh-server openssh-client #{Beaker::HostPrebuiltSteps::CUMULUS_PACKAGES.join(' ')
..
EOF
```
etc.. Such a code is not able to handle our ``FREEBSD_PACKAGES`` which contain "alternatives" in form ``pkg1|pkg2``.

petems: I think that's ok, running FreeBSD in a docker container is a pretty uncommon edgecase

ptomulik:
Well then, merge, maybe?

kevpl:

kevpl:
well, I merged that without realizing it hadn't passed through our automated testing. I've kicked that testing off now. If it fails, I'll revert the change. But hopefully that won't be the case, and it'll confirm that it's fine to leave this change in....

kevpl:
Looks like Jenkins won't pick up changes for testing that have already been merged.

That's fine, our master status job's [last night run (internal link only)](https://jenkins-beaker.delivery.puppetlabs.net/job/qe_beaker_btc-intn-master/899/) was green including this change, so we should be good.

ptomulik:
Thank you vm.

PR #1251, opened 2016-09-20
(maint) Improve beaker vmpooler error response

briancain:
Prior this commit, if a user had an old token and attempted to run
beaker to request vms, it would claim that the requested hosts were not
available in the pooler when in reality their token was invalid and not
authorized to make the request. This commit updates that to check the
status code of the request and print the appropriate error message.


kevpl:

PR #1253, opened 2016-09-21
(maint) add name and roles tags to vmpooler vms

johnmccabe:
Before this commit it was not possible to tell via [vmfloaty](https://github.com/briancain/vmfloaty) or [vmpooler-bitbar](https://github.com/johnmccabe/vmpooler-bitbar/) what a particular vmpooler vm was being used for when created by beaker.

This commit adds tags with the name and roles for vmpooler vms, for example via vmpooler-bitbar:

screen shot 2016-09-22 at 7 19 27 pm

### DIFF
--- a/lib/beaker/hypervisor/vmpooler.rb
+++ b/lib/beaker/hypervisor/vmpooler.rb
@@ -81,13 +81,14 @@ module Beaker
     # @return [Hash] Tag hash
     def add_tags(host)
       host[:host_tags].merge(
-          {
-              'beaker_version'    => Beaker::Version::STRING,
-              'jenkins_build_url' => @options[:jenkins_build_url],
-              'department'        => @options[:department],
-              'project'           => @options[:project],
-              'created_by'        => @options[:created_by]
-          })
+        'beaker_version'    => Beaker::Version::STRING,
+        'jenkins_build_url' => @options[:jenkins_build_url],
+        'department'        => @options[:department],
+        'project'           => @options[:project],
+        'created_by'        => @options[:created_by],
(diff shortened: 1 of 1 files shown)


johnmccabe:
fyi @briancain

pinkypie:
please retest this

briancain:
Oh nice 👍 , tagging with name and role is a great idea.

johnmccabe:
@briancain will be updating vmpooler-bitbar to detect nodes with the dashboard role and offer a link directly to it from the menu, also gonna replace the template in brackets with the name on the main dropdown.

tvpartytonight:
@johnmccabe thank for this PR! I've tested this and found an edge case where if you define a host without any roles, there will be no `roles` key defined for that vm in the vmpooler. This seems more like an issue with the vmpooler and not beaker, so I don't think there are any changes needed for your PR in that respect. Just wanted to give you a headsup that the `roles` key is not guaranteed to exist for the vm after tagging.

Given that we are now on beaker 3.x on master, would you mind rebasing this PR with master? That will allow the testing to go through with beaker 3 and give us better confidence that the changes will work. Thanks!

johnmccabe:
hi @tvpartytonight yeah the absence of the roles key is fine (the johnmccabe/vmpooler-bitbar#1 pr expects that behaviour for example), I'll retarget this later tonight

johnmccabe:
@tvpartytonight thats rebased, I don't think the CI failure is related.

ferglor:
@puppetlabs-jenkins retest this please jabroni

PR #1254, opened 2016-09-23
(maint) add doc for BEAKER_VERSION env vars

tvpartytonight:
[skip ci]

### DIFF
--- a/docs/how_to/test_arbitrary_beaker_versions.md
+++ b/docs/how_to/test_arbitrary_beaker_versions.md
@@ -1,0 +1,46 @@
+# Test arbitrary beaker versions without modifying test code
+
+In order to adjust the beaker version used without commiting a change to a Gemfile,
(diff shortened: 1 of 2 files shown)


kevpl:
these two files look the same. I would imagine that if it needs mentioning that new versions can be tested this way, you'd just do it in the arbitrary version doc. Also, there's no mention of this in the upgrade doc, suggesting it be used to test the new version before upgrading.

Also, the doc name doesn't fit in with docs in that folder, 'testing' should be 'test' so that it reads 'How to test arbitrary beaker versions'. Not all of the docs in the folder conform to this, but most do, and I think it'd be better if we move to that standard for them.

Other than that, 👍!

tvpartytonight:
@kevpl whoops, I missed staging the deletion of the old file name. Comments addressed, should be good to go!

PR #1257, opened 2016-09-28
Require beaker-pe if available

tvpartytonight:
This change pulls in beaker-pe if it is available in the ruby
$LOAD_PATH. This makes it so that project repos only have to specify
beaker-pe in their Gemfile, not require beaker-pe itself in their
test setup.

### DIFF
--- a/acceptance/pre_suite/pe/install.rb
+++ b/acceptance/pre_suite/pe/install.rb
@@ -1,1 +1,3 @@
+# require beaker-pe to load in the additional DSL methods
+require 'beaker-pe'
 install_pe
--- a/lib/beaker.rb
+++ b/lib/beaker.rb
@@ -46,4 +46,13 @@ module Beaker
     # do nothing
   end
 
+  # If beaker-pe is available, pull it in. The gem beaker-pe will need to be
+  # specified in the project Gemfile independent of beaker itself. If not available,
+  # catch LoadError and continue.
+  begin
+    require 'beaker-pe'
+  rescue LoadError
+    # do nothing
+  end
+
 end


mwbutcher:
:+1:  
This would solve for several failures that Integration is seeing in the 3.8.x pipeline.  This will have to be resolved, and this seems like the right way

kevpl:
@puppetlabs/beaker @rick @tvpartytonight and I were discussing the merits of potentially moving to this approach, rather than needing them to include the gem in their dependencies and load it themselves.

One of my worries is that we then have two separate instructions for how to setup a beaker-library. That's not really more of an objection than a hiccup though, we can easily overcome that. What do you think?

kevpl:
@mwbutcher you can resolve your integration failures right now by using the new import mechanism and including beaker-pe. 

Here's an [example branch](https://github.com/puppetlabs/pe_acceptance_tests/compare/2016.5.x...kevpl:bkr941_pinned_test_fixes?expand=1) I made that shows how you can do that for 2016.5. The change shouldn't be hard to port back to older versions of testing if they have acceptance helpers. The [upgrade doc](https://github.com/puppetlabs/beaker/blob/master/docs/how_to/upgrade_from_2_to_3.md) should provide sufficient instructions on how to upgrade to beaker 3.0. If it doesn't, let us know and we'll fix that.

I'd rather make sure we get the time to consider this PR, rather than be unnecessarily pressured into merging it.

tvpartytonight:

tvpartytonight:
@kevpl just to clarify, this change still requires an update to a project's Gemfile, but does not require the project `require 'beaker-pe` in its helper(s).

With this change, we are still forcing users to add the library via their Gemfile, but not requiring that they update their actual code to require it. I was motivated to submit this change because it it is relatively trivial to modify a Gemfile for all projects, but figuring out a way to `require beaker-pe` in all projects is something we can't easily/universally do. This simplifies the testing effort in all our pipelines significantly.

However, I am open to the idea that once this initial refactor is done by users, it will be better in the long term for beaker to have a cleaner separation of concerns in its own independent gems...

mwbutcher:
@kevpl including beaker-pe explicitly works great in 2016.4 and newer, but for 3.8.x we don't have a good place to require it centrally.  The good news is that once we shut down 3.8.x it won't matter, so this fix could be temporary.

tvpartytonight:
@mwbutcher I hacked it into the [install file](https://github.com/puppetlabs/pe_acceptance_tests/blob/3.8.x/setup/install.rb) in beaker's ci for testing 3.8.x installation. Not the best place for it, but, it works.

mwbutcher:
@highb are you okay with that solution?

highb:
👍  yeah, that's an OK hack for now, I guess?

nwolfe:
We could help ourselves remember to get rid of this hack with a JIRA ticket mentioned in the code.

kevpl:
I'm 👍 on getting this in, it makes sense this will pick up a lot for usability, even if it's only a middle step. We should do the work to get testing green here if possible on our pipelines before merging, as it'll be necessary before release if we're to have confidence anyways

tvpartytonight:

tvpartytonight:

tvpartytonight:

PR #1259, opened 2016-09-30
(BKR-949) Fix Keypair Generation on OpenStack

spjmurray:
Fixes a regression in 2.51.0 whereby the hostname was updated to be based upon the
server IP, and get a domain.  Random keypair generation is based upon this hostname
however OpenStack deems the period character unsafe, therefore this functionality
got broken.  This commit attaches both the hostname and fqdn to the host object.  To
maintain backwards compatibility all references to vmhostname now refer to vmfqdn.
Key generation now uses the safe vmhostname variable, and clean up uses vm.key_name
cached in the Fog Server object.

### DIFF
--- a/lib/beaker/hypervisor/openstack.rb
+++ b/lib/beaker/hypervisor/openstack.rb
@@ -197,18 +197,19 @@ module Beaker
 
       @hosts.each do |host|
         ip = get_ip
-        hostname = ip.ip.gsub '.','-'
-        host[:vmhostname] = hostname+'.rfc1918.puppetlabs.net'
-        @logger.debug "Provisioning #{host.name} (#{host[:vmhostname]})"
+        host[:vmhostname] = ip.ip.gsub '.','-'
+        host[:vmfqdn] = host[:vmhostname] + '.rfc1918.puppetlabs.net'
+        host[:keyname] = key_name(host)
+        @logger.debug "Provisioning #{host.name} (#{host[:vmfqdn]})"
         options = {
           :flavor_ref => flavor(host[:flavor]).id,
           :image_ref  => image(host[:image]).id,
           :nics       => [ {'net_id' => network(@options[:openstack_network]).id } ],
-          :name       => host[:vmhostname],
-          :hostname   => host[:vmhostname],
+          :name       => host[:vmfqdn],
+          :hostname   => host[:vmfqdn],
           :user_data  => host[:user_data] || "#cloud-config\nmanage_etc_hosts: true\n",
+          :key_name   => host[:keyname],
         }
-        options[:key_name] = key_name(host)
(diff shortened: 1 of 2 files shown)


spjmurray:
Correct me if I'm wrong but the `:hostname` option on line 208 got added in 2.51.0 also, but it doesn't appear here https://github.com/fog/fog-openstack/blob/master/lib/fog/compute/openstack/models/server.rb can it be deleted?

lib/beaker/hypervisor/openstack.rb
```diff
@@ -197,15 +197,15 @@ def provision
 
       @hosts.each do |host|
         ip = get_ip
-        hostname = ip.ip.gsub '.','-'
-        host[:vmhostname] = hostname+'.rfc1918.puppetlabs.net'
-        @logger.debug "Provisioning #{host.name} (#{host[:vmhostname]})"
+        host[:vmhostname] = ip.ip.gsub! '.','-'
```
kevpl: @spjmurray this was changed to the non-bang version recently in [this changeset](https://github.com/puppetlabs/beaker/commit/5f50452a3b5f70e15a423449684d2033a2488805). If you check that code out, it looks like it was setting it back to '.'s rather than '-'s just before the `host[:ip]` got set on line 237, so that code didn't change the `ip.ip` value.

It looks like you'd rather keep that value as '-'s here, and seem fine setting the `host[:ip]` to use those dashes. Was something broken using the '.'s? Or am I misunderstanding this code?

spjmurray: No Kev, you are right and a good reviewer :)  I ported this from a 2.51.0 tree and I missed this change.  Branch updated to take it into account

kevpl:
I'm not sure why it matters that a particular host property in our openstack hypervisor isn't tracked in fog-openstack. Beaker's hypervisors can definitely contain more information than the actual hypervisor needs in order to help beaker accomplish its tasks.

@lucywyman, you added this code as a part of [this change](https://github.com/puppetlabs/beaker/commit/fff68bbbced9e3cd9fa468dce3456dca33295bec). What do you think of the above comment?

kevpl:
ping @spjmurray, @lucywyman

lucywyman:
Apologies, not sure how I missed the first ping.  :eyes:

lucywyman:
+1 from me.

kevpl:
looks good to me too @spjmurray. If you would add a test to make sure we don't break this functionality by poisoning with dots going forward, then we will be ready to get this tested and merged.

spjmurray:
@kevpl done I think/hope/pray, cast your eyes over it at your leisure

kevpl:

PR #1260, opened 2016-09-30
(BKR-948) require ruby >= 2.2.5

kevpl:
In the release of beaker 3.0, we said that we gave up supporting
ruby 1.9, but we didn't actually stop anyone from using it. This
change will prevent anyone from using beaker 3.0 with any ruby
< 2.2.5

### DIFF
--- a/beaker.gemspec
+++ b/beaker.gemspec
@@ -17,6 +17,8 @@ Gem::Specification.new do |s|
   s.executables   = `git ls-files -- bin/*`.split("\n").map{ |f| File.basename(f) }
   s.require_paths = ["lib"]
 
+  s.required_ruby_version = Gem::Requirement.new('>= 2.2.5')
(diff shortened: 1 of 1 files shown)


kevpl:

cdenneen:
@kevpl @tvpartytonight Can I ask that we change this to >= 2.1.8?
Currently puppet-agent still ships with 2.1.x and and therefore PE only ships with this.
I would believe running tests on a later version of ruby could cause some actual issues with the runtime version doesn't match (if a bug is fixed in later than version shipped).
I think aligning this and beaker-rspec with minimum version that is oldest supported PE would be good practice here.
Any objections modifying this require?

cdenneen:
Added PR if this is acceptable: https://github.com/puppetlabs/beaker/pull/1327

PR #1262, opened 2016-10-05
(BKR-928) add hocon DSL helpers

kevpl:
In Everett, PE will be moving away from using the node classifier
to classify nodes, and instead relying on pe.conf HOCON files. This
change introduces basic CRUD style interactions with HOCON files
for beaker testing.

### DIFF
--- a/acceptance/tests/base/dsl/helpers/hocon_helpers_test.rb
+++ b/acceptance/tests/base/dsl/helpers/hocon_helpers_test.rb
@@ -1,0 +1,94 @@
+require 'hocon/config_value_factory'
+
+test_name 'Hocon Helpers Test' do
+
+  hocon_filename = 'hocon.conf'
+  step 'setup : create hocon file to play with' do
+    hocon_content = <<-END
+      {
+        setting1 : "value1",
+        setting2 : 2,
+        setting3 : False
+      }
+    END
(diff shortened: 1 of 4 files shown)


kevpl:
failures appear unrelated (modules testing has other issues right now, it's being worked on now).

@ericwilliamson would you take a look at this and let me know if there's something else you'd need/want from this?

ericwilliamson:
@kevpl Thanks for the help.


Any concern with this duplicating https://github.com/puppetlabs/beaker/blob/7b594444b0df51694e4ce66dd4c8e343678f1955/lib/beaker/dsl/helpers/tk_helpers.rb#L28

kevpl:
@ericwilliamson sure, I can see some concern there. However, I think Beaker's pattern of just hash smashing, while useful, isn't really what we want from beaker going forward. Abstracting away boilerplate things like copying data or files to and from SUTs seem to be Beaker's strong suit, so using that boilerplate code, but making the method more open & self-service by exposing a block to yield to seems like a better option going forward.

kevpl:
@ericwilliamson let me know if you're OK with my response or if you have a response to it.

If you're good, then I think we're good to merge on this.

ericwilliamson:
👍  looks good to me!

ferglor:
👍

PR #1263, opened 2016-10-05
(BKR-952) Add PuppetVersion

highb:
The `version_is_less` method is widely used for comparing PE and Puppet
version across all of the acceptance tests. But only being able to
perform a less than comparison is limiting and hard to read since
you must check `!version_is_less` to determine if versions are
greater than.

This commit adds a `PuppetVersion` class that extends `Semantic::Version`
which should allow any standard ruby comparison operators to be used with
it.

### DIFF
--- a/spec/beaker/shared/semvar_spec.rb
+++ b/spec/beaker/shared/semvar_spec.rb
@@ -43,6 +43,84 @@ module Beaker
         end
       end
 
+      describe 'puppet_version_comparison' do
+        it 'reports 2015.3.0-rc0-8-gf80879a is less than 2016' do
+          expect( subject.puppet_version_comparison( '2015.3.0-rc0-8-gf80879a', '<', '2016' ) ).to be === true
+        end
+
+        it 'reports 2016 is greater than 2015.3.0-rc0-8-gf80879a' do
+          expect( subject.puppet_version_comparison( '2016', '>', '2015.3.0-rc0-8-gf80879a' ) ).to be === true
+        end
+
+        it 'reports 2015.3.0-rc0-8-gf80879a is equal to 2015.3.0-rc0-8-gf80879a' do
+          expect( subject.puppet_version_comparison( '2015.3.0-rc0-8-gf80879a', '==', '2015.3.0-rc0-8-gf80879a' ) ).to be === true
+        end
+
+        it 'reports 2016 is greater than 2015.3.0-rc0-8-gf80879a' do
+          expect( subject.puppet_version_comparison( '2015.3.0-rc1-4-gabcdefg', '>', '2015.3.0-rc0-8-gf80879a' ) ).to be === true
+        end
+
+        it 'reports 2016 is greater than 2015.3.0-rc0-8-gf80879a' do
+          expect( subject.puppet_version_comparison( '2015.2.0-rc1-9-gabcdefg', '>', '2015.3.0-rc0-8-gf80879a' ) ).to be === false
+        end
+
+        it 'reports 2015.3.0-rc0-8-gf80879a is not less than 2015.3.0' do
+          expect( subject.puppet_version_comparison( '2015.3.0-rc0-8-gf80879a', '<', '2015.3.0' ) ).to be === false
+        end
+
+        it 'reports 2015.3.0-rc0-8-gf80879a is not less than 3.0.0' do
+          expect( subject.puppet_version_comparison( '2015.3.0-rc0-8-gf80879a', '<', '3.0.0' ) ).to be === false
+        end
+
+        it 'reports 3.0.0-160-gac44cfb is not less than 3.0.0' do
+          expect( subject.puppet_version_comparison( '3.0.0-160-gac44cfb', '<', '3.0.0' ) ).to be === false
+        end
+
+        it 'reports 3.0.0-160-gac44cfb is not less than 2.8.2' do
+          expect( subject.puppet_version_comparison( '3.0.0-160-gac44cfb', '<', '2.8.2' ) ).to be === false
+        end
+
+        it 'reports 3.0.0 is less than 3.0.0-160-gac44cfb' do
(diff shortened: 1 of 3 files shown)


pinkypie:
@highb would you be able to include a small docs/how_to on recommended way to use this method to compare puppet versions, maybe with a link in there to how puppet versioning works in general?

highb:
@pinkypie Sure, I can add that to the comment block.

highb:
@pinkypie I added more information about the format of PE versions (as far as I know) and split out the version checking into a separate method for re-usability and readibility.

spec/beaker/shared/semvar_spec.rb
```diff
@@ -43,6 +43,180 @@ module Shared
...
+        end
+
+        it 'rejects a non version string' do
+          expect { subject.puppet_version_split('banana') }.to raise_error(/Unknown version format/)
+        end
+      end
+
+      describe 'puppet_version_comparison' do
+        it 'reports 2015.3.0-rc0-8-gf80879a is less than 2016' do
+          expect( subject.puppet_version_comparison( '2015.3.0-rc0-8-gf80879a', '<', '2016' ) ).to be === true
```
kevpl: @highb I'm curious about this calling style. Why don't we have our own `PuppetVersion` object that inherits from `Semantic::Version` that implements the Ruby version of what I think of as Java's Comparable interface, so that we can just use Ruby's comparison operators rather than passing in the comparison operator as another string parameter?

spec/beaker/shared/semvar_spec.rb
```diff
@@ -43,6 +43,180 @@ module Shared
...
+
+        it 'reports 2016 is greater than 2015.3.0-rc0-8-gf80879a' do
+          expect( subject.puppet_version_comparison( '2015.2.0-rc1-9-gabcdef1', '>', '2015.3.0-rc0-8-gf80879a' ) ).to be === false
+        end
+
+        it 'reports 2015.3.0-rc0-8-gf80879a is less than 2015.3.0' do
+          expect( subject.puppet_version_comparison( '2015.3.0-rc0-8-gf80879a', '<', '2015.3.0' ) ).to be === true
+        end
+
+        it 'reports 2015.3.0-rc0-8-gf80879a is not less than 3.0.0' do
```
kevpl: 💯 this is the test case I want!

kevpl:
@highb I'm assuming this is PR'd against this repo rather than beaker-pe because it works for Puppet as well as PE versions. Is this correct?

spec/beaker/shared/semvar_spec.rb
highb: @kevpl Yeah, that would probably work as well and be cleaner. This is just a little side hack because I was getting really annoyed at `version_is_less` being so inflexible.

highb:
@kevpl Yeah on the PE vs puppet issue, I think the internal puppet builds are just plain semver so they technically would work with this.

spec/beaker/shared/semvar_spec.rb
highb: Heh yeah that was another thing that was bugging me.

kevpl:
@highb it looks like our [automated testing has failed](https://jenkins-beaker.delivery.puppetlabs.net/job/qe_beaker_unit-rb_rspec/3703/ruby=2.2.5/console) on the way the semantic gem is included.

highb:
@kevpl D'oh yeah I noticed that on the last run and tried fixing it to look more like the style of the other requirements in the Gemspec but it still doesn't seem to like it.

highb:
@kevpl I see you made a ticket for this, I'll update the commits with it when I refactor this into a version object.

highb:
@kevpl Also, is there a different namespace I should put it under?

highb:
Woah wait, I made that ticket? Am I having Jira amnesia? wtf?

kevpl:
😆 I was definitely confused by your last comment about me making a ticket. I thought maybe you were referring to another one 😄

highb:
I blame too much task switching. >_<

highb:
Hm, do we need to push the semantic version to our internal rubygem mirror?
```
Fetching source index from http://rubygems.delivery.puppetlabs.net/
Resolving dependencies....
Bundler could not find compatible versions for gem "semantic":
  In Gemfile:
    beaker was resolved to 3.1.0, which depends on
      semantic (~> 1.4)

Could not find gem 'semantic (~> 1.4)', which is required by gem 'beaker', in
any of the sources.
```

kevpl:
pushed up semantic 1.4.1, let's try this again!

highb:

highb:

highb:
@kevpl Updated/refactored to extend `Semantic::Version`. Jenkins is doing strange things, though.

highb:

kevpl:

kevpl:
@highb It looks like we're having a semantic version conflict w/puppet in [this line](https://github.com/puppetlabs/puppet/blob/4.7.0/lib/puppet/pops/pcore.rb#L14). I'm not sure if this is exactly why, but it does look like the conflict could be coming from the fact that puppet vendors the semantic gem ([folder](https://github.com/puppetlabs/puppet/tree/4.7.0/lib/puppet/vendor)).

highb:
@kevpl Gr, well that is frustrating. I guess I could just call the Semantic being used in Puppet? Seems kind of funky though.

kevpl:
@highb looking back at this, I'm a little weirded out by the idea that Puppet's semantic gem should have conflicts with beaker's considering that beaker's ruby & requirements should just live on the coordinator & Puppet's should be isolated to the SUTs themselves. I'm going to re-kick this and look into it further. 

kevpl:
Unfortunately I wasn't able to look at this last week. It might have to wait a while before I can come around to investigating it further.

highb:
@kevpl No worries, thanks for keeping me updated with the status.

tvpartytonight:

kevpl:
@highb I'm going to close this PR, since we haven't made any progress with it. I like the ideas here, but haven't had time to pursue them further.

PR #1265, opened 2016-10-11
(BKR-956) implemented a beaker init subcommand

ferglor:
Currently writing spec tests for this change, but wanted to get a discussion going on how things are headed

### DIFF
--- a/bin/beaker
+++ b/bin/beaker
@@ -4,6 +4,6 @@ require 'rubygems' unless defined?(Gem)
 require 'beaker'
 
 Beaker::CLI.new.execute!
-
 puts "Beaker completed successfully, thanks."
+
 exit 0
--- a/lib/beaker/shared/subcommands_util.rb
+++ b/lib/beaker/shared/subcommands_util.rb
@@ -1,0 +1,37 @@
+module Beaker
+  module Shared
+    # Methods used in execution of Subcommands
+    # - should we write the config?
+    # - reset ARGV 
+    # - execute Beaker
+    module SubcommandsUtil
+      CONFIG_PATH = ".beaker/config"
+    
+      @@write_config = false
+
+      def self.write_config=( val )
+        @@write_config = val
+      end
+      
+      def self.write_config?
+        @@write_config
+      end
+      
+      # Reset ARGV to contain the arguments determined by a specific subcommand
+      # @param [Array<String>] args the arguments determined by a specific subcommand
+      def reset_argv(args)
+        ARGV.clear
+        args.each do |arg|
+          ARGV << arg
+        end
+      end
+
+      # Update ARGV and call Beaker
+      # @param [Array<String>] args the arguments determined by a specific subcommand
+      def execute_beaker(*args)
+        reset_argv(args)
+        Beaker::CLI.new.execute! 
+      end
+    end
+  end
+end
--- a/lib/beaker/cli.rb
+++ b/lib/beaker/cli.rb
@@ -31,9 +31,15 @@ module Beaker
         @execute = false
         return
       end
-      @logger.info("Beaker!")
-      @logger.info(beaker_version_string)
-      @logger.info(@options.dump)
+
+      if Beaker::Shared::SubcommandsUtil::write_config?
+        @options.dump_to_file(Beaker::Shared::SubcommandsUtil::CONFIG_PATH)
+      else
+        @logger.info("Beaker!")
+        @logger.info(beaker_version_string)
+        @logger.info(@options.dump)
+      end
+
       if @options[:parse_only]
         @execute = false
         return
--- a/lib/beaker/options/options_hash.rb
+++ b/lib/beaker/options/options_hash.rb
@@ -38,6 +38,14 @@ module Beaker
         end
       end
 
+      def dump_to_file(output_file)
+        dirname = File.dirname(output_file)
+        unless File.directory?(dirname)
+          FileUtils.mkdir_p(dirname)
+        end
+        File.open(output_file, 'w+') { |f| f.write(dump)  }
+      end
+
     end
   end
 end
--- a/lib/beaker/shared.rb
+++ b/lib/beaker/shared.rb
@@ -1,4 +1,4 @@
-[ 'repetition', 'error_handler', 'host_manager', 'timed', 'semvar', 'options_resolver' ].each do |lib|
+[ 'repetition', 'error_handler', 'host_manager', 'timed', 'semvar', 'options_resolver', 'subcommands_util' ].each do |lib|
   require "beaker/shared/#{lib}"
 end
 module Beaker
@@ -9,6 +9,7 @@ module Beaker
     include Beaker::Shared::Timed
     include Beaker::Shared::Semvar
     include Beaker::Shared::OptionsResolver
+    include Beaker::Shared::SubcommandsUtil
   end
 end
 include Beaker::Shared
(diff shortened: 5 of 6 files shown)


bin/beaker
```diff
@@ -3,7 +3,11 @@
 require 'rubygems' unless defined?(Gem)
 require 'beaker'
 
-Beaker::CLI.new.execute!
+if (Beaker::Subcommands.instance_methods(false) << :help).include? ARGV[0].to_sym
```
ferglor: initially, this line looked like:

if ["help", "init"].include? ARGV[0]

which is a lot easier to read, but the current implementation means that we don't need to update this line every time we add a new subcommand

lib/beaker/logger.rb
```diff
@@ -112,19 +112,19 @@ def initialize(*args)
       end
 
       dests = args
-      dests << STDOUT unless options[:quiet]
+      dests << $stdout unless options[:quiet]
```
ferglor: switching from using STDOUT to using $stdout allows us to specify where logger's stdout should be written to, prior to execution of beaker

ferglor:
Currently writing unit tests for this change, but wanted to get a discussion going on how things are headed

lib/beaker/subcommands.rb
```diff
@@ -0,0 +1,22 @@
...
+module Beaker
+  class Subcommands < Thor
+    include Beaker::Shared::SubcommandsUtil
+
+    desc "init", "Initialises the beaker test environment"
+    long_desc <<-LONGDESC
+      Initialises a beaker environment using a combination of environment variables and options
+    LONGDESC
+    option :host, :default => "hosts.yaml", :type => :string, :desc => "The host configuration to test against"
+    option :test, :default => "acceptance/tests/base/", :type => :string, :desc => "The test suite to run"
```
ferglor: have we decided upon a set of flags to pass to beaker for the init phase? i settled on passing --host and --test for now, but adding more flags should be very straightforward

lib/beaker/subcommands.rb
```diff
@@ -0,0 +1,22 @@
...
+require "fileutils"
+require_relative "shared/subcommands_util"
+
+module Beaker
+  class Subcommands < Thor
+    include Beaker::Shared::SubcommandsUtil
+
+    desc "init", "Initialises the beaker test environment"
+    long_desc <<-LONGDESC
+      Initialises a beaker environment using a combination of environment variables and options
```
ferglor: should we consult someone from ux on the wording of the short and long descriptions here? i just winged it here to have something to print out

bin/beaker
ferglor: initially, this line looked like:

if ["help", "init"].include? ARGV[0]

which is a lot easier to read, but the current implementation means that we don't need to update this line every time we add a new subcommand

thor provides a "help" subcommand out of the box, so we need to append :help to the array of Subcommands instance method names, which so far only includes the "init" subcommand

lib/beaker/subcommands.rb
```diff
@@ -0,0 +1,23 @@
+require "thor"
+require "fileutils"
+require_relative "shared/subcommands_util"
+
+module Beaker
+  class Subcommands < Thor
+    include Beaker::Shared::SubcommandsUtil
```
ferglor: i'm not too sure if i need this include and the require_relative, i was basically trying to mimic static method imports in java, so subcommands.rb only contains subcommand methods

kevpl:
sadly I didn't get to taking a look at this PR today @ferglor. Please checkout your other assigned tickets in the meantime, and I should be able to look at this tomorrow

lib/beaker/shared/subcommands_util.rb
```diff
@@ -0,0 +1,60 @@
...
+        yield
+        $stdout.string
+      ensure
+        $stdout = old_stdout
+      end
+
+      # Extract a JSON string from a block of text
+      # @param [String] text the block of text containing the JSON
+      # @return the JSON string
+      def extract_json(text)
```
ferglor: instead of grepping for json in the beaker output, we could update: https://github.com/puppetlabs/stringify-hash/blob/master/lib/stringify-hash.rb to have a method for dumping to a file? i'm guessing we'd also need to add something like a --write-config flag to beaker?

ferglor:
updated the review - we now pass a flag to beaker to specify that we want to write to the config file. this is slightly more invasive, but it means we can dump the config directly rather than grepping beaker's output. it also means i can revert the changes i had made to the logger, and lets me remove a number of the util methods i had created.

beaker.gemspec
```diff
@@ -53,6 +53,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'stringify-hash', '~> 0.0'
   s.add_runtime_dependency 'beaker-hiera', '~> 0.0'
   s.add_runtime_dependency 'beaker-hostgenerator'
+  s.add_runtime_dependency 'thor'
```
kevpl: @ferglor, have you or @tvpartytonight done any experiments or reading on any other options for implementing subcommands? One potential contender @nwolfe brought up was [commander](https://github.com/commander-rb/commander), have we looked at this?

@rick have any advice for implementing subcommands in the ruby world?

kevpl:
What does it look like when you run `beaker init`? What do you see in one of these files?

Would you mind linking to a gist of a local run, so we can get an idea for what comes out of this process as it currently is?

beaker.gemspec
rick: The things I'm most familiar with are minimal (`optparse` and its variants); I'm passingly aware of `thor`, `commander`, `GLI`, `trollop` and `slop`, but don't have strong opinions on them.

bin/beaker
rick: Can we give this an intention-revealing name, wrap it in a method and use that instead?

ferglor: @rick absolutely!

beaker.gemspec
pinkypie: fwiw Test Kitchen uses thor

bin/beaker
```diff
@@ -3,7 +3,11 @@
...
+  Beaker::Subcommands.start(ARGV)
+else
+  Beaker::CLI.new.execute!
+  puts "Beaker completed successfully, thanks."
```
ferglor: to me it looks a little noisy. so if i run beaker init, the output would be like:

Beaker configuration written to .beaker/config
Beaker completed successfully, thanks.

like the beaker init task is supposed to do one small task so to me it's like saying

HEY I DID THAT ONE TASK YOU ASKED OF ME
I'VE FINISHED DOING WHAT YOU ASKED 

it's like when a child tells you something they've done, and you don't quite acknowledge them, so they tell you again, but with different wording

to me, the "Beaker completed successfully, thanks." makes sense when we've printed a bunch of text to the screen, i.e. during a test run, and you can't be bothered having to read through the block so you do tl;dr and skip to the end to get the gist of what's happened.

ferglor:
@kevpl @rick beaker init gist: https://gist.github.com/ferglor/c01a9ed6925e8bcd331bfd85284a4c7a

kevpl:
it looks like this work is more than what is specified from the ticket itself, or the ticket isn't clear enough. I've commented in the JIRA ticket with more info on this.

kevpl:
hm, weird that the test reports success when I click to it in the Jenkins interface, but that the github integration says it failed. I'll treat this like a one-off transient for now unless we see it repeat.

Since that's the case, this looks good! Merging now. 🚀

PR #1266, opened 2016-10-17
(MAINT) Allow no hosts option to be passed

kevpl:
In previous versions (such as 2.14.1, which was previously
used to run our ec2 zombie killer), beaker could be run
without a hosts file argument. This would allow beaker test
logic to be run without needing to stand up SUTs. During the
time where beaker-hostgenerator support was added, this
ability was removed. This change allows us to run without
the hosts file argument, as was previously done.

### DIFF
--- a/lib/beaker/options/parser.rb
+++ b/lib/beaker/options/parser.rb
@@ -224,7 +224,7 @@ module Beaker
       # @raise [ArgumentError] if a hosts file is generated, but it can't
       #   be read by the HostsFileParser
       def parse_hosts_options
-        if File.exists?(@options[:hosts_file])
+        if @options[:hosts_file].nil? || File.exists?(@options[:hosts_file])
           #read the hosts file that contains the node configuration and hypervisor info
           return Beaker::Options::HostsFileParser.parse_hosts_file(@options[:hosts_file])
         end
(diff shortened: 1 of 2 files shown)


lib/beaker/options/parser.rb
```diff
@@ -224,7 +224,7 @@ def parse_args(args = ARGV)
       # @raise [ArgumentError] if a hosts file is generated, but it can't
       #   be read by the HostsFileParser
       def parse_hosts_options
-        if File.exists?(@options[:hosts_file])
+        if @options[:hosts_file].nil? || File.exists?(@options[:hosts_file])
```
nwolfe: Should this be negated? `!@options[:hosts_file].nil?`

Otherwise it'll enter the `if` statement and `nil` will be passed to `HostsFileParser.parse_hosts_file`, which I'm assuming is bad.

spec/beaker/options/parser_spec.rb
```diff
@@ -246,7 +246,9 @@ def mock_out_parsing
           end
 
           it 'calls beaker-hostgenerator to get hosts information' do
-            parser.instance_variable_set( :@options, {} )
+            parser.instance_variable_set( :@options, {
+              :hosts_file => 'notafile.yml'
```
nwolfe: These changes look like a hosts file *must* be provided?

nwolfe:
Should we update the CLI help text for `--hosts` so it mentions that this value can be:

* A filepath
* A BHG string
* Omitted

ferglor:
👍

nwolfe:
When will we update the CLI docs for this field?  Or do you typically file that sort of work as separate tickets?

PR #1268, opened 2016-10-25
(BKR-965) Re-enable the prior way of setting hostnames but keep the new default

doug-rosser:
With Tony's [check-in](https://github.com/puppetlabs/beaker/commit/a5bbdab25281745e7ee053ffd8bc01dae1180f69) the default hostnames of ec2 VMs went from whatever was configured in the beaker hosts file, to the DNS-name given by the Amazon ec2 service. Unfortunately, this broke the pe_acceptance_tests tooling on ec2, which is used for release scale testing.

This check-in makes the old behavior available again, but only if you set :use_beaker_hostnames: true in the beaker hosts file. In other words, this should NOT break any testing that required Tony's original check-in.

### DIFF
--- a/lib/beaker/hypervisor/aws_sdk.rb
+++ b/lib/beaker/hypervisor/aws_sdk.rb
@@ -634,17 +634,33 @@ module Beaker
     # publicly. Then configure each ec2 machine to that dns_name, so that when facter
     # is installed the facts for hostname and domain match the dns_name.
     #
+    # if :use_beaker_hostnames: is true, set the :vmhostname and hostname of each ec2
+    # machine to the host[:name] from the beaker hosts file.
+    #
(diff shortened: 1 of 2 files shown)


kevpl:
looks good @doug-rosser! Would you add a section on how host names work in the AWS hypervisor in [those docs](https://github.com/puppetlabs/beaker/blob/master/docs/how_to/hypervisors/aws.md)?

doug-rosser:
Docs updated with PR https://github.com/puppetlabs/beaker/pull/1269

bmjen:
@kevpl is it possible to get this released sometime soon? I've confirmed that @doug-rosser 's fix also fixes my problem, which stems from mismatching cert names on master/agent runs on FOSS 3.x on AWS.

PR #1272, opened 2016-10-31
(QENG-4170) Compress files in archive_file_from method

smcelmurry:
Adding tarball compression to `archive_file_from` method. This will save space when archiving to jenkins masters or other file systems.


lib/beaker/dsl/helpers/host_helpers.rb
```diff
@@ -256,13 +256,21 @@ def rsync_to host, from_path, to_path, opts = {}
         #                               'archive/sut-files'.
         #
         # @return [Result] Returns the result of the #scp_from operation.
-        def archive_file_from(host, from_path, opts = {}, archive_root = 'archive/sut-files')
+        def archive_file_from(host, from_path, opts = {}, archive_root = 'archive/sut-files', archive_name = 'sut-files.tgz')
```
tvpartytonight: `archive_name` doesn't have yard docs yet; looks like it should be similar to the `archive_root` documentation.

acceptance/tests/base/dsl/helpers/host_helpers/archive_file_from_test.rb
```diff
@@ -19,17 +19,20 @@
     assert_equal(true, Dir.exists?('archive/sut-files'))
   end
 
-  step "file is copied to local <archiveroot>/<host name> directory" do
+  step "archive is copied to local <archiveroot>/<host name> directory" do
```
tvpartytonight: This step title indicates that it will write it to hostname directory, and I see the code that adds it in the `archive_file_from` method, but I don't see verification of the hostname in the filepath.

lib/beaker/dsl/helpers/host_helpers.rb
```diff
@@ -256,13 +256,21 @@ def rsync_to host, from_path, to_path, opts = {}
...
+          require 'archive/tar/minitar'
           filedir = File.dirname(from_path)
           targetdir = File.join(archive_root, host.hostname, filedir)
           FileUtils.mkdir_p(targetdir)
           scp_from(host, from_path, targetdir, opts)
+          create_tarball(archive_root, archive_name)
         end
 
+        def create_tarball(path, archive_name)
```
tvpartytonight: Does this need to be a private method? Maybe not, but, since it is in this host_helper file, it will be injected into the DSL unless you mark it as private. Do you expect users to use the method directly? If so, it should have some yard docs as well.

acceptance/tests/base/dsl/helpers/host_helpers/archive_file_from_test.rb
```diff
@@ -19,17 +19,20 @@
...
     # Create a remote file to archive
     filepath = default.tmpfile('archive-file-test')
```
tvpartytonight: I know this isn't part of your PR, but, do you know what happens if the `from_path` parameter is a directory and not a file? Or if it doesn't exist?

lib/beaker/dsl/helpers/host_helpers.rb
tvpartytonight: Privatize it! :)

beaker.gemspec
```diff
@@ -31,6 +31,7 @@ Gem::Specification.new do |s|
   # Run time dependencies
   s.add_runtime_dependency 'minitest', '~> 5.4'
   s.add_runtime_dependency 'json', '~> 1.8'
+  s.add_runtime_dependency 'minitar', '~> 0.5.4'
```
kevpl: This requirement should be below the comment about the json dependency. Or above both of them.

tvpartytonight:

lib/beaker/dsl/helpers/host_helpers.rb
```diff
@@ -254,15 +254,27 @@ def rsync_to host, from_path, to_path, opts = {}
...
+        def archive_file_from(host, from_path, opts = {}, archive_root = 'archive/sut-files', archive_name = 'sut-files.tgz')
+          require 'archive/tar/minitar'
           filedir = File.dirname(from_path)
           targetdir = File.join(archive_root, host.hostname, filedir)
           FileUtils.mkdir_p(targetdir)
           scp_from(host, from_path, targetdir, opts)
+          create_tarball(archive_root, archive_name)
         end
 
+        # @visibility private
```
tvpartytonight: I'm not sure what this comment does; I think that instead you want to mark the method private by calling the private method and passing a symbol with the method name.

```ruby
module Foo

  def bar
  end
  private :bar

end

````

Comment (kevpl) [inline lib/beaker/dsl/helpers/host_helpers.rb]: iirc, that comment wipes out the method from any representation in yard docs...

Comment (kevpl) [inline lib/beaker/dsl/helpers/host_helpers.rb]: I do agree with @tvpartytonight on this though that the ability to use this and the docs should be consistent. So either it:

1. doesn't need docs and should be marked private in Ruby (like Tony's suggesting)
1. needs to have functional docs

File: lib/beaker/dsl/helpers/host_helpers.rb [hunk 7]
```diff
@@ -254,15 +254,29 @@ def rsync_to host, from_path, to_path, opts = {}
...
         # @return [Result] Returns the result of the #scp_from operation.
-        def archive_file_from(host, from_path, opts = {}, archive_root = 'archive/sut-files')
+        def archive_file_from(host, from_path, opts = {}, archive_root = 'archive/sut-files', archive_name = 'sut-files.tgz')
+          require 'archive/tar/minitar'
           filedir = File.dirname(from_path)
           targetdir = File.join(archive_root, host.hostname, filedir)
           FileUtils.mkdir_p(targetdir)
-          scp_from(host, from_path, targetdir, opts)
+          result = scp_from(host, from_path, targetdir, opts)
+          return if result.exit_code != 0
```
kevpl: @smcelmurry so if the scp call fails, the circumstance here is that a user calls `archive_file_from`, and everything returns as if nothing went wrong (besides any reporting that `scp_from` might do itself), even though a failure happened?

It seems to me that we'll have a failure but won't return that that's the case, with the user only discovering this when they go to look at their archived files that don't exist. Is this not the case?

It also seems that `create_tarball` would be more resilient if it checked for file existence itself and failed if it was asked to tar a file that doesn't exist. Do you not agree, or is that being handled well by this code already that I'm not seeing?

smcelmurry: After a discussion IRL, it was discovered that `scp_from` isn't returning an error. I just updated this to raise an exception if the file doesn't exist within `archive_file_from`. Hopefully that should suffice!

smcelmurry:
Looks like a timeout with https://jenkins-beaker.delivery.puppetlabs.net/job/qe_beaker_intn-sys_beaker-acceptance-puppet-vpool-puppet-git/1975/agent=windows2008r2,master=centos7/

@kevpl @tvpartytonight It doesn't look related to the PR but I could be mistaken. Have you seen this before?

tvpartytonight:
Sounds like a possible transient to me; I'm going to kick it off again and see if it is reproducible. @puppetlabs-jenkins retest this please.

kevpl:
I'm 👍. @tvpartytonight, if you're satisfied, feel free to merge these changes. Remember, they will need to be squashed.

PR #1273, opened 2016-11-02
(BKR-971) Ensure scp errors close ssh connections

tvpartytonight:
Previous to this commit, beaker scp errors would leave the ssh
connection in a dangling state because of Exceptions that would bubble
up without closing out the ssh connection. This change ensures that when
errors are raised, the #close method is called. While the exceptions are
swallowed, the #close methods called end up raising errors from the
state of the connection object, thus not requiring that the original
asserts be raised.


rick:
saturday night live_ back to the future screen tests part 1 2 on vimeo
*(["Great, Scott!"](https://vimeo.com/50349978#t=123))*

+:100: to this.  I like the focused fix and :tada: on the tests as well.  Good work!

PR #1276, opened 2016-11-04
(BKR-976) Finalize 2.x dependencies

tvpartytonight:
In order to keep ruby dependencies from breaking the last 2.x version of
beaker, we can add a Gemfile.lock file to the release so that all
versions are pinned exactly and ranged versions cannot introduce
failures to bundle install.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,0 +1,269 @@
+PATH
+  remote: .
+  specs:
+    beaker (2.51.0)
+      aws-sdk-v1 (~> 1.57)
+      beaker-answers (~> 0.0)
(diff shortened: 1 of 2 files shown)


kevpl:
I've added these labels to this issue because I think it does require discussion, and that we're going to need to make sure to do more testing on this, particularly of modules and what the consequences will be there, where I don't think they have a separate Gemfile for acceptance testing like Puppet does. 

We're going to want to provide both documentation in-repo and in the release email on what changes are going to happen, and what people will need to do to keep working with beaker successfully in 2.x going forward if that's what they want.

PR #1277, opened 2016-11-08
(BKR-978) Updates beaker to work with SLES

trevor-vaughan:
Requires BKR-931 for AIO to work at all.

Tested against the following nodeset:

```

---
HOSTS:
   sles:
      roles:
         - default
         - master
      platform:   sles-12-x86_64
      box:        opensuse/openSUSE-Tumbleweed-x86_64
      hypervisor: vagrant
     
CONFIG:
  synced_folder : disabled
  log_level: verbose
  type: aio
```

---
This change is [Reviewable](https://reviewable.io/reviews/puppetlabs/beaker/1277)

### DIFF
--- a/lib/beaker/dsl/install_utils/foss_utils.rb
+++ b/lib/beaker/dsl/install_utils/foss_utils.rb
@@ -434,7 +434,11 @@ module Beaker
         # @api private
         def install_puppet_from_rpm_on( hosts, opts )
           block_on hosts do |host|
-            install_puppetlabs_release_repo(host)
+            if host[:type] == 'aio'
+              install_puppetlabs_release_repo(host,'pc1',opts)
+            else
+              install_puppetlabs_release_repo(host,nil,opts)
+            end
 
             if opts[:facter_version]
               host.install_package("facter-#{opts[:facter_version]}")
--- a/lib/beaker/dsl/install_utils/foss_utils.rb
+++ b/lib/beaker/dsl/install_utils/foss_utils.rb
@@ -266,6 +266,12 @@ module Beaker
               if host['platform'] =~ /el-(5|6|7)/
                 relver = $1
                 install_puppet_from_rpm_on(host, opts.merge(:release => relver, :family => 'el'))
+              elsif host['platform'] =~ /sles-(11|12)/
+                relver = $1
+                # Need to get rid of whatever Puppet might have shipped on the
+                # upstream images
+                host.uninstall_package('puppet')
+                install_puppet_from_rpm_on(host, opts.merge(:release => relver, :family => 'sles'))
               elsif host['platform'] =~ /fedora-(\d+)/
                 relver = $1
                 install_puppet_from_rpm_on(host, opts.merge(:release => relver, :family => 'fedora'))
@@ -436,6 +442,7 @@ module Beaker
           block_on hosts do |host|
             if host[:type] == 'aio'
               install_puppetlabs_release_repo(host,'pc1',opts)
+              opts[:version] = 'agent' unless opts[:version]
             else
               install_puppetlabs_release_repo(host,nil,opts)
             end
@@ -1079,7 +1086,7 @@ module Beaker
             onhost_copy_base = opts[:copy_dir_external] || host.external_copy_base
 
             case variant
-            when /^(fedora|el|centos|debian|ubuntu|cumulus|huaweios|cisco_nexus|cisco_ios_xr)$/
+            when /^(fedora|el|centos|sles|debian|ubuntu|cumulus|huaweios|cisco_nexus|cisco_ios_xr)$/
               if arch == 's390x'
                 logger.trace("#install_puppet_agent_dev_repo_on: s390x arch detected for host #{host}. using dev package")
               else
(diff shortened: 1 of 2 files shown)


lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -266,6 +266,12 @@ def install_puppet_on(hosts, opts={})
               if host['platform'] =~ /el-(5|6|7)/
                 relver = $1
                 install_puppet_from_rpm_on(host, opts.merge(:release => relver, :family => 'el'))
+              elsif host['platform'] =~ /sles-(11|12)/
+                relver = $1
+                # Need to get rid of whatever Puppet might have shipped on the
+                # upstream images
+                host.uninstall_package('puppet')
```
kevpl: @trevor-vaughan does the image you're using already have puppet on it, and that's why you're uninstalling it in the install method? 

@johnduarte, @cthorn42 are you guys aware of how our SLES testing images are configured? Or if not, would you point me to who is? I'd like to have them available to discuss our testing setup for this platform internally.

lib/beaker/dsl/install_utils/foss_utils.rb
```diff
@@ -434,7 +440,12 @@ def configure_puppet_on(hosts, opts = {})
         # @api private
         def install_puppet_from_rpm_on( hosts, opts )
           block_on hosts do |host|
-            install_puppetlabs_release_repo(host)
+            if host[:type] == 'aio'
+              install_puppetlabs_release_repo(host,'pc1',opts)
+              opts[:version] = 'agent' unless opts[:version]
```
kevpl: @trevor-vaughan what is `opts[:version]` doing being assigned to `'agent'`? Where else is this used such that you want make sure this is set here?

kevpl:
@trevor-vaughan this will need tests added. We don't have install docs up to the point of being able to require docs changes as well with this, iirc.

lib/beaker/dsl/install_utils/foss_utils.rb
trevor-vaughan: https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/install_utils/foss_utils.rb#L257-L261

lib/beaker/dsl/install_utils/foss_utils.rb
trevor-vaughan: @kevpl Yeah, the official LEAP images have Puppet and a bunch of other things installed so we need to rip them off first. Also, this step doesn't hurt since we're trying to install the "right" version.

The SUSE packages don't differentiate properly between the different versions of Puppet and will conflict.

trevor-vaughan:
@kevpl See Acceptance tests in the commit comment. I've been unable to decipher how to properly test this in RSpec but can certainly observe it working, and not working, with a simple Beaker run against the nodeset in the comment.

These errors would have been caught previously if tests had existed for the existing code.

Unfortunately, at this time, I don't have the cycles to add rspec tests, particularly when the evidence of functionality is concrete.

lib/beaker/dsl/install_utils/foss_utils.rb
johnduarte: @kevpl the vmpooler sles templates do not have puppet applications installed. As an example, here is the output from a zypper query for puppet on sles 11. It shows that the packages are available in the repos, but are not installed (lack of "i" in the first column).

```
# zypper search -s | grep puppet
  | puppet                                     | srcpackage | 2.6.18-0.4.2                  | noarch | oy.os
  | puppet                                     | package    | 2.6.18-0.4.2                  | x86_64 | oy.os
  | puppet-server                              | package    | 2.6.18-0.4.2                  | x86_64 | oy.os
  | rubygem-kiwi2puppet                        | package    | 0.1.0-0.8.2                   | x86_64 | oy.os

```

trevor-vaughan: @kevpl We need to be able to use the upstream images if possible. I get that you don't need to support that internally but users need to be able to test on the latest and greatest quite often.

kevpl: I'm down with that. @johnduarte do you know if there will be any issues running this command on our internal SLES platforms when puppet isn't installed?

lib/beaker/dsl/install_utils/foss_utils.rb
kevpl: it seems to me that there's an assumption that `opts[:version]` will be a version number. Can you point to a place where we're breaking that such that `agent` would be an appropriate value for this field?

trevor-vaughan:
FYI this is still broken in the latest release, as is the AIO issue in the dependency.

Will continue to direct folks to my branch in Slack and on the mailing list.

kevpl:
@trevor-vaughan since no one has cycles to add tests to these changes, I'm going to close this PR for lack of activity.

trevor-vaughan:
@kevpl Do you mind if I post to the mailing list noting that anyone using AIO and SLES and/or RHEL will need to use my PRs until someone adds tests that should have been in the original code to the code?

kevpl:
@trevor-vaughan not at all. You might be able to find someone who's willing to work with you on adding tests to this change so that we could get it merged there too, skipping the need for redirecting at all.

PR #1278, opened 2016-11-09
(PA-620) Run environment.bat before executing commands with cmd.exe

Magisus:
Previously, we were not properly setting up Windows environment
variables before executing arbitrary commands (e.g. 'gem'). This adds
the environment.bat script to the AutoRun registry key, which causes the
script to be run every time cmd.exe is launched, thereby setting up
necessary environment variables.

### DIFF
--- a/lib/beaker/dsl/install_utils/windows_utils.rb
+++ b/lib/beaker/dsl/install_utils/windows_utils.rb
@@ -167,6 +167,13 @@ exit /B %errorlevel%
             reg_query_command = %Q(reg query "#{reg_key}" /v "RememberedPuppetAgentStartupMode" | findstr #{msi_opts['PUPPET_AGENT_STARTUP_MODE']})
             on host, Command.new(reg_query_command, [], { :cmdexe => true })
 
+            # (PA-620) environment.bat should be run before any cmd.exe shell command
+            # in order to properly set up environment variables
+            auto_run_key = "HKLM\\SOFTWARE\\Microsoft\\Command Processor\AutoRun"
+            # TODO What's the best way to get the actual path to the script, respecting arch differences?
+            environment_script = "C:\\Program Files\\Puppet Labs\\Puppet\bin\\environment.bat"
+            reg_add_command = %Q(reg add "#{auto_run_key}" /v "PuppetEnvironment" /d "#{environment_script}")
+
             # emit the misc/versions.txt file which contains component versions for
             # puppet, facter, hiera, pxp-agent, packaging and vendored Ruby
             [


lib/beaker/dsl/install_utils/windows_utils.rb
```diff
@@ -167,6 +167,13 @@ def install_msi_on(hosts, msi_path, msi_opts = {}, opts = {})
             reg_query_command = %Q(reg query "#{reg_key}" /v "RememberedPuppetAgentStartupMode" | findstr #{msi_opts['PUPPET_AGENT_STARTUP_MODE']})
             on host, Command.new(reg_query_command, [], { :cmdexe => true })
 
+            # (PA-620) environment.bat should be run before any cmd.exe shell command
+            # in order to properly set up environment variables
+            auto_run_key = "HKLM\\SOFTWARE\\Microsoft\\Command Processor\AutoRun"
+            # TODO What's the best way to get the actual path to the script, respecting arch differences?
+            environment_script = "C:\\Program Files\\Puppet Labs\\Puppet\bin\\environment.bat"
```
Magisus: This is not the correct path on all hosts. How can I correctly determine whether I will need to look in `Program Files` or `Program Files (x86)` for the script?

joshcooper:
Sorry but I have to ask why we are working around the issue in our tests, instead of fixing the code to behave correctly? If we go this route, then `gem` commands only work when run through beaker, but not when run interactively by real users (outside of the puppet shell).

I know AIX had a test failures and we reverted. Do we know what the failure was? Is the new cert using a digest algorithm that our openssl compile on AIX doesn't support? Can we skip adding the new root cert on AIX? It seems like that would have a much smaller radius of impact.

Magisus:
We have not been able to track down the AIX failures (though I don't think anyone has fully dived into it). The error was during build, and came with no actual error message. It would probably be feasible to skip adding the new cert on AIX, since that platform wasn't affected by the cert change, so far as we know. For reference, previous fix was here: https://github.com/puppetlabs/puppet-agent/commit/480284ab77b3f142b6fd318214c5e6f91b062cce, and at very least we never wrote down a reason for abandoning that approach.

@puppetlabs/agent-team do any of you have more input for Josh on this?

joshcooper:
One other possibility here is to do nothing. It is not an issue in production CI because we set `GEM_SOURCE` to our internal mirror, resulting in puppet's acceptance setup executing:

```
11:55:50 m0net5hi2m0v1g5.delivery.puppetlabs.net (windows2012r2-64-1) 11:55:50$ env PATH="/cygdrive/c/Program Files (x86)/Puppet Labs/Puppet/sys/ruby/bin:/cygdrive/c/Program Files/Puppet Labs/Puppet/sys/ruby/bin:${PATH}" cmd /c gem source --add http://rubygems.delivery.puppetlabs.net
11:55:52   http://rubygems.delivery.puppetlabs.net added to sources
```

So I believe PA-620 is only an issue for those trying to execute `gem` commands on Windows and not using our internal mirror.

* For devs trying to run acceptance locally on Windows (HI-540, PUP-6819), just set `GEM_SOURCE` to our mirror. We should be doing this anyways so that local CI runs are consistent with production CI. Otherwise local tests may pass, but if you forget to mirror the new gem in stickler, production CI will fail.
* For Windows VMs in the classroom (PE-17845), this PR won't fix that problem. They will need to get an updated puppet-agent build with the new rubygems cert, or follow @Iristyle's blog post to install the cert for older puppet-agent builds.
* For devs trying to run Windows acceptance in EC2. I don't know that this PR would help. One option is to use the non-https rubygems endpoint.

Magisus:
Can confirm that setting the gem source works for me, and is an easy workaround for my use case.

tvpartytonight:
Given @joshcooper 's comments, I'm inclined agree with the workarounds. @Magisus is the behavior change proposed heresomething you still believe needs to be addressed in beaker itself?

Iristyle:
Thanks for the comments @joshcooper - sorry I missed them when you originally made them.

We looked at a bunch of options, and I was the one who made this `cmd.exe` `Autorun` reg key suggestion, knowing that it was less than ideal, but thinking that it addressed a more fundamental problem ... but maybe it's overkill / misguided at this time.

#### The Good

* It automatically pulls in the full set of `ENV` vars from [environment.bat](https://github.com/puppetlabs/puppet-agent/blob/master/resources/files/windows/environment.bat) into any command executed through  `cmd.exe` directly / or through a batch file in Cygwin.  Our documentation has always stated that if you want to touch the Ruby distribution, you have to run things in `Start Command Prompt with Puppet`.  Since Ruby commands like `gem` are batch files, they automatically run atop `cmd.exe` and therefore would automatically run `environment.bat` with this registry modification.  The vars set includes:

    * `PATH` - so that `puppet.bat`, `facter.bat` and friends are available
    * `RUBYLIB` - so the search path for Ruby code includes the location of Puppet
    * `FACTER_env_windows_installdir` fact - custom fact used by PE
    * `RUBYOPT` set to automatically load `rubygems` (though this many no longer be necessary)
    * `SSL_CERT_FILE` / `SSL_CERT_DIR` - to find additional certs - which fixes the PA-620 issue given we ship the CA that rubygems.org uses now

* It seems to me that Beaker should prep the command environment so that it behaves consistently with our recommendation... which it does so incompletely:

    * Beaker sets up PATH at https://github.com/puppetlabs/beaker/blob/baceb445aa52952954b3dce7e5aa26034f1ace2c/lib/beaker/dsl/install_utils/foss_utils.rb#L602-L603 but it feels a bit hacky since it's a Cygwin specific solution. 
    * It also adds to `PATH` via `configure_defaults_on`, which calls [`add_puppet_paths_on`](https://github.com/puppetlabs/beaker/blob/7439f35a4e24fc910502f32a2d3bbd9f98f07278/lib/beaker/dsl/install_utils/puppet_utils.rb#L93) which jams `puppetbindir` into `PATH`.  `add_env_var` has a [PS implementation](https://github.com/puppetlabs/beaker/blob/7439f35a4e24fc910502f32a2d3bbd9f98f07278/lib/beaker/host/pswindows/exec.rb#L86) or it falls back to the [`unix` implementation](https://github.com/puppetlabs/beaker/blob/7439f35a4e24fc910502f32a2d3bbd9f98f07278/lib/beaker/host/unix/exec.rb#L101-L117) I believe when running under Cygwin.
    * My point here is why not just setup the environment by calling `environment.bat` which encapsulates that knowledge, rather than Beaker understanding how to rig up `PATH` based on package type / shell type.

* I believe this could open the door for removing the [Cygwin sshd restart](https://github.com/puppetlabs/beaker/blob/baceb445aa52952954b3dce7e5aa26034f1ace2c/lib/beaker/dsl/install_utils/windows_utils.rb#L212-L214) (though there's probably already a nicer way to reload the current profile that doesn't forcefully terminate the SSH connection)

#### The Bad

* Using the `AutoRun` reg key is a little sneaky and since there's only one command allowed, could conflict with how a VM template is setup if that feature is already being used (unlikely I think)

* For batch files that already call `environment.bat` (like `puppet.bat`) this means that `PATH` is modified twice if using `AutoRun` reg key.

* This does not handle pulling in an environment when running commands directly in PowerShell - that would require calling `environment.bat` / collecting up the ENV vars in PowerShell too (best done within `$PROFILE` I think).  It's not a ton of work to do that, but it requires some more code to be added to Beaker.

* This doesn't help old Puppet installations that don't have the right certs in the package, which is what PA-620 is all about.  For packages that shipped prior to the cert bundle being incorporated into the MSI, a solution is still needed.

The future:

* It might make sense to modify the Ruby batch files to call `environment.bat` themselves, so that if `gem` is invoked, it's already configured things properly.  Of course, that only affects new packages - but this will work across all shells.

* We've talked about laying out the package differently so that `environment.bat` goes away - but I think we're still a ways off, and that doesn't help with older / already released package versions either.

Other note:

* Beaker does have a [install_cert_on_windows helper](https://github.com/puppetlabs/beaker/blob/baceb445aa52952954b3dce7e5aa26034f1ace2c/lib/beaker/dsl/install_utils/foss_utils.rb#L1207) which would be a good way to get newer certs into the right place.

#### Summary

My proposition might be overly ambitious and still not solve the immediate problem at hand, so I'm fine with ticketing that and deferring something to later.... I still think it would be a good idea long-term to find a way to get the full environment ready so that no pre-suite or other changes are required to get Beaker configured properly for the installed package.

Iristyle:
Closing this PR since PA-620 has been closed.

This issue will eventually be reopened in a new PR for https://tickets.puppetlabs.com/browse/BKR-986

joshcooper:
Sorry meant to comment earlier. All of the reasons you mention make sense.

I think my main concern is since this is beaker, it would affect all `cmd.exe` invocations, even if you're executing a command unrelated to puppet. And since we're modifying the `PATH` it could lead to surprises.

Beaker does add the "public" bin directory to the `PATH` on *nix and Windows via the ssh environment. But adding `environment.bat` would add in the [private bin](https://github.com/puppetlabs/puppet-agent/blob/master/resources/files/windows/environment.bat#L15) parts too.

PR #1279, opened 2016-11-10
(BKR-40) Pin net-ssh gem to 3.3.0beta1

Iristyle:
- This version of net-ssh includes UTF-8 related fixes when commands or paths exceed 127 characters.
   Unfortunately the author is discontinuing any releases in the 2.* or 3.* series, hasn't yet released any non-alpha or beta
   4.* releases, and has versioned this last release in the 3 series with a beta1 name, so it must be referenced explicitly.

### DIFF
--- a/beaker.gemspec
+++ b/beaker.gemspec
@@ -38,7 +38,7 @@ Gem::Specification.new do |s|
   ## aws-sdk-2 doesn't require json, so we can give this up when we move
 
   s.add_runtime_dependency 'hocon', '~> 1.0'
-  s.add_runtime_dependency 'net-ssh', '~> 3.2'
+  s.add_runtime_dependency 'net-ssh', '3.3.0.beta1'
   s.add_runtime_dependency 'net-scp', '~> 1.2'
   s.add_runtime_dependency 'inifile', '~> 2.0'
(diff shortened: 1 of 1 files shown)


beaker.gemspec
```diff
@@ -38,7 +38,7 @@ Gem::Specification.new do |s|
   ## aws-sdk-2 doesn't require json, so we can give this up when we move
 
   s.add_runtime_dependency 'hocon', '~> 1.0'
-  s.add_runtime_dependency 'net-ssh', '~> 3.2'
+  s.add_runtime_dependency 'net-ssh', '3.3.0beta1'
```
kevpl: looks like this needs to be `3.3.0.beta1`? Gem wasn't on our mirror (yay manual mirroring), but it failed to mirror with the version string here.

Iristyle: Thanks - repushed.

Also didn't realize that GitHub was going to put the branch on this repo, rather than my fork - so we'll have to delete the branch once merged (assuming it works)

MosesMendoza:
👍

Iristyle:
@tvpartytonight didn't you mention there were some problems with this when using `gem install`?

PR #1280, opened 2016-11-14
(BKR-982) OpenStack Cinder V2

spjmurray:
Fog defaults to the V2 API on clouds that support it, but the parameters
have changed, in particular `display_name` becomes `name`.  This checks
the cinder client type and uses the old value for V1 and the new one for
newer versions.

### DIFF
--- a/docs/how_to/hypervisors/openstack.md
+++ b/docs/how_to/hypervisors/openstack.md
@@ -123,3 +123,32 @@ user their own pool.  It's used in allocating new IPs.  It's an options
 parameter in the CONFIG section of the host file:
 
     floating_ip_pool: 'my_pool_name'
+
+### Volumes
+
+Attaching volumes to a VM is supported via the Cinder service.  All versions are transparently
+supported to cater for differences in the APIs.  To create and attach volumes simply add hash
+called 'volumes' to a host in the HOSTS section.  Each key is the name given to the volume upon
+resource creation.  The value is a hash with a single integer parameter 'size' which defines the
+volume size in MB.
+
+**Example OpenStack hosts file with volumes**
+
+    HOSTS:
+      ceph:
+        roles:
+          - master
+        platform: ubuntu-16.04-amd64
+        hypervisor: openstack
+        flavor: m1.large
+        image: xenial-server-cloudimg-amd64-scsi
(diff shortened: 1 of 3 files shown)


pinkypie:
@spjmurray do you know which version of Fog introduces this change?

spjmurray:
@pinkypie you were right to question my logic, seemingly our service catalog updated without my knowledge, event though we *should* have supported V2 for the last year!  I've updated the change to inspect the client type and construct the arguments appropriately.  The new stuff is the default so should be forward compatible with cinder V2, V3 etc.

kevpl:
@spjmurray would you please provide some tests for this? This also seems like something where some docs would be helpful.

spjmurray:
@kevpl Shakes fist - I hate writing rspec :smiley: Okay done that, basically also flexes all of my volume provisioning code.  As such needed a few modifications in order to stub things in ways that make it all possible.  As regards documentation, what are you after, links to cinder APIs or just something to say 'this parameter changed between versions'?

kevpl:
@spjmurray 😆 I can definitely understand that.

In terms of the docs, I'd say anything above the 'this parameter changed between versions' that includes the names of what it went from and to would be great. Links to Cinder API docs would be great for people to understand the change, but they're by no means necessary.

kevpl:
@spjmurray it doesn't look like you added docs. Am I just missing them? If not, do you have any questions about writing them?

kevpl:
@spjmurray last comment is still current

spjmurray:
@kevpl I did add a comment to lib/beaker/hypervisor/openstack.rb:170 but I assume you want more.  Where do you want me to put the documentation?  Sorry for the delay getting lost in xmas etc!

kevpl:
@spjmurray yeah, something in the "Openstack-Specific Hosts File Settings" section of our [openstack hypervisor doc](https://github.com/puppetlabs/beaker/blob/master/docs/how_to/hypervisors/openstack.md#openstack-specific-hosts-file-settings) would be what I'm looking for.

spjmurray:
Back in the office tomorrow, will sort you out first thing now adventofcode.com is finished *cry*

tvpartytonight:

PR #1282, opened 2016-11-17
(BKR-987) Remove explicit json gem dependency

Iristyle:
- aws-sdk-v1 transitively requires the `json` gem, which was added to
   the Gemfile in 8c76afcce87cf7c23adaa6d7c76f00b7040a2e54

   However, it doesn't appear that the `json` version for Beaker was
   specified for any compelling reason, and can cause issues on Windows
   systems when other gems requiring `json` are added to a Gemfile that
   also references Beaker, for instance with puppet modules like
   Registry or ACL.

   Even in Puppet modules, where `bundle install --without system_tests`
   is specified to prevent Beaker installation, if other gems require the
   `json` gem as a dependency, Bundler looks at the requirements of the
   `beaker` gem and its dependencies when determining which `json` gem
   version satisfies the requirement.

   This can lead to Bundler attempting to install a newer `json` gem
   than the one that is Bundled with the Ruby distribution.
   Attempting to install the `json` gem requires the DevKit be installed
   on Windows, as it includes native code that needs to be compiled -
   and therefore is inappropriate for many CI testing scenarios.

 - Realistically `json` should no longer even require being explicitly
   referenced given it ships with Ruby based on the following info:

   - JSON 1.4.2 with 1.9.2 -
     https://github.com/ruby/ruby/blob/ruby_1_9_2/ext/json/lib/json/version.rb
   - JSON 1.5.5 with 1.9.3 -
     https://github.com/ruby/ruby/blob/ruby_1_9_3/ext/json/lib/json/version.rb
   - JSON 1.7.7 with 2.0.0 series -
     https://github.com/ruby/ruby/blob/ruby_2_0_0/ext/json/lib/json/version.rb
   - JSON 1.8.1 with 2.1.0 - through 2.1.7 -
     https://github.com/ruby/ruby/blob/v2_1_7/ext/json/lib/json/version.rb
   - JSON 1.8.1 with 2.2.0 - through 2.2.3 -
     https://github.com/ruby/ruby/blob/v2_2_3/ext/json/lib/json/version.rb

 - Unfortunately, this solution only appears to fix `gem install` type
   workflows, but does not fix `bundler` based workflows when using a
   local gem cache specified with `--path`. While Bundler claims to
   support using both system gems (like the `json` gem shipped with
   Ruby) AND gems specified in the local `Gemfile`, when `--path` is
   specified, Bundler sets its `BUNDLE_DISABLE_SHARED_GEMS: 1` in the
   local .bundle/config, preventing the system `json` from being used.
   In testing, it seemed that there was no way to override this
   behavior.

   That said, should other gem dependencies later remove their explicit
   `json` requirements, at least removing the `json` gem as a Beaker
   dependency will prevent Beaker itself from causing installations.

 - Note Beaker 3 only runs atop Ruby 2, which always includes at least
   `json` 1.7.7.

### DIFF
--- a/beaker.gemspec
+++ b/beaker.gemspec
@@ -33,9 +33,6 @@ Gem::Specification.new do |s|
   # Run time dependencies
   s.add_runtime_dependency 'minitest', '~> 5.4'
   s.add_runtime_dependency 'minitar', '~> 0.5.4'
-  s.add_runtime_dependency 'json', '~> 1.8'
-  ## json: will stay <2.0 while aws-sdk-v1 is in use
-  ## aws-sdk-2 doesn't require json, so we can give this up when we move
 
   s.add_runtime_dependency 'hocon', '~> 1.0'
   s.add_runtime_dependency 'net-ssh', '3.3.0.beta1'


beaker.gemspec
```diff
@@ -33,7 +33,7 @@ Gem::Specification.new do |s|
   # Run time dependencies
   s.add_runtime_dependency 'minitest', '~> 5.4'
   s.add_runtime_dependency 'minitar', '~> 0.5.4'
-  s.add_runtime_dependency 'json', '~> 1.8'
+  s.add_runtime_dependency 'json', '~> 1.4'
```
Iristyle: Maybe better to just pull the dependency altogether and let transitive deps control this.

hunner:
This would also satisfy https://tickets.puppetlabs.com/browse/BKR-987

beaker.gemspec
ferventcoder: Maybe, but I kind of like it being at least a bit explicit. I would almost recommend `~> 1`, just to catch the upgrade to json 2.x

ferventcoder:
I'm semi- 👍  on this. I'm curious if that relaxing actually works or if it still attempts to resolve the gem instead of the built-in version?

beaker.gemspec
joshcooper: It would be nice if we could drop it like rails did. It puts us closer to allowing ruby to just use the builtin version. For example, [`multi_json.gemspec`](https://github.com/intridea/multi_json/blob/master/multi_json.gemspec) doesn't list `json` as a dependency.

If we drop it and some transitive dependency has an explicit `json` dependency (with or without a version), then I think we're back to where we are now -- bundler will try to download the latest json gem that satisfies the constraints.

That could result in us getting a 2.x `json` version. Is that bad? Beaker#master requires ruby >= 2.2.5 and json 2 requires ruby > 2, but beyond that I'm not sure what breakage could result from json 2.

ferventcoder: I don't think it's bad. Kind of why I was just saying at least version 1, but really anything. I guess really just dropping the restriction altogether is 👍

Iristyle: After some more reading / experimenting, I think dropping is the best bet.

However, I'm still not getting the desired behavior in my testing with this change.  Bundler is supposed to be OK with using a mix of system gems and local gems (isolated with `--path`) as long as you are not using its `--deployment` switch.

However, in testing with this combination of gems I'm still seeing Bundler try to pull from rubygems when the local system gem satisfies the dependency requirements.

I've tried explicitly specifying `= 1.5.5` when on Ruby 1.9.3 on Windows *and* I've tried omitting the `json` dependency altogether.  In both cases, it contacts `rubygems` to acquire gem downloads, and will pull one.  When `= 1.5.5` is specified, it downloads it despite having the same version installed as a system gem.  When omitted, it will download `1.8.3`.

`.bundle/config` contains a variable `BUNDLE_DISABLE_SHARED_GEMS` which is supposed to influence Bundlers behavior - but I can't seem to override it at all (through environment variable or command line switch).  If I edit that file to allow shared gems (i.e. the system ones), then proceed to call `bundle install --path .bundle/gems` it reverts the value.  So I'm wondering if bundler is just buggy here?

Note that I have seen the desired behavior when simply using `gem install` (like for the azure sdk gem, which previously had similar problems)

Iristyle: Looks like this is sort-of documented on https://github.com/bundler/bundler/issues/4571 - but remains inconsistent with what documentation seems to imply.

PR #1283, opened 2016-11-17
(BKR-703) Utilize service reload in bounce_service helper method

camlow325:
This commit modifies the bounce_service helper method to try to restart
the service via a `reload` action, falling back to a `stop` and `start`
only on failure.  This change enables beaker to take advantage of
performance improvements which are possible with `reload` when the
service supports it.

### DIFF
--- a/lib/beaker/dsl/helpers/puppet_helpers.rb
+++ b/lib/beaker/dsl/helpers/puppet_helpers.rb
@@ -314,7 +314,7 @@ module Beaker
         #
         # @param [Host] host Host the service runs on
         # @param [String] service Name of the service to restart
-        # @param [Fixnum] curl_retries Number of times to retry the restart command
+        # @param [Fixnum] curl_retries Number of seconds to wait for the restart to complete before failing
         # @param [Fixnum] port Port to check status at
         #
         # @return [Result] Result of last status check
@@ -327,8 +327,14 @@ module Beaker
             apachectl_path = host.is_pe? ? "#{host['puppetsbindir']}/#{service}" : service
             host.exec(Command.new("#{apachectl_path} graceful"))
           else
-            host.exec puppet_resource('service', service, 'ensure=stopped')
-            host.exec puppet_resource('service', service, 'ensure=running')
+            result = host.exec(Command.new("service #{service} reload"),
+                               :acceptable_exit_codes => [0,1,3])
+            if result.exit_code == 0
+              return result
+            else
+              host.exec puppet_resource('service', service, 'ensure=stopped')
+              host.exec puppet_resource('service', service, 'ensure=running')
+            end
           end
           curl_with_retries(" #{service} ", host, "https://localhost:#{port}", [35, 60], curl_retries)
         end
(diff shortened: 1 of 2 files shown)


camlow325:
We've already been experimenting with this change in the full OSS puppetserver#master pipeline, which includes the entire acceptance test suite from the puppetlabs/puppet repo.  We've done our normal full OS matrix runs -- with master/agent role combos.  No changes to any of the existing acceptance tests were needed in order for beaker - via `bounce_service` - to use "reload" to reload the puppetserver service rather than "stop" and "start" and for the tests to continue passing.

For reference purposes, here are some links to Jenkins runs done with Beaker 2.50.0 vs. with these changes (based on code near the Beaker 3.40 tag) and Total Suite Times for the tests being run. 

**Beaker runs with bounce_service using "reload"**

https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppetserver/view/all/job/platform_puppetserver_integration-system_no-conditional_full-master/23/

cent7-64ma-64a

- 91 minutes

https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppetserver/view/all/job/platform_puppetserver_integration-system_no-conditional_full-master/22

redhat7-64ma-windows2008r2-64a

- 105 minutes

——
 
**Beaker runs with bounce_service using "start/stop"**

https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppetserver/view/all/job/platform_puppetserver_integration-system_no-conditional_full-master/21

cent7-64ma-64a

- 166 minutes

https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppetserver/view/all/job/platform_puppetserver_integration-system_no-conditional_full-master/20/

redhat7-64ma-windows2008r2-64a

- 188 minutes

lib/beaker/dsl/helpers/puppet_helpers.rb
```diff
@@ -327,8 +327,14 @@ def bounce_service host, service, curl_retries = nil, port = nil
...
             host.exec(Command.new("#{apachectl_path} graceful"))
           else
-            host.exec puppet_resource('service', service, 'ensure=stopped')
-            host.exec puppet_resource('service', service, 'ensure=running')
+            result = host.exec(Command.new("service #{service} reload"),
+                               :acceptable_exit_codes => [0,1,3])
+            if result.exit_code == 0
+              return result
+            else
+              host.exec puppet_resource('service', service, 'ensure=stopped')
```
camlow325: This approach allows the helper to fall back to using the existing stopped/running logic in the event that the `reload` command fails - which could happen if the service had not been started when the `bounce_service` call is made or if the service does not support the `reload` command.  For the latter, this would allow a newer Beaker which has this logic to still be able to restart the service properly even when run against older versions of Puppet Server, when necessary.

cprice404:
> 166 minutes -> 91 minutes

effyeah

er0ck:
i too have tried this against puppet component acceptance.
it requires removing this:
https://github.com/puppetlabs/puppet/blob/master/acceptance/config/aio/options.rb#L6
(or setting it to true)

er0ck:
huge 👍 , let's get this in and reduce all our acceptance runtimes by like half

kevpl:
NICE 🚀 🚀

PR #1284, opened 2016-11-17
(BKR-987) Remove explicit json gem dependency

Iristyle:
- aws-sdk-v1 transitively requires the `json` gem, which was added to
   the Gemfile in 8c76afcce87cf7c23adaa6d7c76f00b7040a2e54

   However, it doesn't appear that the `json` version for Beaker was
   specified for any compelling reason, and can cause issues on Windows
   systems when other gems requiring `json` are added to a Gemfile that
   also references Beaker, for instance with puppet modules like
   Registry or ACL.

   Even in Puppet modules, where `bundle install --without system_tests`
   is specified to prevent Beaker installation, if other gems require the
   `json` gem as a dependency, Bundler looks at the requirements of the
   `beaker` gem and its dependencies when determining which `json` gem
   version satisfies the requirement.

   This can lead to Bundler attempting to install a newer `json` gem
   than the one that is Bundled with the Ruby distribution.
   Attempting to install the `json` gem requires the DevKit be installed
   on Windows, as it includes native code that needs to be compiled -
   and therefore is inappropriate for many CI testing scenarios.

 - Realistically `json` should no longer even require being explicitly
   referenced given it ships with Ruby based on the following info:

   - JSON 1.4.2 with 1.9.2 -
     https://github.com/ruby/ruby/blob/ruby_1_9_2/ext/json/lib/json/version.rb
   - JSON 1.5.5 with 1.9.3 -
     https://github.com/ruby/ruby/blob/ruby_1_9_3/ext/json/lib/json/version.rb
   - JSON 1.7.7 with 2.0.0 series -
     https://github.com/ruby/ruby/blob/ruby_2_0_0/ext/json/lib/json/version.rb
   - JSON 1.8.1 with 2.1.0 - through 2.1.7 -
     https://github.com/ruby/ruby/blob/v2_1_7/ext/json/lib/json/version.rb
   - JSON 1.8.1 with 2.2.0 - through 2.2.3 -
     https://github.com/ruby/ruby/blob/v2_2_3/ext/json/lib/json/version.rb

 - Unfortunately, this solution only appears to fix `gem install` type
   workflows, but does not fix `bundler` based workflows when using a
   local gem cache specified with `--path`. While Bundler claims to
   support using both system gems (like the `json` gem shipped with
   Ruby) AND gems specified in the local `Gemfile`, when `--path` is
   specified, Bundler sets its `BUNDLE_DISABLE_SHARED_GEMS: 1` in the
   local .bundle/config, preventing the system `json` from being used.
   In testing, it seemed that there was no way to override this
   behavior.

   That said, should other gem dependencies later remove their explicit
   `json` requirements, at least removing the `json` gem as a Beaker
   dependency will prevent Beaker itself from causing installations.

 - Note Beaker 3 only runs atop Ruby 2, which always includes at least
   `json` 1.7.7.

### DIFF
--- a/beaker.gemspec
+++ b/beaker.gemspec
@@ -56,6 +56,10 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'fog-google', '~> 0.0.9' # dropped ruby 1.9 support in 0.1
   s.add_runtime_dependency 'fog', ['~> 1.25', '< 1.35.0']
 
+  # webmock requires addressable as as of 2.5.0 addressable started
+  # requiring the public_suffix gem which requires Ruby 2
+  s.add_runtime_dependency 'addressable', '< 2.5.0'
+
   # So fog doesn't always complain of unmet AWS dependencies
   s.add_runtime_dependency 'unf', '~> 0.1'
 end
--- a/beaker.gemspec
+++ b/beaker.gemspec
@@ -30,7 +30,6 @@ Gem::Specification.new do |s|
 
   # Run time dependencies
   s.add_runtime_dependency 'minitest', '~> 5.4'
-  s.add_runtime_dependency 'json', '~> 1.8'
   s.add_runtime_dependency 'hocon', '~> 1.0'
   s.add_runtime_dependency 'net-ssh', '~> 2.9'
   s.add_runtime_dependency 'net-scp', '~> 1.2'


tvpartytonight:
@Iristyle you added the don't merge label to this PR, but, I'm unclear as to why we shouldn't be merging it; I want to merge and release #1276 for this Wednesday, but it seems like it would be a good idea to get this particular pinning in prior to that. Alternatively, we could merge this in after the #1276 but I'd rather not have to do yet another 2.x release if we don't have to.

Can you elaborate on why you don't want this merged?

Iristyle:
@tvpartytonight I just updated this PR with some more details about the `json` conundrum.  Sorry I didn't get to this earlier.

I'll update the other Beaker 3 PR as well - and you can decide to do a merge-up of this PR OR merge that PR.

tvpartytonight:
When we merged in the Gemfile.lock in #1276 we ended up pinning the json gem to version 1.8.3. This of course can change, but it won't happen just by removing the the json dependency declared in beaker itself. Looking at the fe69184 commit, it looks like we would also have to use a different version of [simplecov](https://github.com/puppetlabs/beaker/commit/fe69184c51e416fa5ebab3bb99c9f23bae3303cc#diff-e79a60dc6b85309ae70a6ea8261eaf95R242) as well. If you still need to go down the path of pushing this dependency all the way back to `json 1.4.z`, you'll need to change the simplecov dependency as well.

kevpl:
last comment is current

tvpartytonight:
Closing this due to inactivity and the fact that 2.x is essentially a no longer supported version. @Iristyle I'd suggest making a PR to the master branch if this is something you'd like to see in the current, active branch of beaker.

PR #1285, opened 2016-11-22
(BKR-981) Make beaker know where the values in the options came from

ferglor:


lib/beaker/cli.rb
```diff
@@ -13,11 +13,15 @@ def initialize
       @timestamp = Time.now
       @options_parser = Beaker::Options::Parser.new
       @options = @options_parser.parse_args
+      @attribution = @options_parser.attribution
       @logger = Beaker::Logger.new(@options)
       InParallel::InParallelExecutor.logger = @logger
       @options[:logger] = @logger
+      @attribution[:logger] = 'runtime'
```
ferglor: I don't like how I'm having to update the attribution hash every time the options hash is updated, so I was wondering what people's thoughts were on wrapping the update of the two hashes in a function something like:

def update_option(key, value, source)
  @options[key] = value
  @attribution[key] = source
end

So we could just replace the @options and @attribution update line pairs with:

update_option(:logger, @logger, 'runtime')

lib/beaker/options/parser.rb
```diff
@@ -181,9 +203,12 @@ def initialize
       # @raise [ArgumentError] Raises error on bad input
       def parse_args(args = ARGV)
         @options                        = @presets.presets
+        tag_sources(@presets.presets, "default")
```
kevpl: Using the value "default" creates an unnecessary mapping here between the presets that already exist and this keyword. We should eliminate unnecessary mappings.

lib/beaker/options/parser.rb
```diff
@@ -181,9 +203,12 @@ def initialize
...
         cmd_line_options                = @command_line_parser.parse(args)
         cmd_line_options[:command_line] = ([$0] + args).join(' ')
+        tag_sources(cmd_line_options, "flag")
         file_options                    = Beaker::Options::OptionsFileParser.parse_options_file(cmd_line_options[:options_file])
+        tag_sources(file_options, "file")
```
kevpl: we have two different file attribution values, so this should contain more detail. `options_file`, perhaps?

lib/beaker/options/parser.rb
```diff
@@ -165,6 +166,34 @@ def initialize
         @command_line_parser = Beaker::Options::CommandLineParser.new
         @presets             = Beaker::Options::Presets.new
         @validator           = Beaker::Options::Validator.new
+        @attribution         = Beaker::Options::OptionsHash.new
```
kevpl: @ferglor I thought you were going to encapsulate the attribution work within the `OptionsHash` object? Did you mean that you'd just use an `OptionsHash` object to keep track of attribution then?

ferglor: @kevpl sorry man, this update was mainly aimed at recursively updating the attribution hash for child items, e.g. the SSH attribute. I intended to make further updates to encapsulate the attribution hash within the OptionsHash, but I got sidetracked with some CI.next stuff. I'll work on these changes today

kevpl:

PR #1286, opened 2016-11-23
(MAINT) clarify puppet_agent_version parameter

kevpl:
Clarifies the puppet_agent_version parameter.

Also pulls some value setting logic out of a loop based on
hosts iteration, since that it's required for the value setting itself.


kevpl:
@Iristyle does this look good for clarification?

ferglor:
👍

PR #1287, opened 2016-11-23
(MAINT) assert on OSX pkg info requirement

kevpl:
Doing work for 4be19a3, I found that
the OS X package info method uses the
option's :download_url parameter, but
doesn't require or assert that it exists.
Added an assertion to validate it's
existance since it's necessary.


ferglor:
👍

PR #1288, opened 2016-12-07
(BKR-1006) Close connection immediately when rebooting Windows

MikaelSmith:
When using `Windows::Exec::reboot`, the SSH connection is expected to
fail. However, it can take a long time to do so (greater than 30 minutes
in some cases). This should be faster, as we know the connection will
fail.

Add a new option to `exec`, `:reset_connection`, that immediately closes
the connection after sending a message. Use it when performing reboot on
a Windows host.

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -340,6 +340,13 @@ module Beaker
           @logger.debug "\n#{log_prefix} executed in %0.2f seconds" % seconds
         end
 
+        if options[:reset_connection]
+          # Expect the connection to fail hard and possibly take a long time timeout.
+          # Pre-emptively reset it so we don't wait forever.
+          close
+          return result
+        end
+
         unless options[:silent]
           # What?
           result.log(@logger)
--- a/lib/beaker/host/windows/exec.rb
+++ b/lib/beaker/host/windows/exec.rb
@@ -2,7 +2,7 @@ module Windows::Exec
   include Beaker::CommandFactory
 
   def reboot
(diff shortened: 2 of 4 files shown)


MikaelSmith:
Ideally this would probably just be the behavior of `:expect_connection_failure`. However, that can be used with beaker helpers that may be used elsewhere, so I'm not sure of the impact of changing the behavior of that option. I'm also not sure whether it makes sense to apply the new behavior for all `reboot` methods; this particular problem only seems to happen with Windows hosts.

Iristyle:
FYI - there's an acceptance test at https://github.com/puppetlabs/beaker/blob/master/acceptance/tests/base/host/reboot_test.rb for this... and when I ran it yesterday (without your change), it basically hung indefinitely.

You should be able to run via something like:

```
TEST=./acceptance/tests/base/host/user_test.rb bundle exec rake test:base
```

MikaelSmith:
The acceptance test seems to work with my change. Ran as
```
TEST=./acceptance/tests/base/host/reboot_test.rb bundle exec rake test:base TEST_TARGET=windows2012r2-64a
```

samwoods1:
This doesn't seem like it would help if there was even a slight delay between issuing the command and the connection failure.  Ideally we would call close at the top of the the while loop in wait_for_connection_failure.  I can't think of any reason why we wouldn't want to do this.  This would just ensure we're using a fresh ssh connection on each attempt to connect.

MikaelSmith:
I assumed explicitly closing the connection would circumvent the point of waiting for the connection failure. And `:expect_connection_failure` is documented in facter/puppet/host helpers, so others might be relying on that behavior.

Also, I don't understand your comment
> doesn't seem like it would help if there was even a slight delay between issuing the command and the connection failure.

PR #1290, opened 2016-12-08
(BKR-1009) Don't fail when ip is unavailable in Vagrantfile

jrwesolo:
I was running into an issue when using the `vagrant_custom` hypervisor and not using any private_network with a known ip address (only the NAT interface, no hostonly). In all examples, I am using the virtualbox provider.

The first run with provisioning worked perfectly. If I ran it a second time without reprovisioning I would get an exception stating "Unable to determine ip for vagrant box #{hostname}". This was because only on reprovisioning does beaker attempt to parse out an ip address from the Vagrantfile.

I propose that instead of raising an exception, we just return `nil` for the ip address. As far as I can tell, beaker connects to the machines using the ssh config file generated from the `set_ssh_config` method. I compared the `host.host_hash` from a fresh run and a run without provisioning and they seem functionally equivalent (the ip address was undefined vs nil, respectively). Testing so far doesn't seem to have any negative impact (I've tried with the `vagrant_custom` and `vagrant` hypervisor).

### DIFF
--- a/lib/beaker/hypervisor/vagrant.rb
+++ b/lib/beaker/hypervisor/vagrant.rb
@@ -147,7 +147,8 @@ module Beaker
           ip = m[2]
           @logger.debug("Determined existing vagrant box #{hostname} ip to be: #{ip} ")
         else
-          raise("Unable to determine ip for vagrant box #{hostname}")
+          ip = nil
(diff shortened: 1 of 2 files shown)


jrwesolo:
Closing this and reopening a new one so I follow the standards outlined in CONTRIBUTING.md.

PR #1292, opened 2016-12-08
(BKR-1010) Fix Cygwin install_package helper

Iristyle:
- Newer Windows templates in vmpooler have been produced through a
   Packer process, and install packages differently than earlier
   templates based on how Cygwin was installed.

   Two additional switches must now be passed for package installs to
   succeed using Cygwins package manager.

   See https://tickets.puppetlabs.com/browse/IMAGES-424 for more detail

   This fixes a bug where `cmdline_args` passed to Cygwins setup*.exe
   are injecetd into the wrong location.  When specify a -R, the
   rootdir value must be passed next.  Previously, additional args
   were injected in between the -R and the rootdir value.

### DIFF
--- a/acceptance/tests/base/host/packages.rb
+++ b/acceptance/tests/base/host/packages.rb
@@ -60,7 +60,11 @@ hosts.each do |host|
 
   assert_equal(false, host.check_for_package(package), "'#{package}' not should be installed")
   logger.debug("#{package} should be installed on #{host}")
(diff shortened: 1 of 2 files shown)


Iristyle:
Verified as fixed after running:

```sh
TEST=./acceptance/tests/base/host/packages.rb bundle exec rake test:base TEST_TARGET=windows2012r2-64a
```

PR #1295, opened 2016-12-13
(BKR-1012) Fix Host#ip on Non-English Windows

Iristyle:
- The current method of detecting a Windows host IP relies on parsing
   very specific output from the `ipconfig` command using either
   `grep` in Cygwin, or `find` in cmd / PowerShell.  This method
   relies on finding the strings `IP Address`, `IPv4 Address` or
   `IPv6 Address` in the output.

   On internationalized versions of Windows, the word `Address` is
   typically localized, as is the case with Japanese Windows. Further,
   calling `grep` against such output, must have `LC_ALL=C` passed as
   well, otherwise text won't be matched.

 - Furthermore, the current method of using `ipconfig` has other
   problems because it may return adapters that are not currently
   IPEnabled, which could lead to retrieval of the an IP address
   associated with a Pseudo-Interface.  By specifying IPEnabled=True
   the list of addresses returned is more limited.

 - The output from WMIC can look like any of the following with
   /list:format specified, and the code is designed to grab the first
   returned non-empty value:

   IPAddress=
   IPAddress={"129.168.0.1"}
   IPAddress={"192.168.0.1","2001:db8:aaaa:bbbb:cccc:dddd:eeee:0001"}

### DIFF
--- a/lib/beaker/host/pswindows/exec.rb
+++ b/lib/beaker/host/pswindows/exec.rb
@@ -42,13 +42,23 @@ module PSWindows::Exec
   end
 
   def get_ip
-    ip = execute("for /f \"tokens=14\" %f in ('ipconfig ^| find \"IP Address\"') do @echo %f", :accept_all_exit_codes => true).strip
-    if ip == ''
-      ip = execute("for /f \"tokens=14\" %f in ('ipconfig ^| find \"IPv4 Address\"') do @echo %f", :accept_all_exit_codes => true).strip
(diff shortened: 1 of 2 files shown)


Iristyle:
There is an existing acceptance that proves `Host#ip` continues to work properly.

Iristyle:

Iristyle:
@kevpl / @tvpartytonight this PR is now all green as well... huzzah.

PR #1299, opened 2016-12-16
(BKR-1016) resolve stub forge from SUT

johnduarte:
This commit modifies the implementation of the `stub_forge_on` and
`with_forge_stubbed_on` methods to resolve the `forge_host` from
the host on which the targeted `forge_host` will be accessed.

This is done to address the case where the SUT and the Beaker
runner have different resolution of the given `forge_host` name.

### DIFF
--- a/lib/beaker/dsl/helpers/puppet_helpers.rb
+++ b/lib/beaker/dsl/helpers/puppet_helpers.rb
@@ -9,6 +9,23 @@ module Beaker
       # for these methods to execute correctly
       module PuppetHelpers
 
+        # Return the regular expression pattern for an IPv4 address
+        def ipv4_regex
+          return /(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)/
+        end
+
+        # Return the IP address that given hostname returns when resolved on
+        # the given host.
+        #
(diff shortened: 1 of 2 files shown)


lib/beaker/dsl/helpers/puppet_helpers.rb
```diff
@@ -606,7 +606,10 @@ def stub_forge_on(machine, forge_host = nil)
           #use global options hash
           primary_forge_name = 'forge.puppetlabs.com'
           forge_host ||= options[:forge_host]
-          @forge_ip ||= Resolv.getaddress(forge_host)
+          ipv4_re = /(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)/
+          match = curl_on(machine, "--verbose #{forge_host}", :accept_all_exit_codes => true).stderr.match(ipv4_re)
+          raise "Failed to resolve forge host '#{forge_host}'" unless match
+          @forge_ip ||= match[0]
```
kevpl: @johnduarte this all looks good, but could we pull these blocks into a separate helper method in this file to be used by both methods since it's duplicate code (especially the gross regex)?

johnduarte: @kevpl the PR has been updated with separate helpers for the regex and the resolve operations.
Thanks.

PR #1301, opened 2016-12-16
(BKR-1018) copy_ssh_to_root not working on windows

azhurbilo:
https://www.microsoft.com/resources/documentation/windows/xp/all/proddocs/en-us/xcopy.mspx?mfr=true

> /i : If Source is a directory or contains wildcards and Destination does not exist, xcopy assumes destination specifies a directory name and creates a new directory. Then, xcopy copies all specified files into the new directory. By default, xcopy prompts you to specify whether Destination is a file or a directory.
> 
> /y : Suppresses prompting to confirm that you want to overwrite an existing destination file.


azhurbilo:
Also it should be applied for 3.x versions too but I don't know about procedure. What I need to do?

PR #1303, opened 2016-12-19
(BKR-1018) copy_ssh_to_root not working on windows

azhurbilo:
https://www.microsoft.com/resources/documentation/windows/xp/all/proddocs/en-us/xcopy.mspx?mfr=true

> /i : If Source is a directory or contains wildcards and Destination does not exist, xcopy assumes destination specifies a directory name and creates a new directory. Then, xcopy copies all specified files into the new directory. By default, xcopy prompts you to specify whether Destination is a file or a directory.
> 
> /y : Suppresses prompting to confirm that you want to overwrite an existing destination file.

### DIFF
--- a/lib/beaker/host_prebuilt_steps.rb
+++ b/lib/beaker/host_prebuilt_steps.rb
@@ -329,7 +329,7 @@ module Beaker
           host.exec(Command.new('cp -r .ssh /cygdrive/c/Users/Administrator/.'))
           host.exec(Command.new('chown -R Administrator /cygdrive/c/Users/Administrator/.ssh'))
         elsif host['platform'] =~ /windows/ and not host.is_cygwin?
-          host.exec(Command.new("if exist .ssh (xcopy .ssh C:\\Users\\Administrator\\.ssh /s /e)"))
+          host.exec(Command.new("if exist .ssh (xcopy .ssh C:\\Users\\Administrator\\.ssh /s /e /y /i)"))
         elsif host['platform'] =~ /osx/
           host.exec(Command.new('sudo cp -r .ssh /var/root/.'), {:pty => true})
         elsif host['platform'] =~ /freebsd/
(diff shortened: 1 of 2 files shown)


lib/beaker/host_prebuilt_steps.rb
```diff
@@ -329,7 +329,7 @@ def copy_ssh_to_root host, opts
           host.exec(Command.new('cp -r .ssh /cygdrive/c/Users/Administrator/.'))
           host.exec(Command.new('chown -R Administrator /cygdrive/c/Users/Administrator/.ssh'))
         elsif host['platform'] =~ /windows/ and not host.is_cygwin?
-          host.exec(Command.new("if exist .ssh (xcopy .ssh C:\\Users\\Administrator\\.ssh /s /e)"))
+          host.exec(Command.new("if exist .ssh (xcopy .ssh C:\\Users\\Administrator\\.ssh /s /e /y /i)"))
```
kevpl: @azhurbilo would you add tests to confirm we don't break this fix in future work? It would be great if you just included these doc references from your comment above in them so that we have the context there as well.

azhurbilo: Test was added.

But I didn't see any examples in other tests how properly add doc references.

kevpl: @azhurbilo yeah, I was thinking of just taking what you have in the first comment and adding that above this line in the if to give context to the command itself, rather than actual doc additions. So it would end up looking like this:

```ruby
elsif host['platform'] =~ /windows/ and not host.is_cygwin?
  # from https://www.microsoft.com/resources/documentation/windows/xp/all/proddocs/en-us/xcopy.mspx?mfr=true:
  #    /i : If Source is a directory or contains wildcards and Destination
  #      does not exist, xcopy assumes destination specifies a directory
  #      name and creates a new directory. Then, xcopy copies all specified
  #      files into the new directory. By default, xcopy prompts you to
  #      specify whether Destination is a file or a directory.
  #
  #    /y : Suppresses prompting to confirm that you want to overwrite an
  #      existing destination file.
  host.exec(Command.new("if exist .ssh (xcopy .ssh C:\\Users\\Administrator\\.ssh /s /e /y /i)"))
elsif host['platform'] =~ /osx/
```

It's a bit lengthy, but I'd rather have a lot of context for something specific like that if I'm coming back to it later for anything.

azhurbilo: done

kevpl:

kevpl:

PR #1306, opened 2017-01-05
Fix for enable_root_login on OSX 10.11 & 10.12

wilson208:
This fixes enable_root_login for osx 10.11 and 10.12 where the wrong ssh config file was being used.

### DIFF
--- a/lib/beaker/host_prebuilt_steps.rb
+++ b/lib/beaker/host_prebuilt_steps.rb
@@ -409,10 +409,14 @@ module Beaker
     def enable_root_login host, opts
       logger = opts[:logger]
       block_on host do |host|
-        logger.debug "Update /etc/ssh/sshd_config to allow root login"
+        logger.debug "Update sshd_config to allow root login"
         if host['platform'] =~ /osx/
-          host.exec(Command.new("sudo sed -i '' 's/#PermitRootLogin no/PermitRootLogin Yes/g' /etc/sshd_config"))
-          host.exec(Command.new("sudo sed -i '' 's/#PermitRootLogin yes/PermitRootLogin Yes/g' /etc/sshd_config"))
+          # If osx > 10.10 use '/private/etc/ssh/sshd_config', else use '/etc/sshd_config'
+          ssh_config_file = '/private/etc/ssh/sshd_config'
+          ssh_config_file = '/etc/sshd_config' if host['platform'] =~ /^osx-10\.(9|10)/i
+
+          host.exec(Command.new("sudo sed -i '' 's/#PermitRootLogin no/PermitRootLogin Yes/g' #{ssh_config_file}"))
+          host.exec(Command.new("sudo sed -i '' 's/#PermitRootLogin yes/PermitRootLogin Yes/g' #{ssh_config_file}"))
         elsif host['platform'] =~ /freebsd/
           host.exec(Command.new("sudo sed -i -e 's/#PermitRootLogin no/PermitRootLogin yes/g' /etc/ssh/sshd_config"), {:pty => true} )
         elsif host['platform'] =~ /openbsd/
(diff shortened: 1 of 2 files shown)


er0ck:
ugh... this works well enough as a fix, and i'm +1 (probably needs tests).
but surely the Host object should know where its sshd config goes?

wilson208:
Have now added some spec tests

tvpartytonight:

wilson208:
@tvpartytonight any way I could get this merged now its green in Jenkins?

lib/beaker/host_prebuilt_steps.rb
```diff
@@ -411,8 +411,10 @@ def enable_root_login host, opts
       block_on host do |host|
         logger.debug "Update /etc/ssh/sshd_config to allow root login"
```
tvpartytonight: nit: this should be updated to just `sshd_config` in the debug message, as the location varies per version.

lib/beaker/host_prebuilt_steps.rb
```diff
@@ -411,8 +411,10 @@ def enable_root_login host, opts
...
         if host['platform'] =~ /osx/
-          host.exec(Command.new("sudo sed -i '' 's/#PermitRootLogin no/PermitRootLogin Yes/g' /etc/sshd_config"))
-          host.exec(Command.new("sudo sed -i '' 's/#PermitRootLogin yes/PermitRootLogin Yes/g' /etc/sshd_config"))
+          ssh_config_file = '/etc/sshd_config'
+          ssh_config_file = '/private/etc/ssh/sshd_config' if host['platform'] =~ /osx-10\.*(11|12)/
```
tvpartytonight: To future-proof this a bit, we should just assume that `/private/etc/ssh/sshd_config` will be the location for future versions of osx. We should change this logic so that osx-1013, osx-1014, and so on will automatically get this new location, not the location that existed in osx-1010.

wilson208:
@tvpartytonight thanks for feedback, have updated to reflect that. Now using a regex to check for versions <= 10.10 and set old location, rather than specifically set new location for 10.11 and 10.12, so should be future proofed.

Thanks.

tvpartytonight:
@puppetlabs-jenkins retest this please.

tvpartytonight:

lib/beaker/host_prebuilt_steps.rb
```diff
@@ -409,10 +409,14 @@ def disable_updates hosts, opts
...
       logger = opts[:logger]
       block_on host do |host|
-        logger.debug "Update /etc/ssh/sshd_config to allow root login"
+        logger.debug "Update sshd_config to allow root login"
         if host['platform'] =~ /osx/
-          host.exec(Command.new("sudo sed -i '' 's/#PermitRootLogin no/PermitRootLogin Yes/g' /etc/sshd_config"))
-          host.exec(Command.new("sudo sed -i '' 's/#PermitRootLogin yes/PermitRootLogin Yes/g' /etc/sshd_config"))
+          # If osx > 10.10 use '/private/etc/ssh/sshd_config', else use '/etc/sshd_config'
+          ssh_config_file = '/private/etc/ssh/sshd_config'
+          ssh_config_file = '/etc/sshd_config' if host['platform'] =~ /^osx-10\.[1-9]{1}[0]{0,1}$/i
```
tvpartytonight: @wilson208 I'm not sure if this regex is working as intended; also, it seems to me that it could be written a bit clearer. How does the following regex look to you? 

````
^osx-10\.(9|10)
````

As far as I can tell, the regex you have won't ever match because of the `$` at the end means that the string terminates, but host platform strings still have an architecture type appended after the version.

wilson208:
@tvpartytonight I have updated this with the regex you suggested above. I was attempting to write a regex that worked for older osx version than 10.9 and when I was testing the regex, wasn't accounting for the architecture (doh). Thanks for the feedback!!

tvpartytonight:
LGTM

PR #1307, opened 2017-01-06
(PDB-3193) Ensure key pair exists after creation

kbarber:
This patch immediately checks a keypair was created so we can detect early
any key creation failures or slowness for further debugging later.

This was created to debug issues we're having in PDB around failed key deletion. Basically, a keypair gets created for the first node, removal is meant to occur normally afterwards for the next creation - however there is a race in that they key shows up as not existing (even though its happy to use it for node creation - yeah, AWS can be weird). So we get an exception and the entire run fails.

https://jenkins-enterprise.delivery.puppetlabs.net/job/enterprise_puppetdb_integration-system-puppetdb_full-master/1398/INSTALL_TYPE=upgrade_latest,LAYOUT=ec2-west-el6-64mda-el6-64a,NIGHTLY=true,PUPPETDB_DATABASE=postgres,label=beaker-ec2/console

Or:

https://gist.github.com/kbarber/b26971516c9b53013ad490f414c05177

All this patch does, is check immediately through a query that the key was created.
The side-effect that I didn't expect, is that this
extra check seems to improve the real problem in tests. See the following PR
for the green passes to prove this:

https://github.com/puppetlabs/puppetdb/pull/2135

I believe this is mainly due to add a bit of time, as the extra time helps with
the wait of creation for keypairs, before the deletion later on, thus seemingly "fixing" the issue
but really I believe adding attenuation.

I don't expect this to fix the situation always, but the extra checking and logging
will help users understand the root cause if the bug resurfaces further down the road. With
this check in place, action can be taken in another patch to 'try again' perhaps, or something
like that ... but I figured that was premature. Since this seems to improve things, and
the logging will help us understand it if it re-occurs, for now this feels enough. Not to mention
I can't reproduce the race with this patch, so its hard to develop on :-).

Anyway, this is pretty important to us as we are currently getting red PuppetDB runs all
the time due to this bug in aws-sdk (not just this bug obviously, but its a big contributor).

Thanks for listening.

### DIFF
--- a/lib/beaker/hypervisor/aws_sdk.rb
+++ b/lib/beaker/hypervisor/aws_sdk.rb
@@ -407,7 +407,7 @@ module Beaker
         end
         wait_for_status(:running, instances)
       rescue Exception => ex
-        @logger.notify("aws-sdk: exception - #{ex}")
+        @logger.notify("aws-sdk: exception #{ex.class}: #{ex}")
         kill_instances(instances.map{|x| x[:instance]})
         raise ex
       end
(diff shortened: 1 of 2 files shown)


kbarber:

PR #1308, opened 2017-01-06
(BKR-940) Update package testing to puppet 4.8.0

tvpartytonight:


kevpl:
Wouldn't this be nicer if we had these values parameter-ized so that we could control them from the job rather than needing code changes? Below is one simple idea of how we might do this..
```ruby
:version        => ENV['BEAKER_PUPPET_VERSION'] || '4.8.0',
```
What do you think?

tvpartytonight:
@ferglor if you merge this and rebase your PR, you should be good to go.

PR #1309, opened 2017-01-09
(MAINT): Use mac address format for randmac that libvirt understands.

Ntr0:
This patch fixes the randmac method of the vagrant libvirt provider
to generate mac addresses that are in a format libvirt understands.

So instead of 080027aabbcc it will generate 08:00:27:aa:bb:cc


tvpartytonight:
Closing as a duplicate of #1313

PR #1311, opened 2017-01-11
(BKR-1029) Updated the gen_hosts quickstart task to accept a prefix

ferglor:


ferglor:
Execution looks like:

`$> bundle exec beaker init --hypervisor=vmpooler
About to run - beaker-hostgenerator redhat7-64default.mdcal-redhat7-64af
Writing default host config to file - ./.beaker/acceptance/config/default_vmpooler_hosts.yaml
WARNING: Starting with beaker-hostgenerator 1.x platform strings for "el" hosts
will correspond to the actual linux distribution name. ie, the platform string
corresponding to a host specified as "centos4_64a" will be "centos-4-x86_64"
rather than "el-4-x86_64". It is recommended that you update your project's test
suites ASAP or be forced to do so when beaker-hostgenerator development moves on
to the 1.x series. We don't intend to backport features or platforms to 0.x.`

and:

`$> bundle exec beaker init --hypervisor=vagrant
About to run - beaker-hostgenerator ubuntu1404-64default.mdcal-ubuntu1404-64af, --hypervisor=vagrant, --global-config={box_url=https://vagrantcloud.com/puppetlabs/boxes/ubuntu-14.04-64-nocm,box=puppetlabs/ubuntu-14.04-64-nocm}
Writing default host config to file - ./.beaker/acceptance/config/default_vagrant_hosts.yaml
WARNING: Starting with beaker-hostgenerator 1.x platform strings for "el" hosts
will correspond to the actual linux distribution name. ie, the platform string
corresponding to a host specified as "centos4_64a" will be "centos-4-x86_64"
rather than "el-4-x86_64". It is recommended that you update your project's test
suites ASAP or be forced to do so when beaker-hostgenerator development moves on
to the 1.x series. We don't intend to backport features or platforms to 0.x.`

kevpl:
@ferglor I'd prefer not to add parameters to rake tasks & keep them as simple as possible. Is there a particular reason to add this?

Also, I'm a little confused as to the output above. Would you put line breaks in it so that it presents as it would on the CLI? Or does it actually present like that?

ferglor:
@kevpl sorry man, the code block I had was only wrapped in single back ticks

How should we go about storing the init artifacts in a .beaker/ folder? Are we eventually going to move away from using the quick start tasks, and implement the init functionality from the ground up as part of the subcommand work? If so, should we hold off on the .beaker/ folder stuff until then?

kevpl:
@ferglor I'm fine with storing the artifacts as you are in that folder, but the problem is that adding parameters to rake tasks just gives us another input layer.

The rake tasks are there for convenience only, they're not another way to configure beaker. We should only allow the minimum amount of things to be configurable through them, and if people want better handles into configuring beaker itself, then they should use beaker itself, starting by tweaking the commands that are run from the rake tasks themselves if they'd like.

In conclusion, I think that you have the right location, but that we shouldn't have this parameterized. This is us being opinionated about where this stuff goes, & if people want to be more customized, they should really use beaker, not use these rake tasks.

ferglor:
@kevpl oh, so just update the rake tasks to always write artifacts to .config?

kevpl:
`.config`? I thought you were using `.beaker`?

but yes, just make them always use the value, rather than taking a parameter to determine it.

ferglor:
@kevpl yes! sorry, watching the pdw training in the background

kevpl:
lol, me too

lib/beaker/tasks/quick_start.rb
```diff
@@ -9,17 +9,17 @@
...
     elsif args[:hypervisor] == 'vmpooler'
       cli = VMPOOLER
     else
       puts "No hypervisor provided, defaulting to vagrant."
-      hosts_file = "acceptance/config/default_vagrant_hosts.yaml"
+      hosts_file = ".beaker/acceptance/config/default_vagrant_hosts.yaml"
       cli = VAGRANT
     end
-    FileUtils.mkdir_p('acceptance/config') # -p ignores when dir already exists
+    FileUtils.mkdir_p(".beaker/acceptance/config") # -p ignores when dir already exists
```
kevpl: @ferglor since we're repeating the `.beaker/acceptance/config` path 3x here, it would be an improvement to pull this path out into a constant.

ferglor:

kevpl:

lib/beaker/tasks/quick_start.rb
```diff
@@ -9,17 +11,17 @@
...
     elsif args[:hypervisor] == 'vmpooler'
       cli = VMPOOLER
     else
       puts "No hypervisor provided, defaulting to vagrant."
-      hosts_file = "acceptance/config/default_vagrant_hosts.yaml"
+      hosts_file = "#{CONFIG_DIR}/default_vagrant_hosts.yaml"
       cli = VAGRANT
     end
-    FileUtils.mkdir_p('acceptance/config') # -p ignores when dir already exists
+    FileUtils.mkdir_p("#{CONFIG_DIR}") # -p creates intermediate directories as required
```
kevpl: @ferglor merging now, but for future reference, you don't need to do interpolation here since the variable is a string anyways.

PR #1314, opened 2017-01-13
(maint) Move rake runtime dependency

lucywyman:


tvpartytonight:
@lucywyman given that we can't introduce this with version 11 because [puppet itself](https://github.com/puppetlabs/puppet/blob/master/acceptance/Gemfile#L18) is pinned to 10, can you change this to be on version 10? Not sure if any of the implementation from @ferglor was dependent on that version...

lucywyman:
Changed!

PR #1316, opened 2017-01-17
(BKR-1035) Environment variable keys always get upper-cased

tdevelioglu:
This commit fixes an issue where the `environment_string` method of a
host unconditionally upper-cased the keys of the environment variables
of a command

### DIFF
--- a/lib/beaker/host/unix/exec.rb
+++ b/lib/beaker/host/unix/exec.rb
@@ -246,7 +246,7 @@ module Unix::Exec
       else
         val = val.to_s
       end
-      env_array << "#{key.to_s.upcase}=\"#{val}\""
+      env_array << "#{key.to_s}=\"#{val}\""
     end
     env_array
   end
--- a/spec/beaker/host/cisco_spec.rb
+++ b/spec/beaker/host/cisco_spec.rb
@@ -107,8 +107,8 @@ module Cisco
 
         it 'turns env maps into paired strings correctly' do
           @options = { :user => 'root' }
-          env_map = { 'var1' => 'ans1', 'var2' => 'ans2' }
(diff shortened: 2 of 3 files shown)


kevpl:

kevpl:
@puppetlabs/windows any worries here from your side? It looks like the `pswindows` host object uses `environment_variable_string_pair_array` [here](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host/pswindows/exec.rb#L163), but we're not certain if you guys are using that host object for Powershell testing anymore...

Is there an assumption on your beaker tests that these will be upper cased?

glennsarti:
Environment variables in Windows are not case sensitive, so `%HERE%` is the same as `%HeRe%`.  Unlike *nix where they are.  Potentially, by not upcasing we will see false negatives on tests on Windows.

lib/beaker/host/unix/exec.rb
```diff
@@ -246,7 +246,7 @@ def environment_variable_string_pair_array env
       else
         val = val.to_s
       end
-      env_array << "#{key.to_s.upcase}=\"#{val}\""
```
joshcooper: This has the likely potential to break existing tests that are relying on the old beaker behavior. If we are going to release this in beaker 3.x, then we need to preserve the old behavior as well (so set both the `upcase` and original values). On Windows, they env variables will overwrite each other, with the same value, so not an issue. On *nix, there will be two env variables, but they will have the same value, so again no issue.

kevpl: @joshcooper thanks for looking at this! I'll get a PR with that up this week under this same ticket reference to get this fixed!

PR #1323, opened 2017-01-20
(BKR-995) Update requirement for net-ssh

tvpartytonight:
Previous to this commit, we pinned net-ssh for fixes related to
UTF-8. This had the unfortunate side effect of breaking "gem install
beaker", as a Gem::DependencyError prevented installation because Gem
couldn't tell that version "3.3.0.beta1" met the dependency for "=>
2.6.5" for net-scp.. Now that version 4.0 has been released, we should
 move the pin to that version so that `gem install beaker` can happen
successfully from rubygems.

### DIFF
--- a/acceptance/tests/base/host/host_test.rb
+++ b/acceptance/tests/base/host/host_test.rb
@@ -316,3 +316,17 @@ step "Ensure scp errors close the ssh connection" do
     on default, 'true'
   end
 end
+
+step 'Ensure that a long 128+ character string with UTF-8 characters does not break net-ssh' do
+  long_string = 'a' * 128 + "\u06FF"
+  on(default, "mkdir /tmp/#{long_string}")
+  result = on(default, 'ls /tmp')
+  assert(result.stdout.include?(long_string), 'Error in folder creation with long string + UTF-8 characters')
+
+  # remove the folder
+  on(default, "rm -rf /tmp/#{long_string}")
+  result = on(default, 'ls /tmp')
+  assert(!result.stdout.include?(long_string), 'Error in folder deletion with long string + UTF-8 characters')
+
+end
+
--- a/beaker.gemspec
+++ b/beaker.gemspec
@@ -34,7 +34,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'minitar', '~> 0.5.4'
 
   s.add_runtime_dependency 'hocon', '~> 1.0'
-  s.add_runtime_dependency 'net-ssh', '3.3.0.beta1'
+  s.add_runtime_dependency 'net-ssh', '~> 4.0'
   s.add_runtime_dependency 'net-scp', '~> 1.2'
   s.add_runtime_dependency 'inifile', '~> 2.0'
   ## inifile: keep <3.0, breaks puppet_helpers.rb:puppet_conf_for when updated


tvpartytonight:
@Iristyle I'm assuming that version 4.0.0 has the commits from 3.3.0.beta1 that you needed; can you let me know what pipeline I should test specifically so I can verify this doesn't break?

tvpartytonight:

tvpartytonight:

tvpartytonight:

tvpartytonight:
pinging @Iristyle; do you know what repo I can test this dependency change with to ensure that the required changes to net-ssh 3.3.0.beta1 are available in net-ssh >= 4.0.0 ?

Iristyle:
Write a test that uses a 128 character or longer path that includes UTF-8 characters. Here's some hints:

```ruby
# different UTF-8 widths
# 1-byte A
# 2-byte ۿ - http://www.fileformat.info/info/unicode/char/06ff/index.htm - 0xDB 0xBF / 219 191
# 3-byte ᚠ - http://www.fileformat.info/info/unicode/char/16A0/index.htm - 0xE1 0x9A 0xA0 / 225 154 160
# 4-byte ܎ - http://www.fileformat.info/info/unicode/char/2070E/index.htm - 0xF0 0xA0 0x9C 0x8E / 240 160 156 142
mixed_utf8 = "A\u06FF\u16A0\u{2070E}"  # Aۿᚠ܎
```

acceptance/tests/base/host/host_test.rb
```diff
@@ -316,3 +316,17 @@
     on default, 'true'
   end
 end
+
+step 'Ensure that a long 128+ character string with UTF-8 characters does not break net-ssh' do
```
tvpartytonight: @Iristyle this test passes with `net-ssh 4.0.0` and `net-ssh 3.3.0.beta1`, although you'll have to take my word for with the beta1 version that I've tested manually. With `net-ssh 3.2.0`, it fails with the following error:

 ````
e9r2bzses1bvojd.delivery.puppetlabs.net (windows2012r2-64-1) 13:00:03$ mkdir /tmp/testingareallylongcharacterlengththatisatleast128characterslongnothinglessmoreisfinebutthislineshouldbethatlongwithsomeUTF8attheendAۿᚠ𠜎
    Encoding::CompatibilityError: incompatible character encodings: ASCII-8BIT and UTF-8
    /Users/toneloc/repos/beaker/vendor/bundle/ruby/2.3.0/gems/net-ssh-3.2.0/lib/net/ssh/buffer.rb:284
````

Comment (Iristyle) [inline acceptance/tests/base/host/host_test.rb]: Yeah, I take your word for it - that's all the behavior that I would expect.

👍

File: acceptance/tests/base/host/host_test.rb [hunk 2]
```diff
@@ -316,3 +316,17 @@
...
+  long_string = "testingareallylongcharacterlengththatisatleast128characterslongnothinglessmoreisfinebutthislineshouldbethatlongwithsomeUTF8attheendA\u06FF\u16A0\u{2070E}"
```
Iristyle: A simpler version should be sufficient:

```ruby
long_string = ('a' * 128) + "\u06FF"
```

Iristyle:
Suggested a minor test mod, but otherwise 👍

kevpl:
@tvpartytonight I'd prefer that these tests be put in an SSH-specific test file, since they're not actually testing the host object itself but something it contains. However, I feel like the host tests are a bucket already, so it's fine to merge this as-is, and leave that re-org for another time.

PR #1325, opened 2017-01-24
(BKR-1046) Use pdb status for started validation

johnduarte:
This commit changes the implementation of the
`sleep_until_puppetdb_started` helper method to query the PuppetDB
status endpoint to verify that the PDB web services are available.

### DIFF
--- a/lib/beaker/dsl/helpers/puppet_helpers.rb
+++ b/lib/beaker/dsl/helpers/puppet_helpers.rb
@@ -676,7 +676,10 @@ module Beaker
         def sleep_until_puppetdb_started(host, nonssl_port = nil, ssl_port = nil)
           nonssl_port = options[:puppetdb_port_nonssl] if nonssl_port.nil?
           ssl_port = options[:puppetdb_port_ssl] if ssl_port.nil?
-          curl_with_retries("start puppetdb", host, "http://localhost:#{nonssl_port}", 0, 120)
+          endpoint = 'status/v1/services/puppetdb-status'
+          retry_on(host,
+                   "curl -m 1 http://localhost:#{nonssl_port}/#{endpoint} | grep '\"state\":\"running\"'",
+                   {:max_retries => 120})
           curl_with_retries("start puppetdb (ssl)",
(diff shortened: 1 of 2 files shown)


johnduarte:
Ping @tvpartytonight . Looking for a review on this. Trying to get it into Beaker 3.11.0. Cheers.

lib/beaker/dsl/helpers/puppet_helpers.rb
```diff
@@ -676,7 +676,10 @@ def stub_forge(forge_host = nil)
         def sleep_until_puppetdb_started(host, nonssl_port = nil, ssl_port = nil)
           nonssl_port = options[:puppetdb_port_nonssl] if nonssl_port.nil?
           ssl_port = options[:puppetdb_port_ssl] if ssl_port.nil?
-          curl_with_retries("start puppetdb", host, "http://localhost:#{nonssl_port}", 0, 120)
+          endpoint = 'status/v1/services/puppetdb-status'
+          retry_on(host,
+                   "curl -m 1 http://localhost:#{nonssl_port}/#{endpoint} | grep '\"state\":\"running\"'",
+                   {:max_retries => 120})
           curl_with_retries("start puppetdb (ssl)",
```
tvpartytonight: @johnduarte I know this isn't part of your code change, but it seems to me that this `curl_with_retries` is not necessary. If the `retry_on` host command is successful, isn't puppetdb ostensibly listening to the ssl port as well? What do you think about removing that `curl_with_retries` after the `retry_on`?

johnduarte: @tvpartytonight The call to `curl_with_retries` validates that the {{ssl}} port is open after we have verified that the service is available on the {{nonssl_port}}. So, I believe it is still required in order to make sure that the {{ssl}} port has been successfully opened.

tvpartytonight: I think that the `curl_with_retries` should be replaced same test as the `retry_on`—namely, it should also look for a status of running—but I'd like to standardize on how we attach certs to requests here, so it's not necessary for this PR.

PR #1326, opened 2017-01-24
(BKR-937) Add support for installing AIX 7 agent on 7.2

branan:
To avoid unnecessarily expanding our build times, our initial support
for AIX 7.2 is through the existing 7.1 package. This will be replaced
with a single unified AIX build in the Agent 2.0 timeframe, but for now
we need to add a little bit of special-case logic to handle 7.2 with
existing packages.


branan:
This looks like possibly a false positive at first glance? @puppetlabs/beaker-maintainers any thoughts?

tvpartytonight:

tvpartytonight:
Test failure in regression is not related to this change.

PR #1328, opened 2017-01-25
(BKR-1047) Don't change module paths for the DSL

tvpartytonight:
This reverts commit 4d0e29fe1c7dc842c171f89f2acf6c74cba4ed42, reversing
changes made to 59cfccedf25a4cdce6296fe010ee22bba07e5855.

In our documentation, we make references to only utilizing elements of
the the DSL directly, instead of using them as a lump sum in TestCase.
During a major version, these paths should never change.


nwolfe:
🙈

ferglor:
👍

PR #1330, opened 2017-01-30
(BKR-1051) Add retry loop for aws key check

mullr:
The kb.exists? check fails intermittently; it one of the failures, manual
inspection shows that the key pair was in fact created. This heavily implies
that there's a race at work; where the key pair doesn't show as existing until
some time after the initial creation.

Add a retry loop around the kb.exists? check to attempt to work around this
issue.

### DIFF
--- a/lib/beaker/hypervisor/aws_sdk.rb
+++ b/lib/beaker/hypervisor/aws_sdk.rb
@@ -811,7 +811,18 @@ module Beaker
       ssh_string = public_key()
       region.key_pairs.import(pair_name, ssh_string)
       kp = region.key_pairs[pair_name]
-      if kp.exists?
+
+      exists = false
+      for tries in 1..10
+        if kp.exists?
+          exists = true
+          break
+        end
+        @logger.debug("AWS key pair doesn't appear to exist yet, sleeping before retry ")
+        backoff_sleep(tries)
+      end
+
+      if exists
         @logger.debug("aws-sdk: key pair #{pair_name} imported")
         kp
       else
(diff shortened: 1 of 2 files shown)


mullr:
@tvpartytonight any thoughts? look ok?

lib/beaker/hypervisor/aws_sdk.rb
```diff
@@ -811,7 +811,18 @@ def create_new_key_pair(region, pair_name)
       ssh_string = public_key()
       region.key_pairs.import(pair_name, ssh_string)
       kp = region.key_pairs[pair_name]
-      if kp.exists?
+
+      exists = false
+      for tries in 1..10
```
tvpartytonight: The exponential backoff here seems a little high; @mullr have you seen how long it takes when this needs to wait? 2**10 is an extra 25 minutes from 2**8, which is about 8 minutes of waiting and retrying.

mullr: I actually just copied this from a similar backoff above; I'll turn it down.

mullr: As far as how long it takes when it needs to wait: we've only seen this intermittently in CI, so I haven't actually seen it happen with this code.

mullr: @tvpartytonight changed to 1..5

tvpartytonight:

mullr:
@tvpartytonight looks like tests are green now. ok to merge?

tvpartytonight:
👍

PR #1331, opened 2017-01-30
BKR-922: Obsolete url path in Beaker causing smoke failures

mchllweeks:


tvpartytonight:
Failure in CI unrelated to code change; reran PE 2016.5.1 regression tests, and everything passed.

PR #1333, opened 2017-02-01
(BKR-1052) Document user config option for ec2 hypervisor

vinzent:
Without the user option beaker fails to login.

```
  Please login as the user "ec2-user" rather than the user "root".
```

Setting the user triggers sanitizing of sshd_config and
authorized_keys.


kevpl:
@vinzent thanks for submitting this fix!

PR #1335, opened 2017-02-03
(maint) fix references to files under concepts

johnmccabe:
Prior to this commit we were linking to the `docs/topics` directory, this has been replaced by the `docs/concepts` directory.

This commit fixes the remaining references to the obsolete directory.


tvpartytonight:
@johnmccabe thanks for the fix!

PR #1336, opened 2017-02-07
(BKR-1054) Retry when RSA keygen fails

tvpartytonight:
There is some sort of race condition that occurs when trying to
generate RSA keys during testing with ruby 2.2.5 and OpenSSL. This
change adds a retry to the key process generation.


tvpartytonight:
@kevpl how do you feel about PRs that change product code to fix flaky spec tests? 🙈

tvpartytonight:
@ferglor you should also take a look here, so we can get the test machine rolling green again!

ferglor:
👍 lgtm

PR #1338, opened 2017-02-14
(BKR-931) Ensure that :type = 'aio' works in EL

petems:
EL and Fedora systems were not installing the pc1 version of Puppet
when :type = aio. This ensures that the case is properly addressed.

BKR-931 #close

Rebase of @trevor-vaughan's work with unit tests.


trevor-vaughan:
@petems You can update my PR directly if you like. It would be appreciated.

petems:
Closed in favour of original PR #1241

PR #1339, opened 2017-02-18
(maint) stub out wait_for_host_in_dashboard

jonathannewman:
wait_for_host_in_dashboard is no longer needed with PE and is causing
ihssues with attempts to turn off node-checkin by default in classifier.

This removes the check that validates a node is seen by the classifier.

It shouldn't be needed anyway as a puppet run has already been done by
the time this is run.


tvpartytonight:

tvpartytonight:
@kevpl I think I'd rather this be fixed in beaker-pe [here](https://github.com/puppetlabs/beaker-pe/blob/master/lib/beaker-pe/install/pe_utils.rb#L501), and probably just remove the entire `install_hosts.each` block. What do you think?

kevpl:
@tvpartytonight agreed. I'm closing this PR since that's the case, & I'll be submitting that PR shortly there.

PR #1342, opened 2017-02-22
(BKR-1059) beaker subcommands should return non-zero exit codes

mchllweeks:
…rroring

### DIFF
--- a/acceptance/tests/subcommands/init.rb
+++ b/acceptance/tests/subcommands/init.rb
@@ -4,24 +4,20 @@ test_name 'use the init subcommand' do
     on default, 'rm -rf /root/*'
   end
 
-  step 'ensure that `beaker init` fails correctly when not provided a hypervisor' do
-    expect_failure('it should return a non-zero code when it fails') do
-      result = on(default, 'beaker init', :accept_all_exit_codes => true)
-      refute_equal(0, result.exit_code, '`beaker init` without a hypervisor argument should return a non-zero exit code')
-    end
+  step 'ensure that `beaker init` exit value should be 1 when not provided with a supported hypervisor' do
+    result = on(default, 'beaker init ec2', :accept_all_exit_codes => true)
+    refute_equal(0, result.exit_code, '`beaker init` without a hypervisor argument should return a non-zero exit code')
   end
 
   step 'ensure that `beaker help init` works' do
     result = on(default, 'beaker help init')
-    assert_equal(0, result.exit_code, '`beaker help init` should return a zero exit code')
+    assert_match(/Usage+/, result.stdout)
   end
 
-  step 'ensure that `beaker init` accepts both vmpooler and vagrant hypervisor arguments' do
-
-    ['vmpooler', 'vagrant'].each do |hypervisor|
+  step 'ensure that `beaker init` accepts no argument as well as accepts either vmpooler or vagrant hypervisor arguments' do
+    ['vmpooler', 'vagrant', ''].each do |hypervisor|
       result = on(default, "beaker init #{hypervisor}")
       assert_match(/Writing host config.+/, result.stdout)
-      assert_equal(0, result.exit_code, "`beaker init #{hypervisor}` should return a zero exit code")
       step 'ensure that the Rakefile is present' do
         on(default, '[ -e "Rakefile" ]')
       end
--- a/lib/beaker/subcommands/subcommand_util.rb
+++ b/lib/beaker/subcommands/subcommand_util.rb
@@ -86,11 +86,11 @@ module Beaker
         execute_rake_task("beaker_quickstart:gen_hosts[vmpooler]")
       end
 
-      # Print a message to the console and exit with 0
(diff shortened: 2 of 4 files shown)


lib/beaker/subcommands/subcommand_util.rb
```diff
@@ -86,11 +86,11 @@ def self.init_vmpooler()
         execute_rake_task("beaker_quickstart:gen_hosts[vmpooler]")
       end
 
-      # Print a message to the console and exit with 0
+      # Print a message to the console and exit with 1
       # @param [String] msg the message to print
       def self.exit_with(msg)
```
nwolfe: Should we make the exit code a parameter of this method?  And have it default to `1`?

I haven't looked at the usages of this method, or the JIRA ticket, but just looking at this method it sounds like it wants to be a general-purpose method for printing and exiting, and we've already got the message parameterized.

mchllweeks:
retest this please

lib/beaker/subcommands/subcommand_util.rb
mchllweeks: fixed.

lib/beaker/subcommands/subcommand_util.rb
```diff
@@ -86,11 +86,11 @@ def self.init_vmpooler()
         execute_rake_task("beaker_quickstart:gen_hosts[vmpooler]")
       end
 
-      # Print a message to the console and exit with 0
+      # Print a message to the console and exit with 1
       # @param [String] msg the message to print
```
nwolfe: Might as well add a docstring for the new `exit_code` parameter too.

mchllweeks:
retest this please

lib/beaker/subcommands/subcommand_util.rb
tvpartytonight: Along with the new docstring, is there a reason why this defaults to 1 instead of 0?

mchllweeks: @tvpartytonight So my thinking was that this method would only used when something errors which is why I defaulted it to 1.  I can change that.  You also mentioned that the name of the method should be changed.  If we have a default of 0 for the exit code, should the name change be something like 'exit_with_msg' or do you have a better suggestion?

mchllweeks: @nwolfe Forgot to fix the doc string. Thx for catching that.

nwolfe: If this method is intended for failures, then it could have an intention-revealing name and be something like `fail`, `fail_with`, `fail_beaker`, `fail_subcommand`, etc.  Having the exit code default to `1` would make sense in this case.

If this method is intended as a general way for subcommands to exit, then maybe something like `exit`, `exit_with`, `exit_beaker`, etc.  Since this doesn't assume failure, we'd probably want the exit code to default to `0`.

Either way I think both the message and exit code should be parameters.

kevpl:
@mchllweeks if you would edit the git comments to be of the form:
```
(BKR-XXX) text text text
```
that would make this conform to our standards. You can do this by running
```
git rebase -i
```
and either squashing the commits or just editing their commit messages in the interactive rebase flow.

mchllweeks:
fixed commit messages

tvpartytonight:
@mchllweeks this looks good; I think that we should set up an acceptance test for this too before merge though.

tvpartytonight:
@mchllweeks specifically, I think [this](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/subcommand.rb#L24) should be changed to `error_with`, which should require a change to the init acceptance test [here](https://github.com/puppetlabs/beaker/blob/master/acceptance/tests/subcommands/init.rb#L7).

mchllweeks:
@tvpartytonight Ah, thx.  I need to update my branch with those changes and will then make the fix to the method name and init acceptance test.

acceptance/tests/subcommands/init.rb
```diff
@@ -4,24 +4,23 @@ def delete_root_folder_contents
...
   end
 
-  step 'ensure that `beaker init` fails correctly when not provided a hypervisor' do
-    expect_failure('it should return a non-zero code when it fails') do
-      result = on(default, 'beaker init', :accept_all_exit_codes => true)
-      refute_equal(0, result.exit_code, '`beaker init` without a hypervisor argument should return a non-zero exit code')
+  step 'ensure that `beaker init` exit value should be 1 when not provided with a supported hypervisor' do
+    begin
+      on(default, 'beaker init ec2', :accept_all_exit_codes => true)
+    rescue SystemExit=>e
```
tvpartytonight: It seems like this was used for the spec test, but the code here should just be saving the `result` object and then testing the `exit_code`. I would suspect that this `SystemExit` rescue is not actually rescuing anything at all.

tvpartytonight:
Test failure is a transient; matrix reloading the tests was just fine.

PR #1344, opened 2017-02-24
(DO NOT MERGE) Winrm / winrm-fs support

Iristyle:
Reviving a really old PR for @DLuCJ to start digging in on.

Good luck!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,3 +19,5 @@ if ENV['BEAKER_PE_PR_REPO_URL']
   ref = ENV['BEAKER_PE_PR_COMMIT']
   gem lib, :git => "git@github.com:#{author}/#{lib}.git", :branch => ref
 end
+
+gem 'ruby-prof', :require => false
--- a/beaker.gemspec
+++ b/beaker.gemspec
@@ -53,8 +53,8 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'beaker-hiera', '~> 0.0'
   s.add_runtime_dependency 'beaker-hostgenerator'
 
-  s.add_runtime_dependency 'winrm', '~> 1.3.0'
-  s.add_runtime_dependency 'winrm-fs', '~> 0.1.0'
+  s.add_runtime_dependency 'winrm', '~> 2.1.2'
+  s.add_runtime_dependency 'winrm-fs', '~> 1.0.1'
 
   # Optional provisioner specific support
   s.add_runtime_dependency 'rbvmomi', '~> 1.9'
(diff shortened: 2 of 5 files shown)


Iristyle:
Heads up @kevpl / @tvpartytonight - this winrm-fs support works, but may not be organized quite how you would expect (and tests may need to be rethought).  I've also talked with @DLuCJ about a few things related to this:

* The bad perf of file copies under Winrm (12x slower than SCP)
  * Related, using `ruby-prof` to find hotspots in the gem that we might want to fix upstream
* Alternative to Winrm file copies (maybe Beaker automatically creates a Windows file share and we optionally use SMB instead to copy things) - this is likely out of scope
* Using the built-in support for msiexec to install the agent over HTTP instead of copying to coordinator then out to agents (which would be a separate ticket)

The third one might be where @DLuCJ looks first to make some improvements in the overall process of Beaker installing agents on WIndows, as it could be little code change and a really good return on investment.

kevpl:
I'm 👍 to this change. I like the way that it's organized, splitting the connection out at the host level.

If you guys want to play around with improving this before it comes into use, I'm down with that. But if that's not going to happen within a month, I think we should close this PR for not having any motion, and if reminders or something else is needed, we find a way to get those done elsewhere.

Iristyle:
@kevpl it's my understanding that @DLuCJ is currently working on this...

kevpl:
🆒 🆒 @Iristyle, thanks for letting me know. We'll wait a few more weeks then before coming back to this.

DLuCJ:
@Iristyle @kevpl Yup, looking to put up some initial PRs for msiexec / winrm-fs this week.  Sorry for the delay.

Iristyle:
I've been using a variation of the following Ruby script to profile the WinRM transfers:

```ruby
require 'winrm-fs'
require 'ruby-prof'

hostname = ENV['WIN_TEST_HOST']
opts = {
  endpoint: "http://#{hostname}:5985/wsman",
  user: 'Administrator',
  password: 'password',
  disable_sppi: true
}
conn = WinRM::Connection.new(opts)
conn.shell(:powershell) do |shell|
  output = shell.run('$PSVersionTable') do |stdout, stderr|
    STDOUT.print stdout
    STDERR.print stderr
  end
  puts "The script exited with exit code #{output.exitcode}"
end

file_manager = WinRM::FS::FileManager.new(conn)

RubyProf.start
file_manager.upload('/Users/Iristyle/Downloads/some_file.pdf', 'C:/')

result = RubyProf.stop

printer = RubyProf::MultiPrinter.new(result)
printer.print(:path => './rubyprof')
````

kevpl:
@DLuCJ any news on this? No rush, I'd just like to know if this is something that's coming soon, or if maybe this should go on the backburner for now?

DLuCJ:
@kevpl I just put up PRs in beaker/beaker-pe for agent installation over HTTP.  They need some cleaning up, but worked as of three weeks ago.   

Generally, I think this should go on the backburner. 
I have some ideas for the first point, but don't think I will manage to get around to it soon.

PR #1345, opened 2017-02-24
(MAINT) fix broken Lets Write a Test doc

kevpl:


kevpl:
@joel-m-allen would you mind taking a look at this re-write of that doc & let me know what you think?

joel-m-allen:
I walked through the new doc on this pull request.  I created the test as specified and it passed. Good news, that.
This pull request is also the first I've been mentioned on, so I got to fumble around with the GIT flow a bit.  After a while I managed to find the "view" button and see the whole file.   One gotcha that I noticed is that if I try to browse your full clone, the links point me back to the master if I take them.  So, experience++.
Questions on this:  is there something I can do to visually verify the cowsay output?

joel-m-allen:
hmm.  my comment about the links leading back to the master doesn't seem accurate any more.  Maybe I'm just getting better at navigating the pull request.

kevpl:
@joel-m-allen you could use `result.stdout` to get at the text & verify what you need.

So as a for instance, another part of the test that could be added would be:
```ruby
assert(result.stdout =~ /pants pants pants/)
```

and you could assert against any part of the ascii art that you'd like to, such as verifying the eyes are still done with the same characters on a particular line.

kevpl:

kevpl:

kevpl:

PR #1348, opened 2017-03-01
(BKR-1066) Quote PATH to allow for spaces

MikaelSmith:
When using `$PATH`, quote it to accomodate spaces in file paths. This is
common on Windows systems, and causes the command to fail.

### DIFF
--- a/lib/beaker/host_prebuilt_steps.rb
+++ b/lib/beaker/host_prebuilt_steps.rb
@@ -173,7 +173,7 @@ module Beaker
         elsif host['platform'] =~ /aix/
           host.exec(Command.new(ROOT_KEYS_SYNC_CMD_AIX % "env PATH=/usr/gnu/bin:$PATH bash"), :accept_all_exit_codes => true)
         else
-          host.exec(Command.new(ROOT_KEYS_SYNC_CMD % "env PATH=/usr/gnu/bin:$PATH bash"), :accept_all_exit_codes => true)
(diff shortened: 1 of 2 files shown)


lib/beaker/host_prebuilt_steps.rb
```diff
@@ -173,7 +173,7 @@ def sync_root_keys host, opts
         elsif host['platform'] =~ /aix/
           host.exec(Command.new(ROOT_KEYS_SYNC_CMD_AIX % "env PATH=/usr/gnu/bin:$PATH bash"), :accept_all_exit_codes => true)
         else
-          host.exec(Command.new(ROOT_KEYS_SYNC_CMD % "env PATH=/usr/gnu/bin:$PATH bash"), :accept_all_exit_codes => true)
+          host.exec(Command.new(ROOT_KEYS_SYNC_CMD % "env PATH=\"/usr/gnu/bin:$PATH\" bash"), :accept_all_exit_codes => true)
```
joshcooper: I'm surprised that this works, because I thought you had to specify cygwin-ized paths, e.g. env PATH=/usr/gnu/path:/cygdrive/c/Windows/..." and cygwin translates them to Windows equivalents before calling `CreateProcess`. For example, in the AIO defaults: https://github.com/puppetlabs/beaker/blob/7439f35a4e24fc910502f32a2d3bbd9f98f07278/lib/beaker/dsl/install_utils/aio_defaults.rb#L19

MikaelSmith: I was running into PATH containing `/cygdrive/c/Program Files/Puppet/bin`.

PR #1349, opened 2017-03-09
(maint) Added a how to doc on running an SSH agent for acceptance tests

ferglor:


docs/how_to/enabling_access_to_vmpooler.md
```diff
@@ -0,0 +1,18 @@
...
+
+~~~bash
+eval `ssh-agent`
+ssh-add ~/.ssh/id_rsa-acceptance
+~~~
+  
+Now run acceptance tests that use the vmpooler:
+
+~~~bash
+bundle exec rake test:subcommands
```
nwolfe: I think it might be better to show the Beaker command here, rather than hide it behind "rake".

One question I'm wondering is whether Beaker's `--keyfile` argument is necessary in this case.

docs/how_to/enabling_access_to_vmpooler.md
```diff
@@ -0,0 +1,18 @@
+# Enabling access to the vmpooler VMs during an acceptance test run
+
+If you are running acceptance tests for Beaker that utilise vmpooler, you will need to run an SSH agent and add the acceptance RSA key prior to running the tests
+
+To load the SSH agent and add the acceptance key, run the following:
+
+~~~bash
+eval `ssh-agent`
+ssh-add ~/.ssh/id_rsa-acceptance
```
nwolfe: I noticed that some usages in ci-job-configs also add the `id_rsa` key.

I'm wondering if both are necessary in some cases?

Is the `id_rsa` used for cloning private repos from GitHub, and the `id_rsa-acceptance` used for SSH'ing into VMPooler VMs?

docs/how_to/enabling_access_to_vmpooler.md
```diff
@@ -0,0 +1,18 @@
+# Enabling access to the vmpooler VMs during an acceptance test run
+
+If you are running acceptance tests for Beaker that utilise vmpooler, you will need to run an SSH agent and add the acceptance RSA key prior to running the tests
```
nwolfe: Missing period at end of sentence

docs/how_to/enabling_access_to_vmpooler.md
ferglor: So my understanding is that the id_rsa key is the private key everyone generates when they join Puppet, the public component of which gets added to puppetlabs-modules and puppetlabs-sshkeys to enable access to GitHub/puppetlabs machines - I think the Jenkins nodes have their own id_rsa keys for accessing repos on GitHub. And yep, I believe that the id_rsa-acceptance key provides access to the VMPooler VMs

docs/how_to/enabling_access_to_vmpooler.md
ferglor: I guess thought it was okay to just use the rake tasks for running Beaker since the focus for this doc is enabling the SSH agent for Beaker runs, rather than focusing on how to run Beaker itself. In this case, the .ssh/id_rsa-acceptance key is specified via the `acceptance/config/acceptance-options.rb` file, but you could also specify the SSH key via the --keyfile flag

ferglor:
[skip ci]

docs/how_to/enabling_access_to_vmpooler.md
kevpl: We do want to use rake tasks as shortcuts to all of the common testing paths in beaker, so they're easier to reproduce for developers. I do think the wording could use some work to make that clearer though. Maybe:

    A common example of this functionality for beaker developers is
    in testing subcommands. There, we setup multiple SUTs that need
    to communicate between themselves. To run our subcommand
    testing to verify that you have agent forwarding setup correctly,
    run this testing rake task:

docs/how_to/enabling_cross_sut_access.md
```diff
@@ -0,0 +1,23 @@
...
+
+~~~bash
+eval `ssh-agent`
+ssh-add <SSH key file path>
+~~~
+  
+Now run your acceptance tests:
+
+~~~bash
+bundle exec rake test:subcommands
```
kevpl: @ferglor [this conversation](https://github.com/puppetlabs/beaker/pull/1349#discussion_r105249915) is still relevant

kevpl: ugh, appears that the link only works if you already have the old change open. It's @nwolfe's first comment

kevpl:

kevpl:

PR #1350, opened 2017-03-09
(BKR-1071) Return fqdn as hostname

MikaelSmith:
The openstack hypervisor previously distinguished between fqdn and
hostname. However, nothing else in Beaker uses the fqdn, and beaker-pe
appears to use fqdn to identify the node. Report the fqdn as the host's
hostname.

### DIFF
--- a/lib/beaker/hypervisor/openstack.rb
+++ b/lib/beaker/hypervisor/openstack.rb
@@ -220,16 +220,15 @@ module Beaker
 
       @hosts.each do |host|
         ip = get_ip
-        host[:vmhostname] = ip.ip.gsub '.','-'
-        host[:vmfqdn] = host[:vmhostname] + '.rfc1918.puppetlabs.net'
+        host[:vmhostname] = ip.ip.gsub('.','-') + '.rfc1918.puppetlabs.net'
         host[:keyname] = key_name(host)
-        @logger.debug "Provisioning #{host.name} (#{host[:vmfqdn]})"
+        @logger.debug "Provisioning #{host.name} (#{host[:vmhostname]})"
         options = {
           :flavor_ref => flavor(host[:flavor]).id,
           :image_ref  => image(host[:image]).id,
           :nics       => [ {'net_id' => network(@options[:openstack_network]).id } ],
-          :name       => host[:vmfqdn],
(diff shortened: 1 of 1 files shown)


MikaelSmith:
@lucywyman I had to change this to get PE install via beaker-pe to work.

lib/beaker/hypervisor/openstack.rb
```diff
@@ -220,16 +220,15 @@ def provision
 
       @hosts.each do |host|
         ip = get_ip
-        host[:vmhostname] = ip.ip.gsub '.','-'
-        host[:vmfqdn] = host[:vmhostname] + '.rfc1918.puppetlabs.net'
```
tvpartytonight: I'm not sure if this is safe to remove, but it looks like `:vmfqdn` is only used within this openstack hypervisor, so as long as no one out in the wild has written any code that relies on this, we should be good to go. @lucywyman @johnduarte any thoughts on the possibility someone has written some code reliant on `:vmfqdn`?

lucywyman: @tvpartytonight  I'm not as in touch with the community as I'd like to be, but in speaking to the folks I know who use both puppet and openstack it sounds like 'no'. I'm comfortable removing it (especially since it doesn't look very community friendly to begin with).

johnduarte: I am not aware of any such usage reliant on `:vmfqdn`

kevpl: it looks like the last person to set this code to use the `:vmfqdn` was @spjmurray in [this change](https://github.com/puppetlabs/beaker/commit/f41fa414ac7f397b7f140db144b40048427aac6e).

@spjmurray, I notice that in your description, you talk about how your commit attaches both a hostname and fqdn to a host and that fqdn is needed for backwards compatibility. Is this openstack backwards compatibility?

spjmurray: @kevpl backwards compatible in the sense that the output was the same a Lucy's original. By all means kill the fqdn! When I run it my adresses are not RFC1918 and definitely not part of puppetlabs.net :-)

PR #1351, opened 2017-03-10
(MAINT) add conditional archive recommendation

kevpl:
[skip ci]


kevpl:
I think the comments in [QENG-4834](tickets.puppetlabs.com/browse/QENG-4834) (internal-only link) align with the doc here, so this should be merge-able now.

@sbeaulie, @puppetlabs/beaker

ferglor:
👍

PR #1329, opened 2017-01-26
(BKR-1049) Updating require to align with oldest supported PE version

cdenneen:
PE and puppet-agent currently shipped with 2.1.9
Oldest supported PE release ships with 2.1.8
https://puppet.com/misc/puppet-enterprise-lifecycle
This is to keep beaker aligned with oldest supported PE release and
current puppet-agent vendored ruby.

PR #1352, opened 2017-03-10
(BKR-1049) update ruby version doc

kevpl:
[skip ci]

This PR documents https://github.com/puppetlabs/beaker/pull/1329,
and should be merged at the same time as that one is, so that the
change is described as soon as it goes in.

@puppetlabs/beaker, @tvpartytonight what do you think?
I originally thought this requirement would be mentioned in other
places, but I'm having trouble finding any. Am I missing some?

### DIFF
--- a/docs/tutorials/installation.md
+++ b/docs/tutorials/installation.md
@@ -1,4 +1,4 @@
-#Beaker Installation
+# Beaker Installation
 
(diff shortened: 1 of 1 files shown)


ferglor:
👍

kevpl:
Thanks to web editing, I've added review changes per our discussion earlier today, but they had to come on as their own commits.

@tvpartytonight please take another look, & note that this will need to be _squashed when merged_.

PR #1353, opened 2017-03-13
(BKR-1035) add old env var behavior

kevpl:
Original implementation of BKR-1035 uses just
the capitalization given to set environment
variables. This change sets the given env var
as well as the capitalized version, so that
we're backward compatible until the next major
release

/cc @joshcooper, @puppetlabs/beaker


kevpl:
unrelated. @puppetlabs-jenkins retest this please

ferglor:
👍

PR #1355, opened 2017-03-15
(BKR-1004) beaker exec implementation

tvpartytonight:
These changes introduce new subcommand functionality for beaker; the
exec subcommand takes a file, directory, or beaker suite name and
launches beaker to test that resource specifically. If no resource is
specified, it reads in default configuration that should already be set
up by previous subcommands.

### DIFF
--- a/lib/beaker/cli.rb
+++ b/lib/beaker/cli.rb
@@ -10,6 +10,8 @@ module Beaker
     |   |   | "
 
     attr_reader :options, :logger
+
+
     def initialize
       @timestamp = Time.now
       # Initialize a logger object prior to parsing; this should be overwritten whence
@@ -81,6 +83,17 @@ module Beaker
       rescue => e
         report_and_raise(@logger, e, "CLI.provision")
       end
+      self
+    end
+
+    #Initialize the network manager so it can initialize hosts for testing for subcommands
+    def initialize_network_manager
+      begin
+        @network_manager = Beaker::NetworkManager.new(@options, @logger)
+        @hosts = @network_manager.provision
+      rescue => e
+        report_and_raise(@logger, e, "CLI.initialize_network_manager")
+      end
     end
 
     #Run Beaker tests.
@@ -102,8 +115,6 @@ module Beaker
           @logger.warn "Interrupt received; exiting..."
           exit(1)
         end
-
-        provision
 
         # Setup perf monitoring if needed
         if @options[:collect_perf_data].to_s =~ /(aggressive)|(normal)/
--- a/lib/beaker/subcommand.rb
+++ b/lib/beaker/subcommand.rb
@@ -53,8 +53,11 @@ module Beaker
     class_option :'exclude-tags', :type => :string, :group => 'Beaker run'
     class_option :'xml-time-order', :type => :boolean, :group => 'Beaker run'
 
+    #global class options that should be available for all beaker subcommands
+    class_option :help, :type => :boolean
+
     desc "init BEAKER_RUN_OPTIONS", "Initializes the required configuration for Beaker subcommand execution"
-    option :help, :type => :boolean, :hide => true
+    option :help, :type => :boolean, :hide => true, :banner => 'whoooo'
     long_desc <<-LONGDESC
       Initializes the required .beaker configuration folder. This folder contains
       a subcommand_options.yaml file that is user-facing; altering this file will
@@ -92,7 +95,6 @@ module Beaker
     end
 
     desc "provision", "Provisions the beaker systems under test(SUTs)"
-    option :help, :type => :boolean, :hide => true
     long_desc <<-LONGDESC
     Provisions hosts defined in your subcommand_options file. You can pass the --hosts
     flag here to override any hosts provided there. Really, you can pass most any beaker
@@ -114,6 +116,13 @@ module Beaker
       # Sanitize the hosts
       cleaned_hosts = SubcommandUtil.sanitize_options_for_save(@cli.combined_instance_and_options_hosts)
 
+      # Update each host provisioned with a flag indicating that it no longer needs
+      # provisioning
+      cleaned_hosts.each do |host, host_hash|
+        host_hash['provision'] = false
+      end
+
+
       # should we only update the options here with the new host? Or update the settings
       # with whatever new flags may have been provided with provision?
       options_storage = YAML::Store.new(SubcommandUtil::SUBCOMMAND_OPTIONS)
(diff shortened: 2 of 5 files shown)


lib/beaker/subcommand.rb
```diff
@@ -61,5 +63,51 @@ def provision()
...
+    desc 'exec', 'execute a directory, file, or beaker suite'
+    long_desc <<-LONG_DESC
+    Run a single file, directory, or beaker suite. If supplied a file or directory,
+    that resource will be run in the context of the `tests` suite; If supplied a beaker
+    suite, then just that suite will run. If no resource is supplied, then this command
+    executes the suites as they are defined in the configuration.
+    LONG_DESC
+    def exec(resource=nil)
+      if options[:help]
+        invoke :help, [], ["exec"]
```
ferglor: So in order for this to work, you need to add `option :help, :type => :boolean, :hide => true` to the subcommand metadata. Having the --help flags hidden isn't really a necessity, I think it helps to keep the help text focused on what the subcommand actually does.

lib/beaker/subcommand.rb
```diff
@@ -61,5 +63,51 @@ def provision()
...
+        return
+      end
+
+      beaker_suites = [:pre_suite, :tests, :post_suite, :pre_cleanup]
```
ferglor: I would be inclined to move this after the `if` block below

lib/beaker/subcommand.rb
```diff
@@ -61,5 +63,51 @@ def provision()
...
+        # this to run in a custom suite. You know, in the future.
+        beaker_suites.each do |suite|
+          @@cli.options[suite] = []
+        end
+        if Pathname(resource).directory?
+          @@cli.options[:tests] = Dir.glob("#{Pathname(resource).expand_path}/*.rb")
+        else
+          @@cli.options[:tests] = Pathname(resource).expand_path.to_s
+        end
+      elsif resource.match(/pre-suite|tests|post-suite|pre-cleanup/)
```
ferglor: Could you add some spec tests around testing multiple suites?

lib/beaker/subcommand.rb
```diff
@@ -49,55 +53,134 @@ def initialize(*args)
...
+      alter the options subcommand execution.
 
-      Initialises a test environment configuration for a specific hypervisor
-
-      Supported hypervisors are: vagrant (default), vmpooler
-
+      Also modified in this operation is the .subcommand_state file that is used
+      by beaker to determine subcommand state. You as a user should never have to
+      edit this file.
     LONGDESC
```
ferglor: So I guess I don't like how this desc is telling the user what's going on under the hood, or how we're telling them about the .subcommand_state file, which they shouldn't really care about? I think a nicer desc would be what was there originally (I think it's important that we list out supported hypervisors at least), and then when we execute init, print out a message like "Written options to .beaker/subcommand_options.yaml". That way, we advertise the config location to the user, which they would probably be inclined to inspect themselves. I think we should hide any references to .subcommand_state as much as possible. I don't know if we should call these files subcommand_*, since I guess "subcommand" is too implementation-specific in that it relates to this particular phase of beaker development. I would be inclined to call them something like ".beaker_state" and "beaker_options", but I'm happy to go with whatever the majority decides 👍

lib/beaker/subcommands/subcommand_util.rb
```diff
@@ -39,6 +41,14 @@ def self.reset_argv(args)
         end
       end
 
+      def self.sanitize_options_for_save(options)
+        # God help us, the YAML library won't stop adding tags to objects, so this
```
ferglor: This line is hilarious, tony++

lib/beaker/subcommand.rb
ferglor: Given the new way we're passing options to the subcommands, is my previous comment still valid? I.e. is the --help option still necessary here, or will we be able to pass that flag as part of the global options that we pass the subcommands? If so, should we remove the hidden --help flags from other subcommands, and then parse CLI options before checking for the --help flag?

lib/beaker/subcommand.rb
```diff
@@ -49,55 +53,134 @@ def initialize(*args)
...
 
-      provisioned = @@config.transaction { @@config[:provisioned] }
 
-      if !provisioned or options[:force]
-        SubcommandUtil.provision(hypervisor, options)
-        SubcommandUtil.store_config({:provisioned => true})
+      # should we only update the options here with the new host? Or update the settings
+      # with whatever new flags may have been provided with provision?
+      options_storage = YAML::Store.new(SubcommandUtil::SUBCOMMAND_OPTIONS)
+      options_storage.transaction do
```
ferglor: I would be inclined to have access to our data stores wrapped in some update/delete util functions that accept a store location and key/value pairs to update?

lib/beaker/subcommand.rb
kevpl: @ferglor I agree, I don't like how it's telling the user what's going on under the hood either & we shouldn't tell them about the `.subcommand_state` file. Also, I think we should tell them about supported hypervisors, since those are needed in the 'new user' use case, which is when we want to be most verbose.

To your point about the file names @ferglor, I disagree. The folder these files live in is `.beaker`, & these files are subcommand-specific, so they should be named as such. If we leave them as general as `beaker`, then people will believe they should be usable outside of the subcommand workflow. "Subcommand" isn't implementation-specific either, it's the most general name for what we're doing & we could implement subcommands in any number of ways.

lib/beaker/subcommand.rb
tvpartytonight: We could pass `--help` as a `class_option` flag. That will allow all subcommands to process it and simply invoke `beaker help *subcommand*` and then return without actually executing that subcommands. I'm not particularly fond of how the help ends up printing the options; it ends up adding a  rather ugly usage that doesn't seem easy to reformat.

```
Options:
  [--help], [--no-help]

Beaker run options:
  -h, [--hosts=HOSTS]
...
```
I really hate that `[--no-help]` malarky, but, I don't see any real way of configuring the help print function to not do [that](https://github.com/erikhuda/thor/blob/master/lib/thor/parser/option.rb#L91). So...I guess I'm fine with help just print like that for now, unless we want to hide it, so that `--help` is an easter egg?

lib/beaker/subcommand.rb
tvpartytonight: I modified that in the previous PR to not expose any content about what's happening with that .subcommand_state.  I also added some more messaging about the subcommand_options.yaml file location and saving to it.

lib/beaker/subcommand.rb
tvpartytonight: I'm also into this idea, but I think we can revisit after we have established our pattern of exactly what we need from the accessor/getter function.

lib/beaker/subcommand.rb
```diff
@@ -53,8 +53,11 @@ def initialize(*args)
     class_option :'exclude-tags', :type => :string, :group => 'Beaker run'
     class_option :'xml-time-order', :type => :boolean, :group => 'Beaker run'
 
+    #global class options that should be available for all beaker subcommands
```
tvpartytonight: @ferglor here is the global option for help that will be available for alal subcommands.

ferglor: 👍

lib/beaker/cli.rb
```diff
@@ -103,8 +116,6 @@ def execute!
           exit(1)
         end
 
-        provision
```
tvpartytonight: @ki I took the `provision` call out of execute, had it return self, and then just added it to the invocation in `bin/beaker`. I think this was a good thing, as it separates `provisioning` and `execution` more cleanly. With this method, however, we can't call things like `configure` and `validate` and thus shouldn't be able to accept those flags.

Alternatively, we could modify provision to change some of the `SUBCOMMAND_OPTIONS` and set it options there to not `validate` or `configure`. If done that way, we can keep `provision` in the `execute!` method.

@ferglor 's destroy might have an opinion, as it seems very intertwined what provision does...

kevpl: @tvpartytonight I like separating them out. If we have to do a little more work invalidating the `validate` & `configure` parameters for the `execute` subcommand, I think it's worth it. 👍

PR #1358, opened 2017-03-20
(BKR-1075) Remove parsing from CLI intialize

tvpartytonight:
Prior to this commit, a Beaker::CLI object would do all options parsing
during initialization; now that parsing is needed by subcommands, we
should pull the parsing from initialization and call it after the object
has been instantiated.

### DIFF
--- a/lib/beaker/cli.rb
+++ b/lib/beaker/cli.rb
@@ -9,8 +9,16 @@ module Beaker
     |   V   |
     |   |   | "
 
+    attr_reader :logger
     def initialize
       @timestamp = Time.now
+      # Initialize a logger object prior to parsing; this should be overwritten whence
+      # the options are parsed and replaced with a new logger based on what is passed
+      # in to configure the logger.
+      @logger = Beaker::Logger.new
+    end
+
+    def parse_options
       @options_parser = Beaker::Options::Parser.new
       @options = @options_parser.parse_args
       @attribution = @options_parser.attribution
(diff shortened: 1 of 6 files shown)


lib/beaker/cli.rb
```diff
@@ -53,6 +47,16 @@ def initialize
         require File.expand_path(helper)
       end
 
+      if @options[:parse_only]
+        @logger.info("Beaker!")
+        @logger.info(VERSION_STRING % @options[:beaker_version])
+        @logger.info(@options.dump)
```
ferglor: I would move lines 51-53 and 88-90 into a new function

tvpartytonight:
Comments addressed, and spec and acceptance also updated to get these tests passing. The beaker-rspec tests also look good to me; @kevpl and @ferglor you mind taking a second look?

lib/beaker/cli.rb
```diff
@@ -9,8 +9,16 @@ class CLI
     |   V   |
     |   |   | "
 
+    attr_reader :logger
     def initialize
       @timestamp = Time.now
+      # Initialize a logger object prior to parsing; this should be overwritten whence
```
ferglor: whence++

spec/beaker/cli_spec.rb
```diff
@@ -2,9 +2,52 @@
...
+
+      describe 'instance variable initialization' do
+        it 'creates a logger for use before parse is called' do
+          expect(Beaker::Logger).to receive(:new).once.and_call_original
+          expect(cli.logger).to be_instance_of(Beaker::Logger)
+        end
+
+        it 'generates the timestamp' do
+          expect(Time).to receive(:now).once
+          cli
```
ferglor: Is this line unfinished?

spec/beaker/cli_spec.rb
```diff
@@ -2,9 +2,52 @@
...
+      end
+
+      describe '#parse_options' do
+        it 'returns self' do
+          expect(cli.parse_options).to be_instance_of(Beaker::CLI)
+        end
+
+        it 'replaces the logger object with a new one' do
+          expect(Beaker::Logger).to receive(:new).with(no_args).once.and_call_original
+          expect(Beaker::Logger).to receive(:new).once.and_call_original
```
ferglor: Is this line needed, given the line preceeding it?

spec/beaker/cli_spec.rb
```diff
@@ -2,9 +2,52 @@
...
+          cli.parse_options
+        end
+      end
+
+      describe '#print_version_and_options' do
```
ferglor: Should we also test for the absence of the version and options if --help, or absence of the options of --version is specified?

spec/beaker/cli_spec.rb
tvpartytonight: Yeah; it's testing to make sure that it receives new twice.

spec/beaker/cli_spec.rb
tvpartytonight: I think this could be addressed by adding a comment to the method that indicates that it should only be executed after `parse_options`.

PR #1359, opened 2017-03-20
(maint) Modified acceptance-options.rb to specify a single acceptance key

ferglor:


ferglor:
@kevpl @tvpartytonight so I'm just going to close this PR if it's okay with you guys? Removing that entry from the :keys array doesn't seem to affect how SSH works on the VMs, i.e. from the tests I've carried out, it seems that we still need to eval ssh-agent and add the acceptance key to enable cross-VM access.

PR #1360, opened 2017-03-24
(maint) Update required rake version to avoid gem dependency error

lucywyman:
I did this locally and ran a few rake commands (namely test:spec, history:gen, and docs:clean) all of which ran and operated as expected. 

In pe-console-ui, I get the following error from running `bundle update beaker`:
```
Resolving dependencies.........
Bundler could not find compatible versions for gem "rake":
  In Gemfile:
    beaker (= 3.13.0) was resolved to 3.13.0, which depends on
      rake (~> 10.0)

    rototiller (~> 1.0) was resolved to 1.0.0, which depends on
      rake (~> 11.0)
```

Since updating rake doesn't (seem to) break anything, using the latest version here seems like the easiest fix, though I'm definitely open to other fixes.

### DIFF
--- a/beaker.gemspec
+++ b/beaker.gemspec
@@ -38,7 +38,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'net-scp', '~> 1.2'
   s.add_runtime_dependency 'inifile', '~> 3.0'
 
-  s.add_runtime_dependency 'rake', '~> 10.0'
+  s.add_runtime_dependency 'rake', '~> 11.0'
   s.add_runtime_dependency 'rsync', '~> 1.0.9'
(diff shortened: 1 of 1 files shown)


tvpartytonight:
@lucywyman sorry for the late reply; with some of the work happening for beaker subcommands, we will be removing this gem as a runtime dependency and moving it back to a development dependency. We expect that change to happen in the next beaker release.

tvpartytonight:
@lucywyman with the beaker 3.14.0 release, we've changed the rake dependency back to what it used to be as a development dependency. So now if you bundle install your Gemfile `--without development` you will not pull in beaker's rake dependency. I'm going to close this PR now, but please feel free to reopen or ping me if you'd still like it considered.

PR #1357, opened 2017-03-20
(BKR-1076) Refactor beaker init and provision

tvpartytonight:
This changes beaker init to focus on setting up a repository for
beaker subcommands, rather than focusing in on generating host files
themselves. This also adds the functionality to create files that
contain the configured options; that is, all options that are not
presets after beaker parses its OptionsHash.

PR #1361, opened 2017-03-27
(BKR-1061) Beaker destroy implementation

ferglor:


ferglor:
Adding spec and acceptance tests, but wanted to get some code up for discussion

lib/beaker/subcommands/subcommand_util.rb
```diff
@@ -167,7 +167,29 @@ def self.provision(hypervisor, options)
...
+        save_object(beaker.network_manager.hypervisors, "#{CONFIG_DIR}/hypervisors.yaml")
       end
+
+
+      def self.cleanup(hypervisor)
+        cli = Beaker::CLI.new.parse_options
+        cli.options[:hosts_file] = "#{CONFIG_DIR}/acceptance/config/default_#{hypervisor}_hosts.yaml"
+        network_manager = cli.create_network_manager
+        network_manager.hosts = load_object("#{CONFIG_DIR}/hosts.yaml")
+        network_manager.hypervisors = load_object("#{CONFIG_DIR}/hypervisors.yaml")
```
tvpartytonight: hmmm...I'm not sure what this `hypervisors.yaml` is, or this`hosts.yaml`...oh, but I now get why the network manager is made readable. Given the changes from #1357 I would think that you would just be able to pull in `.beaker/subcommand_options.yaml`, which should happen during parsing...

lib/beaker/cli.rb
```diff
@@ -69,11 +69,15 @@ def print_version_and_options
       @logger.info(@options.dump)
     end
 
+    def create_network_manager
```
ferglor: Yeah if I recall correctly, provision on the network manager requests new VMs, which means if I was to do that in destroy, we'd have the VMs obtained in `beaker provision`, then we'd request new VMs as part of `beaker destroy`, then destroy them right away, but the VMs obtained in `beaker provision` would still be allocated. So instead I just create a new network manager instance, then update the state to match whatever was configured during `beaker provision`, and call destroy so that the VMs obtained during `beaker provision` are destroyed

lib/beaker/cli.rb
```diff
@@ -71,12 +69,16 @@ def print_version_and_options
       @logger.info(@options.dump)
     end
 
+    # Create a network manager instance
+    def create_network_manager
```
tvpartytonight: There's a few things here:

- I'm guessing we shouldn't have both `initialize` and `create` for `$VERB_HEREnetwork_manager`. Seems confusing...maybe we should just leave the `@network_manager` in as an `:attr_reader` and expect subcommands to call them directly?
- With what is outlined in [BKR-1090](https://tickets.puppetlabs.com/browse/BKR-1090) we could feasibly run `@cli.initialize_network_manager` and then simply just run cleanup after that, without having to store additional files to disk

lib/beaker/subcommand.rb
```diff
@@ -53,18 +53,15 @@ def initialize(*args)
     class_option :'exclude-tags', :type => :string, :group => 'Beaker run'
     class_option :'xml-time-order', :type => :boolean, :group => 'Beaker run'
 
-    #global class options that should be available for all beaker subcommands
```
tvpartytonight: Is there a reason why this global class_option is removed and then added individually per subcommand?

lib/beaker/subcommands/subcommand_util.rb
```diff
@@ -7,40 +8,21 @@ module Beaker
...
-    # - execute the quick start task for the specified hypervisor
     # - capture stdout and stderr
     module SubcommandUtil
-      BEAKER_REQUIRE = "require 'beaker/tasks/quick_start'"
-      HYPERVISORS = ["vagrant", "vmpooler"]
       CONFIG_DIR = ".beaker"
-      CONFIG_KEYS = [:hypervisor, :provisioned]
       SUBCOMMAND_OPTIONS = Pathname("#{CONFIG_DIR}/subcommand_options.yaml")
       SUBCOMMAND_STATE = Pathname("#{CONFIG_DIR}/.subcommand_state.yaml")
+      PERSISTED_HOSTS = Pathname("#{CONFIG_DIR}/.hosts.yaml")
```
tvpartytonight: We should delete these persisted values.

lib/beaker/subcommand.rb
```diff
@@ -53,18 +53,15 @@ def initialize(*args)
...
-    class_option :help, :type => :boolean
-
     desc "init BEAKER_RUN_OPTIONS", "Initializes the required configuration for Beaker subcommand execution"
-    option :help, :type => :boolean, :hide => true, :banner => 'whoooo'
```
tvpartytonight: :1st_place_medal:

lib/beaker/cli.rb
```diff
@@ -9,9 +9,7 @@ class CLI
     |   V   |
     |   |   | "
 
-    attr_reader :options, :logger
-
-
+    attr_reader :logger, :options, :network_manager
```
tvpartytonight: I'm also pretty sure that we no longer need to have this as a readable class var.

lib/beaker/hypervisor.rb
```diff
@@ -74,7 +75,7 @@ def self.create(type, hosts_to_provision, options)
         end
 
       hypervisor = hyper_class.new(hosts_to_provision, options)
-      hypervisor.provision
+      hypervisor.provision if options[:provision]
```
wallrj: I think this change caused https://tickets.puppetlabs.com/browse/BKR-1111

PR #1363, opened 2017-03-31
(MODULES-4594) vmpooler hypervisor: add disks

eputnam:
In order to set up a test environment for Red Hat Satellite, we need a
disk that's larger than the default vmpooler allotment (about 5 GB free
vs the 1.5GB that are free on a redhat host after all is said and done).
This adds a step to the provision function in the vmpooler hypervisor to
add a disk of x size to any host in the host configuration file. It is
then up to the tests to partition, mount, expand, or whatever they
decide to do with it.

### DIFF
--- a/lib/beaker/hypervisor/vmpooler.rb
+++ b/lib/beaker/hypervisor/vmpooler.rb
@@ -215,6 +215,33 @@ module Beaker
       end
 
       @logger.notify 'Spent %.2f seconds tagging VMs' % (Time.now - start)
+
+      # add additional disks to vm
+      @logger.debug 'Looking for disks to add...'
+
+      @hosts.each do |h|
+        hostname = h['vmhostname'].split(".")[0]
+
+        if h['disks']
+          @logger.debug "Found disks for #{hostname}!"
+          disks = h['disks']
+
+          disks.each_with_index do |disk_size, index|
+            start = Time.now
+
+            add_disk(hostname, disk_size)
+
+            done = wait_for_disk(hostname, disk_size, index)
+            if done
+              @logger.notify "Spent %.2f seconds adding disk #{index}. " % (Time.now - start)
+            else
+              raise "Could not verify disk was added after %.2f seconds" % (Time.now - start)
+            end
+          end
+        else
+          @logger.debug "No disks to add for #{hostname}"
+        end
+      end
     end
 
     def cleanup
@@ -246,5 +273,82 @@ module Beaker
       @logger.notify "Spent %.2f seconds cleaning up" % (Time.now - start)
     end
 
+    def add_disk(hostname, disk_size)
+      @logger.notify "Requesting an additional disk of size #{disk_size}GB for #{hostname}"
+
+      if !disk_size.to_s.match /[0123456789]/ || size <= '0'
+        raise NameError.new "Disk size must be an integer greater than zero!"
+      end
+
+      begin
+        uri = URI.parse(@options[:pooling_api] + '/api/v1/vm/' + hostname + '/disk/' + disk_size.to_s)
+
+        http = Net::HTTP.new(uri.host, uri.port)
+        request = Net::HTTP::Post.new(uri.request_uri)
+        request['X-AUTH-TOKEN'] = @credentials[:vmpooler_token]
+
+        response = http.request(request)
+
+        parsed = parse_response(response)
+
+        raise "Response from #{hostname} indicates disk was not added" if !parsed['ok']
+
+      rescue NameError, RuntimeError, Errno::EINVAL, Errno::ECONNRESET, EOFError,
+             Net::HTTPBadResponse, Net::HTTPHeaderSyntaxError, *SSH_EXCEPTIONS => e
+        report_and_raise(@logger, e, 'Vmpooler.add_disk')
+      end
+    end
+
+    def parse_response(response)
+      parsed_response = JSON.parse(response.body)
+    end
+
+    def disk_added?(host, disk_size, index)
+      if host['disk'].nil?
+        false
+      else
+        host['disk'][index] == "+#{disk_size}gb"
+      end
+    end
+
+    def get_vm(hostname)
+      begin
+        uri = URI.parse(@options[:pooling_api] + '/vm/' + hostname)
+
+        http = Net::HTTP.new(uri.host, uri.port)
+        request = Net::HTTP::Get.new(uri.request_uri)
+
+        response = http.request(request)
+      rescue RuntimeError, Errno::EINVAL, Errno::ECONNRESET, EOFError,
+             Net::HTTPBadResponse, Net::HTTPHeaderSyntaxError, *SSH_EXCEPTIONS => e
+        @logger.notify "Failed to connect to vmpooler while getting VM information!"
+      end
+    end
+
+    def wait_for_disk(hostname, disk_size, index)
+      response = get_vm(hostname)
+      parsed = parse_response(response)
+
+      @logger.notify "Waiting for disk"
+
+      attempts = 0
+
+      while (!disk_added?(parsed[hostname], disk_size, index) && attempts < 20)
+        sleep 10
+        begin
+          response = get_vm(hostname)
+          parsed = parse_response(response)
+        rescue RuntimeError, Errno::EINVAL, Errno::ECONNRESET, EOFError,
+               Net::HTTPBadResponse, Net::HTTPHeaderSyntaxError, *SSH_EXCEPTIONS => e
+          report_and_raise(@logger, e, "Vmpooler.wait_for_disk")
+        end
+        print "."
+        attempts += 1
+      end
+
+      puts " "
+
+      disk_added?(parsed[hostname], disk_size, index)
+    end
   end
 end
--- a/docs/how_to/hypervisors/vmpooler.md
+++ b/docs/how_to/hypervisors/vmpooler.md
@@ -18,6 +18,23 @@ An example of a `.fog` file with just the vmpooler details is below:
 :default:
   :vmpooler_token: 'randomtokentext'
 ```
+# Additional Disks
+Using the vmpooler API, Beaker enables you to attach additional storage disks in the host configuration file. The disks are added at the time the VM is created. Logic for using the disk must go into your tests.
+
+Simply add the `disks` key and a list containing the sizes(in GB) of the disks you want to create and attach to that host.
+For example, to create 2 disks sized 8GB and 16GB to example-box:
+
+```yaml
+ example-box:
+    disks:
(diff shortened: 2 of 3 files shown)


lib/beaker/hypervisor/vmpooler.rb
```diff
@@ -246,5 +277,73 @@ def cleanup
...
+
+    def wait_for_disk(hostname)
+
+      response = get_vm(hostname)
+      parsed = parse_response(response)
+
+      @logger.notify "Waiting for disk"
+
+      attempts = 0
+      while (!disk_added?(parsed[hostname]) && attempts < 15)
```
glennsarti: This assumes only one disk could ever be added.  is this a true assumption to make?.  If so, may want to document that beaker can only add a single extra disk for VMPooler

eputnam: more than one can be added as it turns out. point taken.

glennsarti: Only one disk is still an improvement over the current zero.  Just need to document it.

eputnam:
re-ran Jenkins failure (1485) and it passed ([1489](https://jenkins-beaker.delivery.puppetlabs.net/job/qe_beaker_regression_pe-latest-release/agent=sles12,master=centos7/1489/))

lib/beaker/hypervisor/vmpooler.rb
```diff
@@ -246,5 +286,79 @@ def cleanup
...
+
+        http = Net::HTTP.new(uri.host, uri.port)
+        request = Net::HTTP::Post.new(uri.request_uri)
+
+        request['X-AUTH-TOKEN'] = @credentials[:vmpooler_token]
+
+        response = http.request(request)
+      rescue RuntimeError, Errno::EINVAL, Errno::ECONNRESET, EOFError,
+             Net::HTTPBadResponse, Net::HTTPHeaderSyntaxError, *SSH_EXCEPTIONS => e
+        @logger.notify "Failed to connect to vmpooler for additional disk!"
```
tvpartytonight: @eputnam I was curious to know why you chose to have much of this code rescue errors and use the logger to notify that disk provisioning failed. Since failed disk provisioning will certainly cause test failures with tests that rely on it, why not fail earlier?

In code that is similar in this class—the tagging code—beaker doesn't fail out, but that is because failed tagging doesn't affect actual testing, just our metrics around the testing itself.

eputnam: yeah, good call. i shall adjust.

eputnam: I think the HTTP rescues are worth keeping to potentially provide info but i added some failures in the provision method.

lib/beaker/hypervisor/vmpooler.rb
```diff
@@ -215,6 +215,37 @@ def provision
...
+
+      @hosts.each do |h|
+        hostname = h['vmhostname'].split(".")[0]
+
+        if h['disks']
+          @logger.debug "Found disks for #{hostname}!"
+          disks = h['disks']
+          disks.map!.with_index { |disk, index| "#{disk}_#{index}" }
+
+          disks.each do |disk|
```
hunner: It seems like instead of doing a `.map!.with_index` and then a `.each` on that, you could do
```ruby
disks.each_with_index do |disk_size, index|
  #...
```
and it would be more obvious what's going on. Then:
```ruby
def disk_added?(host, disk_size, disk_index)
```
and
```ruby
def wait_for_disk(host, disk_size, disk_index)
   disk = "#{disk_size}_#{disk_index}"
```

lib/beaker/hypervisor/vmpooler.rb
```diff
@@ -246,5 +277,93 @@ def cleanup
...
+
+    def disk_added?(host, disk)
+      disk_size = disk.split('_')[0]
+      disk_index = disk.split('_')[1].to_i
+
+      if !host['disk'].nil?
+        return host['disk'][disk_index] == "+#{disk_size}gb"
+      end
+
+      return false
```
hunner: Explicit returns are usualyl discouraged.
```ruby
if host['disk'].nil?
  false
else
  host['disk'][disk_index] == "+#{disk_size}gb"
end

File: lib/beaker/hypervisor/vmpooler.rb [hunk 5]
```diff
@@ -246,5 +277,93 @@ def cleanup
...
+               Net::HTTPBadResponse, Net::HTTPHeaderSyntaxError, *SSH_EXCEPTIONS => e
+          report_and_raise(@logger, e, "Vmpooler.wait_for_disk")
+        end
+        print "."
+        attempts += 1
+      end
+
+      puts " "
+
+      return true if disk_added?(parsed[hostname], disk)
```
hunner: This can just be `disk_added?(parsed[hostname], disk)` as that returns true/false by itself.

eputnam:
@hunner revised according to requested changes

lib/beaker/hypervisor/vmpooler.rb
```diff
@@ -246,5 +275,91 @@ def cleanup
...
+          parsed = parse_response(response)
+        rescue RuntimeError, Errno::EINVAL, Errno::ECONNRESET, EOFError,
+               Net::HTTPBadResponse, Net::HTTPHeaderSyntaxError, *SSH_EXCEPTIONS => e
+          report_and_raise(@logger, e, "Vmpooler.wait_for_disk")
+        end
+        print "."
+        attempts += 1
+      end
+
+      puts " "
```
tvpartytonight: stray `puts`?

lib/beaker/hypervisor/vmpooler.rb
```diff
@@ -246,5 +275,91 @@ def cleanup
...
+    def wait_for_disk(hostname, disk_size, index)
+
+      response = get_vm(hostname)
+      parsed = parse_response(response)
+
+      @logger.debug "Waiting for disk"
+
+      attempts = 0
+
+      while (!disk_added?(parsed[hostname], disk_size, index) && attempts < 20)
```
tvpartytonight: Would this be better served by using the [fibonacci backoff](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/shared/repetition.rb#L14)?

eputnam: I tried using the fibonacci. it turned out that the process takes around 100 seconds, usually over 100 seconds, so the fibonacci would check at around 100 and then not again until 200 so it would end up taking almost twice as long as it needed to. it seemed like that was less than ideal.

lib/beaker/hypervisor/vmpooler.rb
eputnam: This is so there's a new line for the "Spent %.2f seconds adding disk #{disk.split('_')[1]}. " message. i'm glad you pointed this out, that disk.split is broken.

lib/beaker/hypervisor/vmpooler.rb
tvpartytonight: good to know @eputnam. I think that the other method there, the [repeat_for](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/shared/repetition.rb#L5) would then be appropriate. But, it's not absolutely necessary to use that method, just wanted to make sure you knew those methods were available to you.

lib/beaker/hypervisor/vmpooler.rb
```diff
@@ -215,6 +215,35 @@ def provision
...
+
+            done = wait_for_disk(hostname, disk_size, index)
+            if done
+              @logger.notify "Spent %.2f seconds adding disk #{index}. " % (Time.now - start)
+            else
+              @logger.notify 'Timeout, cannot verify disk was added.'
+            end
+          end
+        else
+          @logger.debug "No disks to add for #{hostname}"
```
tvpartytonight: Should this be a `notify` to match the above "Looking for disks to add"?

lib/beaker/hypervisor/vmpooler.rb
```diff
@@ -215,6 +215,35 @@ def provision
...
+            start = Time.now
+
+            response = add_disk(hostname, disk_size)
+            parsed = parse_response(response)
+
+            done = wait_for_disk(hostname, disk_size, index)
+            if done
+              @logger.notify "Spent %.2f seconds adding disk #{index}. " % (Time.now - start)
+            else
+              @logger.notify 'Timeout, cannot verify disk was added.'
```
tvpartytonight: Does this mean that beaker might not be able to verify that the disks were added, but still continue with the test run?

eputnam: yes. sometimes it just takes a while but maybe it should just fail here, too.

lib/beaker/hypervisor/vmpooler.rb
eputnam: probably

eputnam:
DO NOT MERGE YET. I'm finding new things.

eputnam:
alright, this is ready to go.

lib/beaker/hypervisor/vmpooler.rb
```diff
@@ -246,5 +273,82 @@ def cleanup
       @logger.notify "Spent %.2f seconds cleaning up" % (Time.now - start)
     end
 
+    def add_disk(hostname, disk_size)
+      @logger.notify "Requesting an additional disk of size #{disk_size}GB for #{hostname}"
```
kevpl: @eputnam would you put in a sentence to this effect in the doc for adding disks above? Just reading the docs, I'm not seeing what the numbers mean above. This line of code clarifies that, but we have a problem in beaker where people typically have to look at the code to figure out what's going on because our docs aren't clear enough.

eputnam: @kevpl , I tweaked the doc a little bit, let me know what you think.

kevpl: 👍 looks good!

eputnam:
i re-kicked the failing job. See for successful runs: 
https://jenkins-beaker.delivery.puppetlabs.net/job/qe_beaker_regression_pe-latest-release/1519/
https://jenkins-beaker.delivery.puppetlabs.net/job/qe_beaker_intn-sys_vpool-beaker-rspec-support/2304/

kevpl:
@eputnam the rspec-support job wasn't the one that was failing, the puppet-legacy job was the one that was. I'll just re-kick the whole thing now.

eputnam:

kevpl:
@eputnam hopefully we'll get #1377 resolved soon. Once that's done, we'll re-kick testing here.

kevpl:

PR #1364, opened 2017-04-01
(BKR-1091) vCloud datacenter selection support

ipcrm:
Prior to this commit you couldn't specify the datacenter to search for a
given folder.  The `find_datacenter` method will return the first found
datacenter, and in the case that you have linked vcenters this falls
apart.  This commit adds the ability for `datacenter` to be included in
@options.  Related, folder search now respects datacenter as well as
uses the traverse method.  Prior to this commit some folder searches
were not possible (for example a base level because it didn't contain a
`/` which was being split on).

Other minor updates; docs updated and template can be provided as ENV var.

### DIFF
--- a/lib/beaker/hypervisor/vcloud.rb
+++ b/lib/beaker/hypervisor/vcloud.rb
@@ -10,6 +10,7 @@ module Beaker
 
       raise 'You must specify a datastore for vCloud instances!' unless @options['datastore']
       raise 'You must specify a folder for vCloud instances!' unless @options['folder']
+      raise 'You must specify a datacenter for vCloud instances!' unless @options['datacenter']
       @vsphere_credentials = VsphereHelper.load_config(@options[:dot_fog])
     end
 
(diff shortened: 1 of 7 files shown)


lib/beaker/hypervisor/vcloud.rb
```diff
@@ -147,15 +154,31 @@ def provision
 
         try = (Time.now - start) / 5
         duration = run_and_report_duration do
-          @hosts.each_with_index do |h, i|
-            wait_for_dns_resolution(h, try, attempts)
+          @hosts.each do |host|
+            #wait_for_dns_resolution(h, try, attempts)
+            until @vsphere_helper.find_vms(host['vmhostname'])[host['vmhostname']].summary.guest.ipAddress != nil
```
ferglor: Is it possible that this `until` will never exit? Should we be taking a maximum number of attempts into account?

lib/beaker/hypervisor/vsphere_helper.rb
```diff
@@ -118,29 +118,23 @@ def find_vms names, connection = @connection
...
-        else
-          abort "Unexpected object type encountered (#{base.class}) while finding folder"
-      end
+  def find_folder(dc,foldername)
+    datacenter = @connection.serviceInstance.find_datacenter(dc)
+    base = datacenter.vmFolder.traverse(foldername)
+    if base != nil
+      base
+    else
+      abort "Unexpected object type encountered (#{base.class}) while finding folder"
```
ferglor: is it worth calling #{base.class} here if base will be nil at this point?

lib/beaker/hypervisor/vcloud.rb
ferglor: We have a `repeat_fibonacci_style_for` method available here: lib/beaker/shared/repetition.rb

ipcrm:
@ferglor @kevpl Revised based on reviews, thx!

PR #1365, opened 2017-04-03
(BKR-1092) Keystone V3 Support

spjmurray:
Primary change is to add in optional user and project domain names for authentication
against a Keystone V3 endpoint.  The secondary change is to tidy up how keystone
credentials are handled as there was a whole bunch of duplication.  If there is any
concern around the disappearing conditional protecting the region selection worry not
as Fog checks for nil before populating the authentication request body.


spjmurray:
And get this Kevin... I actually wrote spec tests without having to be told for once :)

kevpl:
👍 looks lgtm.

PR #1366, opened 2017-04-03
(BKR-1093) Fix Nova Keypair Generation

spjmurray:
Firstly this fixes the rspec test that doesn't work, shame on me.  Secondly it
ensures only the hostname part of the host's FQDN is used in keypair generation,
as the FQDN includes periods and is illegal according to Nova.


kevpl:

PR #1367, opened 2017-04-05
(BKR-1095) Apply work-around for Cisco Nexus to timesync

LuvCurves:


lib/beaker/host/cisco.rb
```diff
@@ -70,8 +70,12 @@ def package_config_dir
     # @return [String] Command string as needed for this host
     def prepend_commands(command = '', user_pc = '', opts = {})
       return user_pc unless command.index('vsh').nil?
+      if self[:platform] =~ /cisco_nexus/
```
tvpartytonight: @LuvCurves I'm surprised this conditional is necessary; shouldn't this be assumed to match the `cisco` case [here](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host.rb#L60)? Do we have multiple cisco platforms, some that match `/cisco/` but don't match `/cisco_nexus/`?

kevpl: @tvpartytonight yeah, I believe this is true, that there are platform strings like `cisco_exr` and such that would match the host object & not match `cisco_nexus`. But those have changed over time, so I'm not sure if that's gone away. @LuvCurves ?

lib/beaker/command.rb
```diff
@@ -73,12 +73,16 @@ def cmd_line host, cmd = @command, env = @environment, pc = @prepend_cmds
...
         cmd = cmd.gsub('"') { '\\"' }
       end
 
       # This will cause things like `puppet -t -v agent` which is maybe bad.
-      cmd_line_array = [env_string, prepend_commands, cmd, options_string, args_string, append_command]
+      if cmd =~ /ntpdate/ && host[:platform] =~ /cisco_nexus/
+        cmd_line_array = [env_string, prepend_commands, cmd, options_string, args_string]
+      else
+        cmd_line_array = [env_string, prepend_commands, cmd, options_string, args_string, append_command]
+      end
```
kevpl: @LuvCurves there would be less duplication if this was re-written like so:
```ruby
cmd_line_array = [env_string, prepend_commands, cmd, options_string, args_string]
cmd_line_array << append_command unless (cmd =~ /ntpdate/ && host[:platform] =~ /cisco_nexus/)
```
or breaking the appending command into a multiline conditional because the test is maybe not the simplest to read. Either way, I'm not the biggest fan of re-writing the array here.

LuvCurves: Changed

kevpl:
as long as testing proves good, then I'm 👍

PR #1372, opened 2017-04-07
(BKR-954) Add Ability to Disable Openstack Volume Management

ipcrm:
Prior to this commit beaker openstack support assumed that the
`volumes` service was provisioned.  This service is not always
available in an Openstack deployment.  This commit gives you the
option to disable that functionality.  If you do not disable it
on an instance that does not publish this functionality, even if
you don't configure the volumes setting in `CONFIG`, beaker runs
will fail.

### DIFF
--- a/docs/how_to/hypervisors/openstack.md
+++ b/docs/how_to/hypervisors/openstack.md
@@ -152,3 +152,15 @@ volume size in MB.
             size: 10000
           journal:
             size: 1000
+
+
+In the event your using an OpenStack instance that does not deploy the volume service you can disable that functionality to prevent beaker runs from failing.  Either using an ENV variable or setting the following value in the `CONFIG` section of your hosts file(valid values are `true` or `false`):
+
+```
+openstack_volume_support: false
+```
+
+You can also configure this setting via an environment variable:
+
+```
+export OS_VOL_SUPPORT=false
--- a/docs/how_to/hypervisors/openstack.md
+++ b/docs/how_to/hypervisors/openstack.md
@@ -154,7 +154,7 @@ volume size in MB.
             size: 1000
 
 
-In the event your using an OpenStack instance that does not deploy the volume service you can disable that functionality to prevent beaker runs from failing.  Either using an ENV variable or setting the following value in the `CONFIG` section of your hosts file(valid values are `true` or `false`):
+In the event you're using an OpenStack instance that does not deploy the volume service you can disable that functionality to prevent beaker runs from failing.  Either using an ENV variable or setting the following value in the `CONFIG` section of your hosts file(valid values are `true` or `false`):
 
 ```
 openstack_volume_support: false
@@ -163,4 +163,4 @@ openstack_volume_support: false
 You can also configure this setting via an environment variable:
 
 ```
-export OS_VOL_SUPPORT=false
+export OS_VOLUME_SUPPORT=false
--- a/lib/beaker/hypervisor/openstack.rb
+++ b/lib/beaker/hypervisor/openstack.rb
@@ -63,6 +63,17 @@ module Beaker
         raise "Unable to create OpenStack Network instance (api_key: #{@options[:openstack_api_key]}, username: #{@options[:openstack_username]}, auth_url: #{@options[:openstack_auth_url]}, tenant: #{@options[:openstack_tenant]})"
       end
 
+      # Validate openstack_volume_support setting value, reset to boolean if passed via ENV value string
+      if not @options[:openstack_volume_support].is_a?(TrueClass) || @options[:openstack_volume_support].is_a?(FalseClass)
+        if @options[:openstack_volume_support].match(/\btrue\b/i)
+          @options[:openstack_volume_support] = true
+        elsif @options[:openstack_volume_support].match(/\bfalse\b/i)
+          @options[:openstack_volume_support] = false
+        else
+          raise "Invalid value provided for CONFIG setting openstack_volume_support, current value #{@options[:openstack_volume_support]}"
+        end
+      end
+
     end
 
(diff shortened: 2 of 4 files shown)


lib/beaker/options/presets.rb
```diff
@@ -134,6 +134,7 @@ def presets
           :openstack_keyname      => ENV['OS_KEYNAME'],
           :openstack_network      => ENV['OS_NETWORK'],
           :openstack_region       => ENV['OS_REGION'],
+          :openstack_volume_support => ENV['OS_VOL_SUPPORT'] || true,
```
tvpartytonight: Should this be `OS_VOLUME_SUPPORT` to match the symbol key names?

docs/how_to/hypervisors/openstack.md
```diff
@@ -152,3 +152,15 @@ volume size in MB.
             size: 10000
           journal:
             size: 1000
+
+
+In the event your using an OpenStack instance that does not deploy the volume service you can disable that functionality to prevent beaker runs from failing.  Either using an ENV variable or setting the following value in the `CONFIG` section of your hosts file(valid values are `true` or `false`):
```
ferglor: This should be "In the event **you're** using..."

lib/beaker/hypervisor/openstack.rb
```diff
@@ -63,6 +63,17 @@ def initialize(openstack_hosts, options)
...
+      # Validate openstack_volume_support setting value, reset to boolean if passed via ENV value string
+      if not @options[:openstack_volume_support].is_a?(TrueClass) || @options[:openstack_volume_support].is_a?(FalseClass)
+        if @options[:openstack_volume_support].match(/\btrue\b/i)
+          @options[:openstack_volume_support] = true
+        elsif @options[:openstack_volume_support].match(/\bfalse\b/i)
+          @options[:openstack_volume_support] = false
+        else
+          raise "Invalid value provided for CONFIG setting openstack_volume_support, current value #{@options[:openstack_volume_support]}"
+        end
+      end
```
kevpl: @ipcrm this seems like an awful lot of work for what you're trying to get at. Is there some use case where a string or something like that must be passed in, so you can't just rely on this being of type `TrueClass` or `FalseClass`?

We would prefer to rely on people setting the yaml correctly and failing on the case when people have set the wrong type for these. That keeps the code here simple, and in places like hypervisors or parsers, it's not a lot of wasted time to fail since things are just getting started.

also, it seems to be the case that if someone passed in a `FalseClass` `false` value, this wouldn't set the `@options[:openstack_volume_support]` value correctly, would it? I would imagine the outer if statement would need an else case to take care of that, no?

ipcrm: @kevpl The reason for that block (67-75) is to handle inputs from the environment var.  Basically checking if the string is true/false and then converting to an actual boolean.  I was trying to find a way to just cast it as needed, but couldn't get it working and fell back to this.  Open to suggestions :)

ipcrm:
@tvpartytonight Updated variable name

ipcrm:
@ferglor Corrected spelling

ipcrm:
@kevpl Took a crack at simplifying that logic, please review.

lib/beaker/hypervisor/openstack.rb
```diff
@@ -63,6 +63,11 @@ def initialize(openstack_hosts, options)
         raise "Unable to create OpenStack Network instance (api_key: #{@options[:openstack_api_key]}, username: #{@options[:openstack_username]}, auth_url: #{@options[:openstack_auth_url]}, tenant: #{@options[:openstack_tenant]})"
       end
 
+      # Validate openstack_volume_support setting value, reset to boolean if passed via ENV value string
+      @options[:openstack_volume_support] = true  if @options[:openstack_volume_support].to_s.match(/\btrue\b/i)
+      @options[:openstack_volume_support] = false if @options[:openstack_volume_support].to_s.match(/\bfalse\b/i)
+      [true,false].include? @options[:openstack_volume_support] or raise "Invalid openstack_volume_support setting, current: @options[:openstack_volume_support]"
```
kevpl: @ipcrm what advantages does this approach have over a much simpler one like below?
```ruby
@options[:openstack_volume_support] = @options[:openstack_volume_support].to_s == 'true'
```

lib/beaker/hypervisor/openstack.rb
```diff
@@ -229,7 +234,6 @@ def provision
         vm = @compute_client.servers.create(options)
 
         #wait for the new instance to start up
-        start = Time.now
```
kevpl: @ipcrm was this just a mistake, or was there a specific reason for this change?

lib/beaker/hypervisor/openstack.rb
ipcrm: @kevpl My only concern with this is that if I provide an invalid configuration the proposed block will essentially auto-convert any incorrect value to false.  In this scenario the consumer of beaker will not have any idea that the setting they are using is incorrect (since there is no scenario where we raise an error).  I feel that explicitly failing instead of auto-munging the parameter makes more sense from a usability standpoint.

lib/beaker/hypervisor/openstack.rb
ipcrm: @kevpl That variable isn't used.

kevpl:
@ipcrm looks lgtm to me right now. Please prepend the beaker ticket reference to all commits in the history to match our CONTRIBUTING standard.

Once that's done, we can look at merging this in.

ipcrm:
@kevpl Commits updated.

kevpl:
@ipcrm thanks! looks like a transient failure, re-running now.

PR #1374, opened 2017-04-12
(BKR-222) docker support for systemd based OS

alwinmark:
This Fixes espacially CentOS7 and most probably RHEL7 as well, as any other OS which is using sshd threw systemctl instead of init.d.

What happens inside the container currently:
```
[root@91ddf1f362e8 ~]# service sshd status 
-bash: service: command not found
[root@91ddf1f362e8 ~]# systemctl status sshd
Failed to get D-Bus connection: Operation not permitted
```

testable with: `docker run --rm -it centos:7 /bin/bash`

and installing sshd with:

```
$yum install -y openssh-server openssh-clients
$sshd-keygen
[...]
```

[1] https://rhatdan.wordpress.com/2014/04/30/running-systemd-within-a-docker-container/

### DIFF
--- a/acceptance/tests/base/external_resources_test.rb
+++ b/acceptance/tests/base/external_resources_test.rb
@@ -1,9 +1,11 @@
 test_name 'External Resources Test' do
   step 'Verify EPEL resources are up and available' do
     def epel_url_test(el_version)
-      url = "#{options[:epel_url]}/epel-release-latest-#{el_version}.noarch.rpm"
+      url_base = options[:epel_url]
+      url_base = options[:epel_url_archive] if el_version == 5
+      url = "#{url_base}/epel-release-latest-#{el_version}.noarch.rpm"
       curl_headers_result = default.exec(Command.new("curl -I #{url}"))
-      assert_match(/200 OK/, curl_headers_result.stdout, "EPEL #{el_version} should be reachable")
+      assert_match(/200 OK/, curl_headers_result.stdout, "EPEL #{el_version} should be reachable at #{url}")
     end
 
     step 'Verify el_version numbers 5,6,7 are found on the epel resource' do
--- a/lib/beaker/host_prebuilt_steps.rb
+++ b/lib/beaker/host_prebuilt_steps.rb
@@ -440,7 +440,11 @@ module Beaker
         if host['platform'] =~ /debian|ubuntu|cumulus/
           host.exec(Command.new("sudo su -c \"service ssh restart\""), {:pty => true})
         elsif host['platform'] =~ /arch|centos-7|el-7|redhat-7|fedora-(1[4-9]|2[0-9])/
-          host.exec(Command.new("sudo -E systemctl restart sshd.service"), {:pty => true})
+          if host['hypervisor'] = "docker"
+            host.exec(Command.new("/bin/kill -HUP `cat /var/run/sshd.pid`"), {:pty => true})
+          else
+            host.exec(Command.new("sudo -E systemctl restart sshd.service"), {:pty => true})
+          end
         elsif host['platform'] =~ /centos|el-|redhat|fedora|eos/
           host.exec(Command.new("sudo -E /sbin/service sshd reload"), {:pty => true})
         elsif host['platform'] =~ /(free|open)bsd/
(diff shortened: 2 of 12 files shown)


alwinmark:
Oh, i've currently still a problem because I missed one place where it tries to use systemctl

alwinmark:
Ok I found them, I added a special case, but I would suggest not to use systemctl at all, because the maintainance effort would be much less, if you would directly use sshd, as it is installed very similar on all linux derivates

alwinmark:
Ok now everything should be gucci

lib/beaker/host_prebuilt_steps.rb
```diff
@@ -440,7 +440,11 @@ def enable_root_login host, opts
         if host['platform'] =~ /debian|ubuntu|cumulus/
           host.exec(Command.new("sudo su -c \"service ssh restart\""), {:pty => true})
         elsif host['platform'] =~ /arch|centos-7|el-7|redhat-7|fedora-(1[4-9]|2[0-9])/
-          host.exec(Command.new("sudo -E systemctl restart sshd.service"), {:pty => true})
+          if host['hypervisor'] = "docker"
```
kevpl: @alwinmark shouldn't this be `==`, not `=`?

lib/beaker/hypervisor/docker.rb
```diff
@@ -264,7 +264,7 @@ def dockerfile_for(host)
 
         # How to start a sshd on port 22.  May be an init for more supervision
         # Ensure that the ssh server can be restarted (done from set_env) and container keeps running
-        cmd = host['docker_cmd'] || ["sh","-c","service #{service_name} start ; tail -f /dev/null"]
+        cmd = host['docker_cmd'] || ["sh","-c","/usr/sbin/sshd; tail -f /dev/null"]
```
kevpl: @alwinmark it looks like with this change, `service_name` isn't being used anymore. Is this true? If that's the case, then the setting statements above should be removed.

If you need to pull out `service_name`, please also refactor the platform-specific statements into a new method, to help get this one under control from a length perspective.

tvpartytonight:
Hey @alwinmark , thanks for your PR. I had a question about your PR message

>  as any other OS which is using sshd threw systemctl instead of init.d.

You've implemented a fix here that is docker specific, but the real issue seems to be differences between systemctl and init.d. I think that the logic should be implemented so that it isn't `hypervisor` aware.

Also, we'd like to see some tests for this change as well.

alwinmark:
@tvpartytonight what do I have to do, to run the acceptance tests locally? Then I would write an additional test for centos7 el7 and others. 

I now have removed the logic which seperates between init systems to restart sshd by a more generic one.

Another approach to solve this issue (and another one too), is to fake systemctl, but thats really hacky. But it would than enable beaker tests for service restarts as well.

kevpl:
@alwinmark there are rake tasks for acceptance testing beaker in our `Rakefile`. Check out `test:acceptance:base` to get started.

Also, it would be a big help to us if you could rebase your changes on top of already merged ones, so that we don't see them in the commit log or diffs. We typically don't merge changes that have polluted git logs, so this will need to happen for merging anyways.

acceptance/tests/base/external_resources_test.rb
```diff
@@ -1,9 +1,11 @@
 test_name 'External Resources Test' do
   step 'Verify EPEL resources are up and available' do
     def epel_url_test(el_version)
-      url = "#{options[:epel_url]}/epel-release-latest-#{el_version}.noarch.rpm"
+      url_base = options[:epel_url]
+      url_base = options[:epel_url_archive] if el_version == 5
```
ferglor: Would it be nicer to have this assignment as `url_base = el_version == 5 ? options[:epel_url_archive] : options[:epel_url]`?

kevpl:
@alwinmark last comment is still current

alwinmark:
Yes I think we have to close this one. Either we have to wrap systemd completely or systemd user have to use other hypervisors.

Because we have problems with beaker tests where the manifest is doing service restarts, this wouldn't fix our problem anymore.

Don't know what do you think about it? Drop Docker support for Centos7 and everything which is really using systemd?

tvpartytonight:
@alwinmark we will be closing this PR, but keeping the JIRA ticket open and prioritising a fix for a future sprint. Thanks for bringing this issue to our attention!

PR #1375, opened 2017-04-12
BKR-1089 Fixing Custom Facts added to environment on Windows

cdenneen:
* This version of the fix removes the unnecessary space

PR #1376, opened 2017-04-12
BKR-1089 Fix Custom Facts added to environment on windows hosts

cdenneen:
* This fix quotes the fact to avoid extra spaces

### DIFF
--- a/lib/beaker/host/pswindows/exec.rb
+++ b/lib/beaker/host/pswindows/exec.rb
@@ -164,9 +164,31 @@ module PSWindows::Exec
 
     environment_string = ''
     env_array.each_with_index do |env|
-      environment_string += "set #{env} && "
+      environment_string += "set \"#{env}\" && "
     end
     environment_string
+  end
+
+  def environment_variable_string_pair_array env
+    env_array = []
+    env.each_key do |key|
+      val = env[key]
+      if val.is_a?(Array)
+        val = val.join(':')
+      else
+        val = val.to_s
+      end
+      # doing this for the key itself & the upcase'd version allows us to remain
+      # backwards compatible
+      # TODO: (Next Major Version) get rid of upcase'd version
+      key_str = key.to_s
+      keys = [key_str]
+      keys << key_str.upcase if key_str.upcase != key_str
+      keys.each do |env_key|
+        env_array << "#{env_key}=\"#{val}\""
+      end
+    end
+    env_array
   end
 
   # Overrides the {Windows::Exec#ssh_permit_user_environment} method,
--- a/spec/beaker/host/pswindows/exec_spec.rb
+++ b/spec/beaker/host/pswindows/exec_spec.rb
@@ -50,5 +50,22 @@ module Beaker
         expect( instance.mv(origin, destination, false) ).to be === 0
       end
     end
+    describe '#environment_string' do
+      let(:host) { {'pathseparator' => ':'} }
+
+      it 'returns a blank string if theres no env' do
+        expect( instance.environment_string( {} ) ).to be == ''
+      end
+
+      it 'takes an env hash with var_name/value pairs' do
+        expect( instance.environment_string( {:HOME => '/', :http_proxy => 'http://foo'} ) ).
+          to be == 'set "HOME="/"" && set "http_proxy="http://foo"" && set "HTTP_PROXY="http://foo"" && '
+      end
+
+      it 'takes an env hash with var_name/value[Array] pairs' do
+        expect( instance.environment_string( {:LD_PATH => ['/', '/tmp']}) ).
+          to be == "set \"LD_PATH=\"/:/tmp\"\" && "
+      end
+    end
   end
 end


glennsarti:
Both #1376 and #1375 are fixes for issue BKR-1089.  Just need to decide which is preferable.

kevpl:
@cdenneen we would prefer this PR over #1375, as it's more explicit, and less likely to be whitespace-fixed during a later PR without the necessary context needed to know that it shouldn't be. 

Please add some spec tests to make sure we don't break this in future versions, and we'll get this through. Note that we have a release coming out this week, so the quicker the turn around, the more likely you can get the improvement this week rather than having to wait.

cdenneen:
@kevpl not sure how I would spec test this since it only occurs on windows with cmd.exe?

kevpl:
@cdenneen the spec tests are really narrowed in, so you'd just be testing that specific method, and more importantly, testing that it doesn't mess up your quoting in the future. I would add a `describe #environment_string` section in the [`exec_spec.rb`](https://github.com/puppetlabs/beaker/blob/master/spec/beaker/host/pswindows/exec_spec.rb) file, testing what happens if you get a few different parameter-scenarios & what the output should look like

cdenneen:
Thanks Kevin I will give it a shot but I've never written a test like this
only from rspec-puppet perspective so I'm not sure I'll be able make it
work as expected. If you have a moment and submit PR to this PR I can merge
it to save some time getting this fix out, I'd appreciate the assist.

cdenneen:
@kevpl As far as I can tell the .travis-ci.yml isn't configured properly for this repo so trying to create proper tests is a bit hard to make sure they are valid and working.
I'm trying to run spec tests locally but it seems to hang after:

```
Beaker::CLI
  initializing and parsing
    instance variable initialization
      creates a logger for use before parse is called
      generates the timestamp
    #parse_options
      returns self
      replaces the logger object with a new one
    #print_version_and_options
      prints the version and dumps the options
  #configured_options
    returns a list of options that were not presets
  #combined_instance_and_options_hosts
    combines the options and instance host objects
  execute!
    test fail mode
```

No failure just sits and does nothing.

cdenneen:
@kevpl copying from the unix #environment_string test but replacing `never` to `true` on the `is_powershell` I would be applying the following patch:

```
diff --git i/spec/beaker/host/pswindows/exec_spec.rb w/spec/beaker/host/pswindows/exec_spec.rb
index 6a365473..359db2cd 100644
--- i/spec/beaker/host/pswindows/exec_spec.rb
+++ w/spec/beaker/host/pswindows/exec_spec.rb
@@ -50,5 +50,24 @@ module Beaker
         expect( instance.mv(origin, destination, false) ).to be === 0
       end
     end
+
+    describe '#environment_string' do
+      let(:host) { {'pathseparator' => ':'} }
+
+      it 'returns a blank string if theres no env' do
+        expect( instance ).to receive( :is_powershell? ).true
+        expect( instance.environment_string( {} ) ).to be == ''
+      end
+
+      it 'takes an env hash with var_name/value pairs' do
+        expect( instance.environment_string( {:HOME => '/', :http_proxy => 'http://foo'} ) ).
+          to be == 'env HOME="/" http_proxy="http://foo" HTTP_PROXY="http://foo"'
+      end
+
+      it 'takes an env hash with var_name/value[Array] pairs' do
+        expect( instance.environment_string( {:LD_PATH => ['/', '/tmp']}) ).
+          to be == "env LD_PATH=\"/:/tmp\""
+      end
+    end
   end
 end
```

But while I can commit this to a PR branch that is currently able to be cleanly merged, not sure how this would affect merging or if it would even pass so I haven't committed to my fork.

kevpl:
@cdenneen I don't see any reason why that would hurt the merge-ability of your PR, it would just be more code, it doesn't appear to conflict with anything already in `pswindows/exec_spec.rb`.

My one caveat is that it doesn't look like the tests that you copied over include any testing of the code you actually changed. I would make sure to add at least one to verify we don't break your changes in the future, then we should be good for merging once this passes through the automation.

For your issue running spec tests locally, I would make sure that running them on the master branch is successful just to cover your bases before running them on your branch. I'm not seeing any hangups right now with the master branch's spec testing. We don't use travis for our spec testing, we run that in our internal Jenkins jobs.

cdenneen:
@kevpl issue with the environment strings on windows was an extra space being added to the fact... so "foo" turned into "foo " so I would think these tests actually should be verifying behavior or am I wrong?
As far as locally... I have tried master branch and it hangs in same spot... is there something I need to configure for the tests after CLI to run?

cdenneen:
@kevpl ok so running `bundle exec rake spec` hangs but running `bundle exec rspec` works.

Here is current issue... as I see on `lib/beaker/host/pswindows/exec.rb:163` it provides environment_variable_string_pair_array but with adding above tests it errors with:

```
 NoMethodError:
       undefined method `environment_variable_string_pair_array' for #<Beaker::PSWindowsExecTest:0x007fcea65e03b8>
     # ./lib/beaker/host/pswindows/exec.rb:163:in `environment_string'
     # ./spec/beaker/host/pswindows/exec_spec.rb:66:in `block (3 levels) in <module:Beaker>'
```

cdenneen:
@kevpl Should be good now if you can review

kevpl:
@cdenneen yeah, you're right, it seems like the testing should cover the case that you're talking about.  Would you mind filing a ticket on the `bundle exec rake spec` vs `bundle exec rspec` issue, and referencing it here so that we can have those linked?

Other than that, everything looks good to move forward towards merging.

cdenneen:
@kevpl Done https://tickets.puppetlabs.com/browse/BKR-1121

cdenneen:
@kevpl @tvpartytonight can this be merged?

kevpl:
@cdenneen we'll kick the automation today since it's been a while, & if everything goes well, we'll merge it later today. Apologies for the delay.

kevpl:
transient unrelated failure, rekicking...

cdenneen:
@kevpl Can you recheck this?

kevpl:

kevpl:

PR #1384, opened 2017-05-01
(BKR-1113) do_scp_to now also copies files that start with a dot

pegasd:
With this patch do_scp_to also copies all files that start with a dot (this issue was noticed when using copy_module_to).

The fixed pattern is taken from glob() documentation (https://ruby-doc.org/core-2.4.1/Dir.html#method-c-glob).

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -438,7 +438,7 @@ module Beaker
           @logger.trace result.stdout
         end
       else # a directory with ignores
-        dir_source = Dir.glob("#{source}/**/*").reject do |f|
+        dir_source = Dir.glob("#{source}/**/{*,.*}").reject do |f|
           f.gsub(/\A#{Regexp.escape(source)}/, '') =~ ignore_re #only match against subdirs, not full path
         end
         @logger.trace "After rejecting ignored files/dirs, going to scp [#{dir_source.join(", ")}]"
--- a/acceptance/tests/base/dsl/helpers/host_helpers/scp_to_test.rb
+++ b/acceptance/tests/base/dsl/helpers/host_helpers/scp_to_test.rb
@@ -34,6 +34,19 @@ test_name "dsl::helpers::host_helpers #scp_to" do
     end
   end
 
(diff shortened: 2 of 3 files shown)


lib/beaker/host.rb
```diff
@@ -438,7 +438,7 @@ def do_scp_to source, target_path, options
           @logger.trace result.stdout
         end
       else # a directory with ignores
-        dir_source = Dir.glob("#{source}/**/*").reject do |f|
+        dir_source = Dir.glob("#{source}/**/{*,.*}").reject do |f|
```
ferglor: Can you update this PR with tests for this?

epivenpro: I gave it a go in the next commit.
However, the spec tests pass with both versions of the code, which is probably due to FakeFS implementation in respect to `glob()`'s behavior.
Unfortunately, getting the acceptance suite going was a little outside of scope of my weekend endeavour, but the test should look similar to this.

pegasd: I gave it a go in the next commit.
However, the spec tests pass with both versions of the code, which is probably due to FakeFS implementation in respect to `glob()`'s behavior.
Unfortunately, getting the acceptance suite going was a little outside of scope of my weekend endeavour, but the test should look similar to this.

tvpartytonight:

lib/beaker/host.rb
kevpl: @pegasd this looks like a good change, but we'd like to not change the default behavior, but instead have this guarded behind an option like (for example) `:copy_hidden_files` for the method.

tvpartytonight:
@pegasd last comment still relevant.

kevpl:
@pegasd last comment is still current. This PR will be closed in a week if there's no activity.

lib/beaker/host.rb
pegasd: Was away for a while, just now catching up.
@kevpl Is there a simple example I can dig into to hide this behavior behind an option?

pegasd: Also, how will you use that option from a puppet module when running tests?

PR #1385, opened 2017-05-02
(MAINT) add SSH setup link to vmpooler doc

kevpl:
[skip ci]

@genebean, @joel-m-allen, I think one of you were talking about needing the SSH doc for vmpooler hosts. Is that correct?


tvpartytonight:
LGTM 👍

PR #1386, opened 2017-05-02
(BKR-1116) Update wait_for_connection_failure to allow reconnect

samwoods1:
Fixed issue in wait_for_connection_failure where sleep would increment by 1 second for each iteration of wait
Added options to wait_for_connection_failure to allow specific wait intervals, enabling waiting for disconnect then reconnect
Added spec tests for wait_for_connection_failure

### DIFF
--- a/lib/beaker/ssh_connection.rb
+++ b/lib/beaker/ssh_connection.rb
@@ -111,15 +111,18 @@ module Beaker
     # @option options [Boolean] :pty Should we request a terminal when attempting
     #                                to send a command over this connection?
     # @option options [String] :stdin Any input to be sent along with the command
+    # @param options [Integer] :poll_interval A set interval of time to poll for connectivity, defaults to nil
+    # @param options [Integer] :max_tries The maximum number of times to retry connection, defaults to 11
     # @param [IO] stdout_callback An IO stream to send connection stdout to, defaults to nil
     # @param [IO] stderr_callback An IO stream to send connection stderr to, defaults to nil
     # @return [Boolean] true if connection failed, false otherwise
     def wait_for_connection_failure options = {}, stdout_callback = nil, stderr_callback = stdout_callback
       try = 1
       last_wait = 2
-      wait = 3
+      wait = options[:poll_interval] ? options[:poll_interval] : 3
+      max_tries = options[:max_tries] ? options[:max_tries] : 10
       command = 'echo echo' #can be run on all platforms (I'm looking at you, windows)
-      while try < 11
+      while try <= max_tries
         result = Result.new(@hostname, command)
         begin
           @logger.notify "Waiting for connection failure on #{@hostname} (attempt #{try}, try again in #{wait} second(s))"
@@ -147,12 +150,16 @@ module Beaker
         slept = 0
         stdout_callback.call("sleep #{wait} second(s): ")
         while slept < wait
-          sleep slept
+          sleep 1
           stdout_callback.call('.')
           slept += 1
         end
         stdout_callback.call("\n")
-        (last_wait, wait) = wait, last_wait + wait
+        if options[:poll_interval]
+          wait = options[:poll_interval]
+        else
+          (last_wait, wait) = wait, last_wait + wait
+        end
         try += 1
       end
       false
--- a/spec/beaker/ssh_connection_spec.rb
+++ b/spec/beaker/ssh_connection_spec.rb
@@ -284,5 +284,57 @@ module Beaker
 
     end
 
+    describe '#wait_for_connection_failure' do
+      before :each do
+
+      end
+
+      it 'waits at increasing intervals for connection failure' do
+        @mock_ssh = double('ssh').as_null_object
+        @mock_stdout_callback = Object.new
+        expect( @mock_ssh ).to receive( :open_channel ).exactly(10).times
+        expect( Net::SSH ).to receive( :start ).with( ip, user, ssh_opts) { @mock_ssh }
+        connection.connect
+
+        expect( @mock_ssh ).to receive( :loop ).exactly(10).times
+        allow( @mock_stdout_callback ).to receive(:call)
+
+        expect( connection ).to receive(:sleep).with(1).exactly(605).times
+        # 605 = Sum of 10 times at increasing intervals: 3,5,8,13,21,34,55,89,144,233
+        connection.wait_for_connection_failure({}, @mock_stdout_callback)
+      end
+
+      it 'waits at a set interval if poll_interval is provided' do
+        @mock_ssh = double('ssh').as_null_object
+        @mock_stdout_callback = Object.new
+        expect( @mock_ssh ).to receive( :open_channel ).exactly(10).times
+        expect( Net::SSH ).to receive( :start ).with( ip, user, ssh_opts) { @mock_ssh }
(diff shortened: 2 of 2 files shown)


samwoods1:
looks like on SLES we had a couple of failures:
Failed Tests Cases:
          Test Case /var/lib/jenkins/workspace/qe_beaker_intn-sys_beaker-acceptance-base-vpool/agent/sles12/acceptance/tests/base/dsl/helpers/host_helpers/check_for_package_test.rb reported: #<Minitest::Assertion: Expected false to be truthy.>
            Test line: /var/lib/jenkins/workspace/qe_beaker_intn-sys_beaker-acceptance-base-vpool/agent/sles12/acceptance/tests/base/dsl/helpers/host_helpers/check_for_package_test.rb:82:in `block (4 levels) in run_test'
          Test Case /var/lib/jenkins/workspace/qe_beaker_intn-sys_beaker-acceptance-base-vpool/agent/sles12/acceptance/tests/base/dsl/helpers/host_helpers/install_package_test.rb reported: #<Minitest::Assertion: package was not successfully installed>
            Test line: /var/lib/jenkins/workspace/qe_beaker_intn-sys_beaker-acceptance-base-vpool/agent/sles12/acceptance/tests/base/dsl/helpers/host_helpers/install_package_test.rb:35:in `block (4 levels) in run_test'

samwoods1:

tvpartytonight:

lib/beaker/ssh_connection.rb
```diff
@@ -113,13 +113,15 @@ def close
     # @option options [String] :stdin Any input to be sent along with the command
     # @param [IO] stdout_callback An IO stream to send connection stdout to, defaults to nil
     # @param [IO] stderr_callback An IO stream to send connection stderr to, defaults to nil
+    # @param [Integer] poll_interval A set interval of time to poll for connectivity, defaults to nil
+    # @param [Integer] max_tries The maximum number of times to retry connection, defaults to 11
     # @return [Boolean] true if connection failed, false otherwise
-    def wait_for_connection_failure options = {}, stdout_callback = nil, stderr_callback = stdout_callback
+    def wait_for_connection_failure options = {}, stdout_callback = nil, stderr_callback = stdout_callback, poll_interval = nil, max_tries = 11
```
kevpl: @samwoods1 I'm wondering why you prefer adding new parameters rather than having these as options hash values?

samwoods1: I've just always preferred that over hash values, but I can update it since that's how Beaker works all over the place anyways.

lib/beaker/ssh_connection.rb
```diff
@@ -147,12 +149,16 @@ def wait_for_connection_failure options = {}, stdout_callback = nil, stderr_call
...
           stdout_callback.call('.')
           slept += 1
         end
         stdout_callback.call("\n")
-        (last_wait, wait) = wait, last_wait + wait
+        if poll_interval
+          wait = poll_interval
+        else
+          (last_wait, wait) = wait, last_wait + wait
+        end
```
kevpl: @samwoods1 if I'm understanding your implementation correctly, this will poll at the same time interval if you pass the `poll_interval` parameter, or do the fibonacci backoff it's not provided. Is that correct?

I'm curious why you want to provide polling at a set interval in addition to the fibonacci backoff that already exists. This code in general (historically, unrelated to your changes) seems like a reproduction of the [repetition.rb](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/shared/repetition.rb) code, and so my immediate response would be that if this code is touched, it should be refactored to use the shared repetition code. Having both cases supported seems to be a block to that refactor, so I'm curious about what's motivating you to need the static polling interval.

spec/beaker/ssh_connection_spec.rb
```diff
@@ -284,5 +284,57 @@ module Beaker
 
     end
 
+    describe '#wait_for_connection_failure' do
```
kevpl: yay tests! 💯

lib/beaker/ssh_connection.rb
kevpl: @samwoods1 I'm not super against it on principle, I was just wondering if either:

1. you had a different reason than the [usual arguments](https://www.google.com/search?client=opera&q=ruby+options+hash+argument&sourceid=opera&ie=UTF-8&oe=UTF-8)
1. you were sure to be aware of all users of this method, which hopefully [this search](https://github.com/search?utf8=✓&q=org%3Apuppetlabs+wait_for_connection_failure+extension%3Arb&type=Code&ref=advsearch&l=&l=) covers (standard signature change issue)...

lib/beaker/ssh_connection.rb
samwoods1: The motivation is that I am doing snapshot and restore VMs for the HA tests and using this method to determine when the SSH connection goes down - meaning that the restore has begun, then waiting for the connection to come back up afterwards.  The problem was that if it went through a few iterations of waiting for the connection failure then it would sometimes completely miss the agent going offline since it goes offline for only 15 seconds or so and then get stuck waiting forever for the connection to fail.  I am thinking the same issue could arise if for example someone were using this method to wait for a system reboot to happen - although this would be less likely because the system reboot usually begins nearly immediately and can take quite a while to complete.

lib/beaker/ssh_connection.rb
samwoods1: 1. No different reasons, just enjoy the additional intellisense in my IDE when there are actual parameters to methods.
2. It shouldn't impact any existing calls since the new params are at the end of the method and have default values.

lib/beaker/ssh_connection.rb
kevpl: @samwoods1 I'm down with the motivation here, and having the toggle-ability of using the pattern that fits you, but I still don't think the logic should be executed here.

Considering what you need is the ability to toggle between the fibonacci backoff or a static poll timer, a method should be added to `repetition.rb` that allows for what you're looking for, pulling the mechanics of repetition out of `wait_for_connection_failure`. That method would look something like this:
```ruby
def repeat_choosing_method_for(max_tries=11, poll_interval=nil, &block)
  if poll_interval
    repeat_for(poll_interval, block)
  else
    repeat_fibonacci_style_for(max_attempts, block)
  end
end
```

Once that's implemented, all the repetition logic will be out of the `wait_for_connection_failure` method, and it won't have to care about any of that. This has a few benefits over the current implementation:

1. you can use this toggle-ability between repetition-types in other places in your testing
1. it's less likely that there will be timing bugs like this, the more centralized the repetition implementations are

kevpl:
[latest conversation](https://github.com/puppetlabs/beaker/pull/1386#discussion_r115857813) still ongoing inside one of the outdated change folds.

kevpl:
@samwoods1 last comment is still current.

kevpl:
@samwoods1 last comment is still current. If there isn't any motion on this PR in a week, we'll close it due to inactivity.

samwoods1:
I have temporarily removed the snapshot/restore functionality from the HA tests, so this got bumped down on my priority list.  When I add that back to the HA tests, I will get back to this.  We can close the PR if you want to wait until that happens.  I don't have a good timeframe on when that will be.

PR #1391, opened 2017-05-15
Disable reverse DNS on Docker containers

Sharpie:
This patch adds `UseDNS no` to the sshd configuration added to Docker
containers. This prevents SSH connections from triggering revers DNS lookups
which can take up to 30 seconds to time out per connection. Disabling this
behavior cut a simple beaker-rspec "hello world" from 14 minutes to 12 seconds
on the centos:centos7 image.


ferglor:
👍

PR #1394, opened 2017-05-19
(BKR-840) Fix docs to use more trivial package

rishijavia:
Updated docs to use HTTPD package.

### DIFF
--- a/docs/tutorials/lets_write_a_test.md
+++ b/docs/tutorials/lets_write_a_test.md
@@ -1,74 +1,82 @@
 ## The Task
 
-Consider if mcollectived incorrectly spawned a new process with every puppet agent run on Ubuntu 10.04.  We need an acceptance test to check that a new process is not spawned and to ensure that this issue does not regress in new builds.
+We will write a test to check if a package (specifically HTTPD) is installed and running. To do this we will write two files:
+1. `install.rb` - This file will install the package and start the service
+2. `mytest.rb` - This file will have our core tests that checks if the package is installed and running
 
-## Figure Out Test Steps
+Note: We will exclude Windows OS from our testing due to variation of package name.
+
+## Figure out steps
 
 What needs to happen in this test:
 
-* Install PE
-* Restart mcollective twice
-* Check to see if more than one mcollective process is running
+* Install and run HTTPD
+  * Install HTTPD if its not available on our SUT
+  * Start HTTPD service
+* Testing
+  * Test HTTPD is installed
+  * Test HTTPD service is running
 
 ## Create a host configuration file
-    $ beaker-hostgenerator redhat7-64ma > redhat7-64ma.yaml
+    $ beaker-hostgenerator redhat7-64 > redhat7-64.yaml
 
-## Install PE
+This command will generate a host file for our system under test (SUT). It will use vmpooler as hypervisor for the host. Please check out [this](https://github.com/puppetlabs/beaker/tree/master/docs/how_to/hypervisors) doc to learn more about hypervisors for beaker.
 
-We prefer to install PE once and then run a set of tests, so PE installation should not be part of the actual acceptance test.
+## Install and run HTTPD
 
-    $ mkdir setup
-    $ cd setup
-    $ cat > install.rb << RUBY
-    test_name "Installing Puppet Enterprise" do
-    install_pe
-    RUBY
-    $ cd ..
+Make a file named `install.rb` and put the following code into it:
 
-This places our install steps in a ruby script (install.rb) which will run on localhost, but the #install_pe method knows where to install puppet enterprise based on the host configuration in use.
-The install.rb script is used in our commandline to beaker, below.
+```ruby
+test_name "Installing and runnning HTTPD" do
+
+  # Don't run the install script on the following platform
+  confine :except, :platform => 'windows'
+
+  step "Install HTTPD" do
+    hosts.each do |host|
+      # Install HTTPD if it is not available on our SUT
+      install_package(host, 'httpd') unless check_for_package(host, 'httpd')
+    end
+  end
+
+  step "Start HTTPD" do
+    hosts.each do |host|
+      # Start HTTPD service
+      on(host, "service httpd start")
+    end
+  end
+
+end
+```
+
+This places our install steps in a ruby script (`install.rb`) which will run on your SUT. The install.rb script is used in our commandline to beaker, below.
 
 ## Create a test file
 
-We need to create a test file to run.
-
-### Define some test commands to run
-#### Restart mcollective twice
-
-Here's our magic command that restarts mcollective:
-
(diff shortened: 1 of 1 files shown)


docs/tutorials/lets_write_a_test.md
```diff
@@ -1,74 +1,81 @@
...
+  * Test HTTPD is running
 
 ## Create a host configuration file
     $ beaker-hostgenerator redhat7-64ma > redhat7-64ma.yaml
 
-## Install PE
+## Install and run HTTPD
 
-We prefer to install PE once and then run a set of tests, so PE installation should not be part of the actual acceptance test.
+We prefer to install and run HTTPD once and then run a set of tests, so PE installation should not be part of the actual acceptance test. Make a file named `install.rb` and put the following code into it:
```
ferglor: Should this be "We prefer to install and run HTTPD once and then run a set of tests, so **HTTPD** installation should not be part of the actual acceptance test"?

docs/tutorials/lets_write_a_test.md
```diff
@@ -1,74 +1,81 @@
...
+    end
+  end
+
+  step "Make sure HTTPD is running" do
+  	hosts.each do |host|
+   		 on(host, "systemctl is-active httpd") do |result|
+         # Check if HTTPD is running
+         assert_equal(0, result.exit_code)
+       end
+  	end
```
ferglor: Indentation seems a little messed up here

docs/tutorials/lets_write_a_test.md
rishijavia: I think dropping it is a good idea

docs/tutorials/lets_write_a_test.md
rishijavia: my bad! totally fixing this right now

docs/tutorials/lets_write_a_test.md
```diff
@@ -1,74 +1,81 @@
...
-    restart_command = "bash -c '[[ -x /etc/init.d/pe-mcollective ]] && /etc/init.d/pe-mcollective restart'"
-
-#### Check to see if more than one mcollective process is running
+```ruby
+test_name "Installing and runnning HTTPD" do
+
+	step "Make sure HTTPD is not installed" do
+	  hosts.each do |host|
+		# Check if HTTPD is not installed
+		assert_equal(false, check_for_package(host, 'httpd'))
```
tvpartytonight: It doesn't seem correct that this will fail if `httpd` is already installed; I'd rather see that this test just ensure that httpd is installed, not fail if `httpd` is already available on an SUT.

docs/tutorials/lets_write_a_test.md
```diff
@@ -1,74 +1,81 @@
...
+		install_package(host, 'httpd')
+	  end
+	end
+
+	step "Start HTTPD" do
+	  hosts.each do |host|
+		# Start HTTPD service
+		on(host, "service httpd start")
+	  end
+	end
```
kevpl: @rishijavia too much indenting here

docs/tutorials/lets_write_a_test.md
```diff
@@ -1,74 +1,81 @@
...
 
-Here's our magic command that throws an error if more than one mcollective process is running:
+end
+```
 
-    process_count_check = "bash -c '[[ $(ps auxww | grep [m]collectived | wc -l) -eq 1 ]]'"
+This places our install steps in a ruby script (`install.rb`) which will run on localhost. The install.rb script is used in our commandline to beaker, below.
```
tvpartytonight: I think that the term `localhost` here should be replaced with "your machine" or "your coordinator".  We shouldn't conflate the networking term `localhost` to mean the machine beaker is running on.

docs/tutorials/lets_write_a_test.md
```diff
@@ -1,74 +1,81 @@
...
-  # Restart once
-  on(host, restart_command) { assert_equal(0, exit_code) }
-  # Restart again
-  on(host, restart_command) { assert_equal(0, exit_code) }
-  # Check to make sure only one process is running
-  on(host, process_count_check) { assert_equal(0, exit_code) }
+test_name "Check if HTTPD is installed and running" do
+
+  # Don't run these tests on the following platform
+  confine :except, :platform => 'windows'
```
tvpartytonight: Not sure if this should be addressed in your `install.rb` for earlier feedback about SUTs not being suitable.

rishijavia:
test this please

docs/tutorials/lets_write_a_test.md
```diff
@@ -1,74 +1,80 @@
...
-* Restart mcollective twice
-* Check to see if more than one mcollective process is running
+* Install and run HTTPD
+  * Install HTTPD if its not available on our SUT
+  * Start HTTPD service
+* Testing
+  * Test HTTPD is installed
+  * Test HTTPD service is running
 
 ## Create a host configuration file
```
kevpl: @rishijavia reading through the doc in its entirety, things are looking good overall. I'm noticing that this section seems orphaned as-is, however.

I know that you don't actually have changes for this section, so I'm not asking for changes to your changes 💃 ,  but I think it would be awesome if we could put in a little blurb about hypervisors, just saying how this will create a hosts file for `x` hypervisor, & if you'd like to learn more about that, checkout `y` doc. Like [this blurb](https://github.com/puppetlabs/beaker/blob/master/docs/tutorials/quick_start_rake_tasks.md#pre-requisites).

rishijavia: @kevpl good idea! Added that under the `Create a host configuration file` section.

docs/tutorials/lets_write_a_test.md
```diff
@@ -1,74 +1,82 @@
...
     $ beaker-hostgenerator redhat7-64ma > redhat7-64ma.yaml
```
tvpartytonight: Nit: the `hostgenerator` string doesn't need the "ma" here, because you never refer to the hosts by their roles.

docs/tutorials/lets_write_a_test.md
```diff
@@ -1,74 +1,82 @@
...
-We prefer to install PE once and then run a set of tests, so PE installation should not be part of the actual acceptance test.
-
-    $ mkdir setup
-    $ cd setup
-    $ cat > install.rb << RUBY
-    test_name "Installing Puppet Enterprise" do
-    install_pe
-    RUBY
-    $ cd ..
+This command will generate a host file for our system under test (SUT). It will use vmpooler as hypervisor for the host. Please check out [this](https://github.com/puppetlabs/beaker/blob/master/docs/tutorials/quick_start_rake_tasks.md#hypervisor-argument) doc to learn more about hypervisors for beaker.
```
tvpartytonight: I don't think that you want to link to that subsection of the quick start tasks, but link to the general hypervisors section: https://github.com/puppetlabs/beaker/tree/master/docs/how_to/hypervisors

PR #1396, opened 2017-05-26
(BKR-1124) Pass info text to summary logger to print on file

rishijavia:
Spec test breaking [here](https://github.com/puppetlabs/beaker/blob/master/spec/beaker/test_suite_spec.rb#L247)

`summary_logger` is the instance that logs the summary to terminal and the file.

`print_test_result` was passing the test_cases that failed to be notified using `@logger` instance [here](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/test_suite.rb#L155)

I changed it to return the string and then am notifying it using the `summary_logger` instance.

### DIFF
--- a/lib/beaker/test_suite.rb
+++ b/lib/beaker/test_suite.rb
@@ -116,22 +116,22 @@ module Beaker
 
         summary_logger.notify "Failed Tests Cases:"
         (grouped_summary[:fail] || []).each do |test_case|
-          print_test_result(test_case)
+          summary_logger.notify print_test_result(test_case)
         end
 
         summary_logger.notify "Errored Tests Cases:"
         (grouped_summary[:error] || []).each do |test_case|
-          print_test_result(test_case)
+          summary_logger.notify print_test_result(test_case)
         end
 
         summary_logger.notify "Skipped Tests Cases:"
         (grouped_summary[:skip] || []).each do |test_case|
-          print_test_result(test_case)
+          summary_logger.notify print_test_result(test_case)
         end
 
         summary_logger.notify "Pending Tests Cases:"
         (grouped_summary[:pending] || []).each do |test_case|
-          print_test_result(test_case)
+          summary_logger.notify print_test_result(test_case)
         end
 
         summary_logger.notify("\n\n")
@@ -152,7 +152,7 @@ module Beaker
         else
           test_case.test_status
         end
-        @logger.notify "  Test Case #{test_case.path} #{test_reported}"
+        "  Test Case #{test_case.path} #{test_reported}"
       end
 
       # Writes Junit XML of this {TestSuiteResult}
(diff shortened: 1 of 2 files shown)


kevpl:
@rishijavia if I'm understanding this right, it looks like you're losing the functionality of the error printouts if things aren't successful for the test, is that right? Have you looked at what the error/failure cases for this printing looks like before & after?

I'm wondering if maybe rather than swapping out the logger instead of using the convenience method, if the `summary_logger` should be passed into the convenience method, so we can keep the same functionality we had before, but we could now be able to rely on the `summary_logger` to write things down in both places?

Of course another question for this is whether or not we can change `summary_logger`'s signature. Are these the only uses of it? Either way, this wouldn't be the worst, as it seems we could provide a default value that allows old signature use for the method anyway...

rishijavia:
@kevpl you are right, we loose the color scheme we have defined for the `@logger` instance for terminal output. A thing to note though is that with the original code, we are using the same color for all the messages (for skip, failed, errored and pending) as we are using the `.notify` method for all which results to blue color.

Anyways, before submitting the PR I tried sending the `summary_logger` to the convenience method and was able to retain the color scheme of the `@logger` instance and was able to print the output on the file too, but whenever you call `summary_logger.notify` it would output on terminal too. And in this way we would have duplicate test cases printed on terminal:
1. By the `@logger` instance in blue
2. By the `summary_logger` instance in normal font color

I have yet to look into changing the `summary_logger`'s signature and play around with that.

rishijavia:
@kevpl so if I turn on color notifications for `summary_logger` from it's [signature](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/test_suite.rb#L350), I was able to get color scheme we have on `@logger` instance and was able to get the terminal output with the color scheme. But on the other hand, each of the line in file output starts with the terminal color code for the line for e.g.:
```
[00;00m[00;34mSkipped Tests Cases:
[00;00m[00;34m  Test Case acceptance/tests/cowsay.rb 
```
Feel like leaving color off on `summary_logger` is the call to make `.txt` file more readable.

The other option I was able to come up is that we can make `summary_logger` quite at its signature so it doesn't output to terminal and let `@logger` instance take care of terminal output and `summary_logger` would print to file. But for this option, each message would have to be saved in a variable so we can notify it on both of our logger instances (doesn't seem like a good practice).

kevpl:
@rishijavia I agree, it seems that leaving color off on the `summary_logger` is the simpler option to make the text file usable, so let's go with that one for now.

rishijavia:
ok to test

rishijavia:
test this please

rishijavia:
test this please

kevpl:
@rishijavia these changes still remove the ability to have errors have their stacktraces printed out, since we're not using the `print_test_result` method. I'm not seeing that addressed in our earlier conversation, am I just missing that?

rishijavia:
@kevpl `summary_logger` is a `Logger` instance that prints at both terminal and the `.txt` summary file. I changed the functionality such that `print_test_result` returns a string and when I call  `summary_logger.notify print_test_result(test_case)` it logs it on terminal stacktrace and the `.txt` summary file.

kevpl:
@rishijavia ah, thanks for clearing that up for me! Everything looks good! 👍

PR #1397, opened 2017-05-26
(maint) Update contributing resource links

johnduarte:
This commit replaces the stale link to the old puppet wiki with a link to the
contributor guidelines.


kevpl:
ah, thanks for fixing this @johnduarte! 👍

PR #1398, opened 2017-05-31
(BKR-1096) Block hosts flag with provision subcommand

rishijavia:
Removed hosts as a Thor class_option so it cannot be used by all subcommands. It is restricted to be used only with init command. Hosts is added as Thor method_option with init. Changed spec and acceptance tests accordingly.

### DIFF
--- a/acceptance/tests/subcommands/init.rb
+++ b/acceptance/tests/subcommands/init.rb
@@ -7,11 +7,12 @@ test_name 'use the init subcommand' do
 
   step 'ensure beaker init writes YAML configuration files to disk' do
     delete_root_folder_contents
-    result = on(default, 'beaker init --hosts centos6-64')
-    assert_match(/Trying as beaker-hostgenerator input/, result.stdout)
+    on(default, 'beaker init --hosts centos6-64')
     subcommand_options = on(default, "cat #{SubcommandUtil::SUBCOMMAND_OPTIONS}").stdout
     subcommand_state = on(default, "cat #{SubcommandUtil::SUBCOMMAND_STATE}").stdout
-    assert(YAML.parse(subcommand_options).to_ruby.class == Hash)
+    parsed_options = YAML.parse(subcommand_options).to_ruby
+    assert(parsed_options["HOSTS"].count == 1)
+    assert(parsed_options.class == Hash)
     assert(YAML.parse(subcommand_state).to_ruby.class == Hash)
   end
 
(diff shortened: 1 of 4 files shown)


acceptance/tests/subcommands/provision.rb
```diff
@@ -10,8 +10,12 @@ def delete_root_folder_contents
     delete_root_folder_contents
     on(default, 'beaker init')
     result = on(default, 'beaker provision --hosts centos6-64')
+    assert_match(/ERROR/, result.raw_output)
+    on(default, 'beaker init --hosts centos6-64')
+    result = on(default, 'beaker provision')
     assert_match(/Using available host/, result.stdout)
 
+
```
kevpl: @rishijavia is this just leftover?

rishijavia: @kevpl yes, updating commit

tvpartytonight:
@rishijavia this looks good; I think that if we are going to merge this work in, we should also make it so that `beaker init` requires a `--hosts` argument, otherwise users will be kind of stuck if the simply do `beaker init` without `--hosts`. I'll make a ticket for that work, it should be simply making `method_option` for hosts required for init.

rishijavia:
test this please

rishijavia:
test this please

acceptance/tests/subcommands/init.rb
```diff
@@ -7,7 +7,8 @@ def delete_root_folder_contents
 
   step 'ensure beaker init writes YAML configuration files to disk' do
     delete_root_folder_contents
-    on(default, 'beaker init')
+    result = on(default, 'beaker init --hosts centos6-64')
+    assert_match(/Trying as beaker-hostgenerator input/, result.stdout)
```
tvpartytonight: I'm not clear why this assert is here; it seems to be testing that beaker is trying input as a hostgenerator string, but that is not what this step is trying to test; it's ensuring that YAML files are outputted to disk on the SUT. It seems like this assert is testing the implementation not pertinent to subcommand. 

A better test might be to read in the subcommand_options and ensure that the HOSTS object has a single element in the array.

acceptance/tests/subcommands/provision.rb
```diff
@@ -10,8 +10,10 @@ def delete_root_folder_contents
     delete_root_folder_contents
     on(default, 'beaker init')
     result = on(default, 'beaker provision --hosts centos6-64')
+    assert_match(/ERROR/, result.raw_output)
```
tvpartytonight: I'd like to ensure that the error message at least has some reference to the flag that is not allowed. Something like `/ERROR(.+)--hosts/` would work.

rishijavia:
test this please

tvpartytonight:
👍

PR #1399, opened 2017-06-03
Issue where beaker gets stuck when an invalid Vagrantfile had been generated

dillongilmore:
I'm not sure what the correct solution is here. In one case if beaker generates an invalid Vagrantfile then beaker will remain stuck without a way to get unstuck unless you manually go find the Vagrantfile and delete it.

In another case if the destroy fails to destroy an existing VM and the original Vagrantfile is deleted the user might be in a very bad state where there's a VM but Vagrant will not work with that existing VM unless the Vagrantfile is restored. I'm not sure what the chances of this actually happening.

Another solution might be to try to destroy then raise a custom exception to let the user know about these 2 states and extend beaker's cleanup rake task to clear the Vagrantfile. This moves the decision from here to the user's judgment.

Let me know what the best choice is and I can make whatever changes need to happen. I found myself in this situation where beaker kept failing even though I knew it should of been fine. I had to inspect the source code to realize I needed to clear out the old Vagrantfile.

Thanks!!

### DIFF
--- a/lib/beaker/hypervisor/vagrant.rb
+++ b/lib/beaker/hypervisor/vagrant.rb
@@ -175,7 +175,24 @@ module Beaker
       if @options[:provision]
         #setting up new vagrant hosts
         #make sure that any old boxes are dead dead dead
-        vagrant_cmd("destroy --force") if File.file?(@vagrant_file)
+        begin
+          vagrant_cmd("destroy --force") if File.file?(@vagrant_file)
+        rescue RuntimeError => e
+          # LATER: use <<~MESSAGE once we're on Ruby 2.3
+          @logger.debug(%Q{
+            Beaker failed to destroy the existing VM's. If you think this is
+            an error or you upgraded from an older version of beaker try
+            verifying the VM exists and deleting the existing Vagrantfile if
+            you believe it is safe to do so. WARNING: If a VM still exists
+            please run 'vagrant destroy'.
+
+            cd #{@vagrant_path}
+            vagrant status
+            vagrant destroy # only need to run this is a VM is not created
+            rm #{@vagrant_file} # only do this if all VM's are actually destroyed
+          }.each_line.map(&:strip).join("\n"))
+          raise e
+        end
 
         make_vfile @hosts, @options
 
--- a/spec/beaker/hypervisor/vagrant_spec.rb
+++ b/spec/beaker/hypervisor/vagrant_spec.rb
@@ -420,6 +420,13 @@ EOF
         vagrant.provision
       end
 
+      it "notifies user of failed provision" do
+        vagrant.provision
+        expect( vagrant ).to receive( :vagrant_cmd ).with( "destroy --force" ).and_raise( RuntimeError )
+        expect( options['logger'] ).to receive( :debug ).with( /Vagrantfile/ )
+        expect{ vagrant.provision }.to raise_error( RuntimeError )
+      end
+
       it "can cleanup" do
         expect( vagrant ).to receive( :vagrant_cmd ).with( "destroy --force" ).once
         expect( FileUtils ).to receive( :rm_rf ).once


kevpl:
Hi @dillongilmore, thanks for submitting this question/situation.

Is the Vagrantfile cleared during the `destroy` command you're highlighting in your changes, or is the file being cleared out just by our `make_vfile` method that writes a new one?

I would say that if we can control when the clearing happens, then we should log the failure to destroy, keep the Vagrantfile around around, and quit the run, letting the user know the circumstances around the failure, so they can manually take care of whatever issue is causing the failure.

I'm open to entertaining other failure modes, but with the information I have now, that seems like the best option, if it's do-able. Let me know if you have any questions or comments for me about this.

dillongilmore:
It's being overwritten by `make_vfile`.

This makes sense to me. What should the debug message be? I think we should include the full absolute path to the `@vagrantfile` or even include the shell command to clear the vagrantfile manually. ` rm @vagrantfile`

Something along the lines of:

```
Beaker failed to destroy the existing VM's. If you think this is an error or you upgraded from an older version of beaker try verifying the VM exists and deleting the existing Vagrantfile.

    cd @vagrant_path
    vagrant status
    rm @vagrantfile
```

kevpl:
@dillongilmore 👍 sounds good to me.

rishijavia:
Hi @dillongilmore,

Thank you for contributing and submitting this PR!

Would you mind editing this PR to match our [contributing guideline](https://github.com/puppetlabs/beaker/blob/master/CONTRIBUTING.md#getting-started)? It requires you to create a JIRA ticket about the same and editing your git commit to match the required commit message format with the JIRA ticket number on it (more information can be found on the same contributing page). You can edit your current commit message by `git commit --amend` and force push it.

Also as this is a functionality change, we would need some spec tests for this.

Thank you!

dillongilmore:
Thanks for the info!!

I went ahead and made those changes and tested it. I'm not sure exactly how to add a unit test. If someone can point me in the right direction I can add that. In this case we need to somehow force Beaker to output an invalid Vagrantfile so that the vagrant destroy will fail and the error message will output.

To test this manually you can use the vagrant-libvirt hypervisor then edit the generated Vagrantfile's mac address to something invalid and that should do the trick.

LMK what else I need to do!! Thanks!

rishijavia:
@dillongilmore thanks for creating a ticket and making these changes.

We use Rspec for unit/spec testing. For this to merge we need updated spec tests to pass through our CI (we will kick it on once you submit updated tests).

An example of Spec test can be found [here](https://github.com/puppetlabs/beaker/blob/master/spec/beaker/hypervisor/vagrant_spec.rb#L417-L421). You might want to make a similar block for this case and can use [expect-receive](https://relishapp.com/rspec/rspec-mocks/v/3-6/docs/basics/expecting-messages) and [return](https://relishapp.com/rspec/rspec-mocks/v/3-6/docs/configuring-responses/returning-a-value) statements to create a Runtime error and test if and expected error message is raised. Spec tests are usually used to make sure that the method API is working as expected, its calling the functions like it should and its raising errors appropriately.

More information about writing and running testing can be found in this [section](https://github.com/puppetlabs/beaker/blob/master/CONTRIBUTING.md#testing) of our contributing docs.

kevpl:
@dillongilmore last comment is still current

kevpl:
@dillongilmore last comment is still current. If no more activity happens in a week, we'll close this PR due to inactivity.

dillongilmore:
Hey there. Sorry for that. I've been busy with work/vacations. I still plan to make the requested spec tests as requested this weekend. Thanks!!

dillongilmore:
Hi there, new to rspec. I have this, which is passing

        it "notifies user of failed provision" do                                                                                                                                                                  
          allow( vagrant ).to receive( :vagrant_cmd ).with( "destroy --force" ).and_raise( RuntimeError )                                                                                                          
          vagrant.provision                                                                                                                                                                                        
        end

Is this the right way to do this? I tried to put an expect on the logger, but I didn't see any examples for that so I avoided doing it.

rishijavia:
@dillongilmore, yes you are on the right path.

Here you are stubbing (fake calling) the rescue block of you code but are not actually testing if the message is passed to `@logger` object.

Before your allow statement you need to expect `@logger` object to to receive `debug` method call with the argument of the error message. But instead of testing the whole error message you can use a regex to test a part of it. An example of the same can be found [here](https://github.com/puppetlabs/beaker/blob/6415a791b66efd0d6c73f2d48e9ee56aa594df7e/spec/beaker/test_suite_spec.rb#L291). Your call to `@logger` object will be different than the example and you'll be expecting `:debug` instead of `:error` and instead of the last `.never` part you will have `with` (more [rspec docs](https://relishapp.com/rspec/rspec-mocks/v/3-2/docs/setting-constraints/matching-arguments) about with). Somewhere along the lines of:

`expect(@logger).to receive(:debug).with(/myregex/)`

dillongilmore:
Got it!! Thank you for your patience and guidance on this. Small chance, but hopefully this helps other people.

Note I had to call vagrant.provision twice in my spec tests. The first time to generate the Vagrantfile and the 2nd time to get vagrant destroy to be called and raise the exception. I didn't see any other tests for that so I'm not sure if that function call was covered before or not.

rishijavia:
Absolutely! This looks good to me, kicking off CI

rishijavia:

kevpl:

kevpl:
@dillongilmore nit for next time: there are parens around the ticket ref or `MAINT` keyword. If that's right, then we won't squash when we merge.

Thanks for your contribution! It should be out in the next cut of Beaker this Wednesday! Refer to our [wiki](https://github.com/puppetlabs/beaker/wiki) for the status & release date of our next release at any time!

PR #1400, opened 2017-06-05
(BKR-1136) Use kill instead of stop on docker containers

Sharpie:
The stop command is called during the cleanup phase for the docker hypervisor
in order to shut containers down prior to removing them. This command defaults
to waiting for 10 seconds before sending a SIGKILL to whatever process the
container is running. For containers running a process like init that ignores
the initial stop and term signals, this results in significant time being spent
on politeness when the container is going to be completely destroyed anyway.

This patch switches the stop command out for a simple kill.


mchllweeks (approved):

ferglor:
👍

PR #1401, opened 2017-06-05
(BKR-994) Read subcommands options from home directory

rishijavia:
This patch changed options priority to read options from `$HOME/.beaker/subcommands_options.yaml` with the lowest priority after the presets.
Added respective spec tests and modified others to fit this in the priority chain. Please note that as the priority hierarchy is changing there are some places in the module and their spec files that require modification in order to fit this in.
The new options priority now (from highest to lowest):
1. ENV variables
2. ARGV variables
3. 'CONFIG' section of hosts file
4. Options file
5. Subcommand options
6. Config file from home directory (this patch)
7. Presets

### DIFF
--- a/lib/beaker/options/parser.rb
+++ b/lib/beaker/options/parser.rb
@@ -205,7 +205,8 @@ module Beaker
       #   3.  the 'CONFIG' section of the hosts file
       #   4.  options file values
       #   5.  subcommand options, if executing beaker subcommands
-      #   6.  default or preset values are given the lowest priority
+      #   6.  subcommand options from $HOME/.beaker/subcommand_options.yaml
+      #   7.  default or preset values are given the lowest priority
       #
       # @param [Array] args ARGV or a provided arguments array
       # @raise [ArgumentError] Raises error on bad input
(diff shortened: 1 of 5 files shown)


lib/beaker/options/parser.rb
```diff
@@ -216,7 +217,12 @@ def parse_args(args = ARGV)
         cmd_line_options[:command_line] = ([$0] + args).join(' ')
         @attribution = @attribution.merge(tag_sources(cmd_line_options, "flag"))
 
-        # Subcommands are the first to get merged into presets
+        # Global options from $HOME/.beaker/subcommands_options.yaml are first to get merged into presets
+        homedir_options = Beaker::Options::OptionsFileParser.parse_homedir_options
```
ferglor: Just thinking out loud, I'm wondering if we should refactor SubcommandOptionsParser to accept a file path, and use that here? Since presumably we don't want to parse these options on init, which the SubcommandOptionsParser takes care of

rishijavia: @ferglor I agree. @tvpartytonight and I talked about this and it makes sense for this to go in `SubcommandOptionsParser`. I'm refactoring the code and will update this commit

lib/beaker/options/parser.rb
```diff
@@ -205,7 +205,8 @@ def update_option(key, value, source)
       #   3.  the 'CONFIG' section of the hosts file
       #   4.  options file values
       #   5.  subcommand options, if executing beaker subcommands
-      #   6.  default or preset values are given the lowest priority
+      #   6.  options from $HOME/.beaker/subcommands_options.yaml
```
kevpl: @rishijavia does this step (reading `$HOME/.beaker/...`) happen on all beaker executions (both a full beaker run & in subcommands)? We definitely don't want `subcommand_options.yaml` (looks like both subcommand & options are plural, that's not right, right @tvpartytonight ?) to be read in the context of a full beaker run.

rishijavia: @kevpl yes the same was pointed out by @ferglor in his review.
I'm changing this to be a part of `SubcommandOptionsParser` rather than `OptionsFileParser`

PR #1402, opened 2017-06-06
(BKR-994) Read subcommands options from home directory

rishijavia:
This patch changed options priority to read options from $HOME/.beaker/subcommands_options.yaml with the lowest priority after the presets.
Added respective spec tests and modified others to fit this in the priority chain. Please note that as the priority hierarchy is changing there are some places in the module and their spec files that require modification in order to fit this in.
The new options priority now (from highest to lowest):
1. ENV variables
2. ARGV variables
3. 'CONFIG' section of hosts file
4. Options file
5. Subcommand options
6. Subcommand options file from home directory (this patch)
7. Presets

PS - Also, there was an empty file (`lib/beaker/options/homedir_options_file_parser.rb`) merged from my poor github management that is being deleted here :(

### DIFF
--- a/lib/beaker/options/parser.rb
+++ b/lib/beaker/options/parser.rb
@@ -205,7 +205,8 @@ module Beaker
       #   3.  the 'CONFIG' section of the hosts file
       #   4.  options file values
       #   5.  subcommand options, if executing beaker subcommands
-      #   6.  default or preset values are given the lowest priority
+      #   6.  subcommand options from $HOME/.beaker/subcommand_options.yaml
+      #   7.  default or preset values are given the lowest priority
       #
       # @param [Array] args ARGV or a provided arguments array
       # @raise [ArgumentError] Raises error on bad input
@@ -216,7 +217,12 @@ module Beaker
         cmd_line_options[:command_line] = ([$0] + args).join(' ')
         @attribution = @attribution.merge(tag_sources(cmd_line_options, "flag"))
 
-        # Subcommands are the first to get merged into presets
+        # Global subcommand options from $HOME/.beaker/subcommand_options.yaml are first to get merged into presets
+        homedir_options = Beaker::Options::SubcommandOptionsParser.parse_subcommand_options(args, home_dir=true)
+        @attribution = @attribution.merge(tag_sources(homedir_options, "homedir"))
+        @options.merge!(homedir_options)
+
+        # Subcommands are the second to get merged into presets
         subcommand_options = Beaker::Options::SubcommandOptionsParser.parse_subcommand_options(args)
         @attribution = @attribution.merge(tag_sources(subcommand_options, 'subcommand'))
         @options.merge!(subcommand_options)
--- a/lib/beaker/options/parser.rb
+++ b/lib/beaker/options/parser.rb
@@ -218,12 +218,12 @@ module Beaker
         @attribution = @attribution.merge(tag_sources(cmd_line_options, "flag"))
 
         # Global subcommand options from $HOME/.beaker/subcommand_options.yaml are first to get merged into presets
-        homedir_options = Beaker::Options::SubcommandOptionsParser.parse_subcommand_options(args, home_dir=true)
+        homedir_options = Beaker::Options::SubcommandOptionsParser.parse_subcommand_options(args, ENV['HOME']+'/.beaker/subcommand_options.yaml')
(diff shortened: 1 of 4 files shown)


spec/beaker/options/parser_spec.rb
```diff
@@ -147,6 +147,11 @@ module Options
                             }                                
           } }
           let(:subcommand_file) {@subcommand_file || {:level => 'fifth'}}
+          let(:homedir_file) {@homedir_file || {:level => 'sixth',
+                                                :ssh => {
+                                                    :auth_methods => 'auth_home_123'
+                                                }
```
kevpl: this should only be indented 2 spaces

lib/beaker/options/subcommand_options_file_parser.rb
```diff
@@ -3,18 +3,30 @@ module Options
...
+        if File.exist?(options_file_path)
+          result = YAML.load_file(options_file_path)
+        end
+        result
+      end
+
       # @return [OptionsHash, Hash] returns an empty OptionHash or loads subcommand options yaml
       #   from disk
-      def self.parse_subcommand_options(argv)
+      def self.parse_subcommand_options(argv, home_dir=false)
```
kevpl: I'm confused as to how this (passing a `home_dir` boolean argument) is better than having the user specify the path to the file & using the constant in the call if you'd like to use the home directory.

Or even better, we could use the default behavior to not need to specify the home directory (although I'm not sure if the home directory or the current directory would be a better default).

ferglor: So instead of passing in a boolean here to determine whether or not we should read from the home directory, I would pass in the file to read from as a string, e.g. `parse_subcommand_options(argv, options_file)`. That way, we don't need the `if` branch with the early return statement, but rather, we can just do `result = SubcommandOptionsParser.parse_options_file(options_file)`, and then, we can change line 221 of parser.rb to be something like `homedir_options = Beaker::Options::SubcommandOptionsParser.parse_subcommand_options(args, ENV['HOME']+'/.beaker/subcommand_options.yaml')`, and change line 226 of parser.rb to something like: `subcommand_options = Beaker::Options::SubcommandOptionsParser.parse_subcommand_options(args, Beaker::Subcommands::SubcommandUtil::SUBCOMMAND_OPTIONS)`.

rishijavia: @kevpl updated code to use `file_option` argument and using constants for both, homedir file and subcommands options file.

lib/beaker/options/parser.rb
```diff
@@ -216,8 +217,13 @@ def parse_args(args = ARGV)
         cmd_line_options[:command_line] = ([$0] + args).join(' ')
         @attribution = @attribution.merge(tag_sources(cmd_line_options, "flag"))
 
-        # Subcommands are the first to get merged into presets
-        subcommand_options = Beaker::Options::SubcommandOptionsParser.parse_subcommand_options(args)
+        # Global subcommand options from $HOME/.beaker/subcommand_options.yaml are first to get merged into presets
+        homedir_options = Beaker::Options::SubcommandOptionsParser.parse_subcommand_options(args, ENV['HOME']+'/.beaker/subcommand_options.yaml')
```
kevpl: the string here is the same as `SUBCOMMAND_OPTIONS` used below, [right](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/subcommands/subcommand_util.rb#L15)? If that's the case, why not use that same thing here?

Also, these lines are getting crazy long. Maybe set the file names on separate lines? Or, it seems like this code in this block & the one below is the same code, is it not? If that's the case, why not loop over the file sources and not have the duplicate code?

spec/beaker/options/parser_spec.rb
```diff
@@ -147,6 +147,11 @@ module Options
                             }                                
           } }
           let(:subcommand_file) {@subcommand_file || {:level => 'fifth'}}
+          let(:homedir_file) {@homedir_file || {:level => 'sixth',
+                                :ssh => {
+                                    :auth_methods => 'auth_home_123'
+                                }
```
kevpl: white space is still a problem here...

rishijavia: goddamn rubymine! 😠

PR #1405, opened 2017-06-10
(BKR-1139) Add /usr/local/bin to SSH PATH on OSX

Iristyle:
- Typically an interactive OSX SSH shell has /usr/local/bin added
   to PATH.

   However, for non-interactive SSH shells, OSX locks down security a
   bit tighter and doesn't add /usr/local/bin by default. Beaker has
   already added `PermitUserEnvironment yes` to /etc/ssh/sshd_config
   to be more lenient and allow additional elements to be added by
   the way of ~/.ssh/environment file written during Beaker host
   initialization.

 - This is necessary because of OSX SIP (System Integrity Protection)
   that marks /usr/bin read-only, and suggests /usr/local/bin be the
   target destination for any user installed binaries.

   For instance, should the Ruby `gem` command be used to install
   gems with binstubs (like Bundler), they are written to
   /usr/local/bin. Without this change the `bundle` command is not
   in PATH. This can prevent Puppet source based acceptance testing
   through ci:test:git from finding binaries it may install during
   setup.

### DIFF
--- a/lib/beaker/host/unix/exec.rb
+++ b/lib/beaker/host/unix/exec.rb
@@ -294,6 +294,8 @@ module Unix::Exec
       add_env_var('PKG_PATH', "http://ftp.openbsd.org/pub/OpenBSD/#{version}/packages/#{arch}/")
     elsif self['platform'] =~ /solaris-10/
       add_env_var('PATH', '/opt/csw/bin')
+    elsif self['platform'] =~ /osx/
+      add_env_var('PATH', '/usr/local/bin')
     end
 
     #add the env var set to this test host
--- a/spec/beaker/host/unix/exec_spec.rb
+++ b/spec/beaker/host/unix/exec_spec.rb
@@ -157,5 +157,83 @@ module Beaker
         expect(instance.selinux_enabled?).to be === false
       end
     end
+
+    describe '#ssh_set_user_environment' do
+      before :each do
+        allow( instance ).to receive( :exec )
+        allow( instance ).to receive( :[] ).with( 'platform' ).and_return( 'centos-6-x86' )
+        allow( instance ).to receive( :[] ).with( :ssh_env_file ).and_return( '' )
+      end
+
+      it 'adds the env argument hash to the environment' do
+        allow( instance ).to receive( :mkdir_p )
+        allow( instance ).to receive( :[] ).with( :ssh_env_file ).and_return( '' )
+
+        env_hash = {
+          'key1' => 'value1',
+          'key2' => 'value2',
+          'key3' => 'value3'
+        }
+        expect( instance ).to receive( :add_env_var ).with( 'PATH', '$PATH' )
+        env_hash.each do |k,v|
+          expect( instance ).to receive( :add_env_var ).with( k, v )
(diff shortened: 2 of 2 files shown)


Iristyle:
More info on the OSX and why `/usr/local/bin` isn't in the SSH env by default - https://git-annex.branchable.com/forum/OSX__39__s_default_sshd_behaviour_has_limited_paths_set/

kevpl:
@Iristyle I've taken the time to add some tests to this in https://github.com/Iristyle/beaker/pull/1. If you'd either merge those into this branch, or add your own tests, we'd like to have some tests to make sure we don't break this functionality in the future. Once that's in, we'll get this through automation once more & merge it. Thanks!

kevpl:
build timed out on the reboot test 🤢 , not something that we typically see, but it doesn't seem related to the changes in this PR, so I'll re-kick.

Iristyle:
@kevpl it turns out that other platforms look to have similar needs (SLES for instance).  I'm wondering if it would be that harmful to just include `/usr/local/bin` in PATH for all platforms even if it they don't explicitly need it?

kevpl:
@Iristyle I've created #1410 to try this out against our testing. I like the idea, but I worry that there's some one-off OS that won't eat the error if the path doesn't exist or something hinky like that..

Iristyle:
@kevpl thanks.

A couple of additional notes:

* I can't think of a situation where it would be harmful to include `/usr/local/bin` automatically in `PATH` - and doing it in one spot for all platforms is pretty convenient / future-proof, rather than incrementally adding individual platforms, which will probably be cumbersome over time.
* But... where does `/usr/local/bin` belong in `PATH` order?  I believe it should be at the front of `PATH` and preferred over other system directories as this is where newer tools would be installed.  Interestingly, while validating `ci:test:git` against SLES11, `/usr/local/bin` did not end up at the beginning of the evaluated `PATH`, even with it at the beginning of `PATH` in `~/.ssh/environment`. It appears SLES has a different way of merging environments together than other platforms, so the resultant `PATH` does not prefer `/usr/local/bin` as desired.  In the SLES11 case, I had to `"echo 'PATH=/usr/local/bin:$PATH' > ~/.bashrc"` to get the desired behavior.


For reference, a few more notes on the SLES11 `ci:test:git` case:

* SLES11 system Ruby is 1.8.7, which is not compatible with Puppet
* To enable `ci:test:git` to work I needed to build / install Ruby 2.4
* I used the simplest most vanilla build process possible (i.e. no RVM or rbenv) - grab a tarball and kick off gcc
* The `~/.bashrc` solution was the only one that would allow the newly compiled / installed Ruby to override the pre-installed system Ruby
* I did not try uninstalling the system Ruby package, which might also achieve the same thing - but uninstalling existing packages probably isn't a great solution as package names / dependencies will vary by platform. It's probably best to just let multiple Ruby versions peacefully co-exist.

kevpl:
this conversation has been sitting fallow between this PR, #1406, & #1410 for a very long time.

Since that's the case, it doesn't seem like a pressing issue, so I'm going to close these PRs.

PR #1406, opened 2017-06-11
(BKR-1140) Add LANG / LC_ALL en_us.UTF-8 for SSH 

Iristyle:
- Without a default LANG variable set on hosts under test, Ruby will
   default to setting its Encoding.default_external to US-ASCII

   In particular, this appears to impact vmpooler host types:

   arista4-32
   cumulus25-64
   osx1009
   osx1010
   osx1011
   sles10

 - Alternatively, image templates could be modified, but injecting
   this with Beaker by default seems like a more robust solution.

 - Also alternatively, hosts can use the :host_env setting to inject
   the LANG variable.  However, this is currently difficult to do
   in conjunction with beaker-hostgenerator which is typically used
   in Puppet CI pipelines.

### DIFF
--- a/lib/beaker/host/unix/exec.rb
+++ b/lib/beaker/host/unix/exec.rb
@@ -296,6 +296,11 @@ module Unix::Exec
       add_env_var('PATH', '/opt/csw/bin')
     end
 
+    # some hosts may omit the LANG variable, which sets Ruby to US-ASCII
+    # UTF-8 is the accepted standard encoding for Puppet tools
+    add_env_var('LANG', 'en_US.UTF-8')
+    add_env_var('LC_ALL', 'en_US.UTF-8')
+
     #add the env var set to this test host
     env.each_pair do |var, value|
       add_env_var(var, value)


Iristyle:
Although this has fixed OSX it doesn't appear to have worked as desired on cumulus25-64 because of either `LC_ALL=C`or `LANGUAGE=C`:

```
y7gr3fd9merdngg.delivery.puppetlabs.net (cumulus25-64-1) executed in 0.12 seconds
=> ["SHELL=/bin/bash",
 "SSH_CLIENT=10.32.130.9 61087 22",
 "LC_ALL=C",
 "USER=root",
 "SSH_AUTH_SOCK=/tmp/ssh-WEvjXgcSoi/agent.2134",
 "MAIL=/var/mail/root",
 "PATH=PATH:/usr/bin:/opt/puppet-git-repos/hiera/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/bin/X11",
 "PWD=/root",
 "LANG=en_us.UTF-8",
 "SHLVL=1",
 "HOME=/root",
 "LANGUAGE=C",
 "LOGNAME=root",
 "SSH_CONNECTION=10.32.130.9 61087 10.32.124.223 22",
 "VTYSH_PAGER=/bin/cat",
 "_=/usr/bin/env"]
[3] pry(#<Beaker::TestCase>)> on agent, 'ruby -e "puts Encoding.default_external"'

    y7gr3fd9merdngg.delivery.puppetlabs.net (cumulus25-64-1) 21:20:30$ ruby -e "puts Encoding.default_external"
      US-ASCII

    y7gr3fd9merdngg.delivery.puppetlabs.net (cumulus25-64-1) executed in 0.14 seconds
=> #<Beaker::Result:0x007fc1c2910c18
 @cmd="ruby -e \"puts Encoding.default_external\"",
 @exit_code=0,
 @host="cumulus25-64-1",
 @output="US-ASCII\n",
 @raw_output="US-ASCII\n",
 @raw_stderr="",
 @raw_stdout="US-ASCII\n",
 @stderr="",
 @stdout="US-ASCII\n">
```

kevpl:
@Iristyle it seems like breaking that logic into its own method would be good, especially with the additional work you're mentioning being needed for Cumulus.

That will also help with the testing that was added in https://github.com/puppetlabs/beaker/pull/1405, as it will break with these changes (due to using `add_env_var` explicitly), but can be stubbed out by method name for those tests if this is refactored into its own method.

The changes needed for Cumulus aren't in the PR, just in the comment. Is there a reason we're waiting on adding them into the PR? Are the discuss & don't merge labels for this, or another reason?

Iristyle:
@kevpl I'm trying to remove https://github.com/puppetlabs/puppet/blob/ebd96213cab43bb2a8071b7ac0206c3ed0be8e58/acceptance/tests/direct_puppet/supports_utf8.rb#L13 which is currently being skipped due to the SSH encoding configuration.

So far, this changes makes OSX pass, but I've having a little difficulty with the others at the moment, which has led to a few questions:

* Is Beaker OK with using UTF-8 as the default Encoding over SSH (honestly, it's probably the only one that Beaker is capable of  counting on)?
* Do we want to add a helper to Beaker that ensures platforms are properly configured for UTF-8 over SSH?
* Do we want to make changes to images? Or would we prefer to make changes there?
* Do all platforms even allow being set to UTF-8? (I seem to recall @LuvCurves mentioning some network platforms as problematic?)

I didn't want to get too far without answering those questions and making sure we're all on the same page.  This PR is a pretty tactical response to the problem at hand, and there could be better ways, etc.

kevpl:
this conversation has been sitting fallow between this PR, #1405, & #1410 for a very long time.

Since that's the case, it doesn't seem like a pressing issue, so I'm going to close these PRs.

PR #1411, opened 2017-06-19
(BKR-1144) Remove excess checking and make tagging earlier.

hunner:
The tags can't be passed to create() in aws-sdk-v1 but we can move the call closer to creation.

This also reduces the duplicate status checks.

### DIFF
--- a/lib/beaker/hypervisor/aws_sdk.rb
+++ b/lib/beaker/hypervisor/aws_sdk.rb
@@ -47,11 +47,6 @@ module Beaker
       # Perform the main launch work
       launch_all_nodes()
 
-      wait_for_status_netdev()
-
-      # Add metadata tags to each instance
-      add_tags()
-
       # Grab the ip addresses and dns from EC2 for each instance to use for ssh
       populate_dns()
 
@@ -405,7 +400,12 @@ module Beaker
           instance = create_instance(host, ami_spec, nil)
           instances.push({:instance => instance, :host => host})
         end
+        # Add metadata tags to each instance
+        add_tags(instances)
+        # Verify that all hosts are running
         wait_for_status(:running, instances)
+        # Verify extra network device checks, if any
+        wait_for_status_netdev(instances)
       rescue Exception => ex
         @logger.notify("aws-sdk: exception #{ex.class}: #{ex}")
         kill_instances(instances.map{|x| x[:instance]})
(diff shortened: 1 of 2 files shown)


lib/beaker/hypervisor/aws_sdk.rb
```diff
@@ -405,7 +400,12 @@ def launch_all_nodes
           instance = create_instance(host, ami_spec, nil)
           instances.push({:instance => instance, :host => host})
         end
+        # Add metadata tags to each instance
```
tvpartytonight: @hunner from my testing, I'm inclined to think this will shave off ~30 seconds of time; is that sufficient for your needs? I recall in hipchat that we were looking at 20 minute long provision times, which still wouldn't be quite long enough before the reaper came and deleted the instance. 

A different solution might be to add the tags to the `config` object that is passed to `region.instances.create`  in the `create_instance` method. That would ensure that the tags are passed during instantiation. Have you tried to refactor the code into that method?

tvpartytonight: oh, duh, I should have read your first comment. do you know if that restriction is lifted in the v2 of this gem?

hunner: @tvpartytonight The change is that the tagging comes before the "ok" check, not after (which is what takes 20 minutes).

For v2, what restriction do you speak of? The tags cannot be passed in the v1 of the api on instantiation, only in v2, and we are using v1 currently.

(this change is a ci-blocker for f5 module, also).

tvpartytonight: @hunner the restriction I was referring to was inability to specify tags during instance creation for v1.

Given this [blog post](https://aws.amazon.com/blogs/developer/upcoming-stable-release-of-aws-sdk-for-ruby-version-2/), it looks like we could drop in v2 of this gem beside the v1 implementation. Then we could change the `create_instance` method to use the v2 implementation, which would allow beaker to send tags in the original `create_instance` method and remove the `add_tags` method entirely.

A quick survey of the code shows that mapping the config hash to the v2 [create_instances](http://docs.aws.amazon.com/sdkforruby/api/Aws/EC2/Resource.html#create_instances-instance_method) should be relatively straightforward. I'm partial to fixing this issue in the way I described, as it fixes the broader issue.

Is there a pressing need to get this patch in for a release? You mentioned the f5 ticket, but I'm not clear if there is an impending urgent release for f5 scheduled.

hunner: We did the f5 release with manual qa, so no it's pushed now, but I didn't want it to linger too long.

I'll get the PR updated for v2 as well.

tvpartytonight: I think that if you can retarget your work to the [beaker-aws](https://github.com/puppetlabs/beaker-aws) repo, we can merge your work in after the initial release is out the door. We definitely don't need to discover all the gotchas ourselves, so I hope that you can target your work there and we can merge in your changes soon.

rishijavia: I was planning to rebase [beaker-aws](https://github.com/puppetlabs/beaker-aws) on the master branch of beaker once we merge #1434. So I would say if you are planning to work on this targeting beaker-aws, wait till the end of the day to have up to date files on beaker-aws.

PR #1412, opened 2017-06-20
(BKR-1084) Add --skip-last-teardown option

james-stocks:
Add option to skip teardown in the last test, and leave SUT in last testing state.


james-stocks:
Closing - as discussed in the ticket; this option should be for the exec subcommand and not for the main program

PR #1388, opened 2017-05-08
(BKR-895) pull puppet-specific functionality out of beaker

kevpl:
### Tasks to complete before merging

- [x] create automated testing for beaker-puppet PRs
- [x] release a first beaker-puppet gem
- Fix up copies of these jobs to use beaker-puppet
  - [x] [Beaker Acceptance - Reuse Vpool Hosts](https://jenkins-beaker.delivery.puppetlabs.net/job/qe_beaker_intn-sys_beaker-acceptance-reuse-vpool/) ([copied job](https://jenkins-beaker.delivery.puppetlabs.net/job/qe_beaker_intn-sys_beaker-acceptance-reuse-vpool-debug-beaker-puppet/))
  - [x] [Beaker Acceptance - Puppet](https://jenkins-beaker.delivery.puppetlabs.net/job/qe_beaker_intn-sys_beaker-acceptance-puppet-vpool/) ([copied job](https://jenkins-beaker.delivery.puppetlabs.net/job/qe_beaker_intn-sys_beaker-acceptance-puppet-vpool-puppet-pkg-bkr895/))

PR #1414, opened 2017-06-23
(BKR-871) Remove cygwin installation from package installation

tvpartytonight:
Previous to this commit, beaker would ensure that the cygwin executable
was available prior to package installation and download the executable
if it was not found. Later versions of cygwin fixed a flaw in permissions
for the executable, which broke the way beaker ensured and downloaded
the cygwin executable.

This change removes the notion of checking for cygwin at all. Rather
than conflate the installation of cygwin with package installation,
beaker should separate those concerns. Cygwin should be downloaded and
installed separately from package installation, or handled at the image 
level for the VM used.

Also included in this PR is an optimization for determining the the
architecture of the host; rather than spend time introspecting the
system, a host should just examine its own platform string to determine
the architecture.

[Related JIRA Images ticket](https://tickets.puppetlabs.com/browse/IMAGES-6)

### DIFF
--- a/lib/beaker/host/windows/exec.rb
+++ b/lib/beaker/host/windows/exec.rb
@@ -108,4 +108,14 @@ module Windows::Exec
     false
   end
 
+  # Determine if cygwin is actually installed on the SUT. Differs from
+  # is_cygwin?, which is just a type check for a Windows::Host.
+  #
+  # @return [Boolean]
+  def cygwin_installed?
+    output = exec(Beaker::Command.new('cygcheck --check-setup cygwin'), :accept_all_exit_codes => true).stdout 
+    return true if output.match(/cygwin/) && output.match(/OK/)
+    false
+  end
+
 end
--- a/lib/beaker/host/windows/pkg.rb
+++ b/lib/beaker/host/windows/pkg.rb
@@ -12,9 +12,6 @@ module Windows::Pkg
   end
 
   def install_package(name, cmdline_args = '')
-    cygwin = ""
-    rootdir = ""
-
     arch = identify_windows_architecture
 
     if arch == '64'
@@ -25,15 +22,6 @@ module Windows::Pkg
       cygwin = "setup-x86.exe"
     end
 
-    if not check_for_command(cygwin)
-      command = "curl --retry 5 https://cygwin.com/#{cygwin} -o /cygdrive/c/Windows/System32/#{cygwin}"
-      begin
-        execute(command)
-      rescue Beaker::Host::CommandFailure
-        command.sub!('https', 'http')
-        execute(command)
-      end
-    end
     execute("#{cygwin} -q -n -N -d -R #{rootdir} -s http://cygwin.osuosl.org -P #{name} #{cmdline_args}")
   end
 
@@ -101,23 +89,7 @@ module Windows::Pkg
 
   # @api private
   def identify_windows_architecture
-    arch = nil
-    execute("echo '' | wmic os get osarchitecture", :accept_all_exit_codes => true) do |result|
-      arch = if result.exit_code == 0
-        result.stdout =~ /64/ ? '64' : '32'
-      else
-        identify_windows_architecture_from_os_name_for_win2003
-      end
-    end
-    arch
+    platform.arch =~ /64/ ? '64' : '32'
   end
 
-  # @api private
-  def identify_windows_architecture_from_os_name_for_win2003
-    arch = nil
-    execute("echo '' | wmic os get name | grep x64", :accept_all_exit_codes => true) do |result|
-      arch = result.exit_code == 0 ? '64' : '32'
-    end
-    arch
-  end
 end
--- a/lib/beaker/host_prebuilt_steps.rb
+++ b/lib/beaker/host_prebuilt_steps.rb
@@ -107,6 +107,7 @@ module Beaker
         when host['platform'] =~ /cumulus/
           check_and_install_packages_if_needed(host, CUMULUS_PACKAGES)
         when (host['platform'] =~ /windows/ and host.is_cygwin?)
+          raise RuntimeError, "cygwin is not installed on #{host}" if !host.cygwin_installed?
           check_and_install_packages_if_needed(host, WINDOWS_PACKAGES)
         when (host['platform'] =~ /windows/ and not host.is_cygwin?)
           check_and_install_packages_if_needed(host, PSWINDOWS_PACKAGES)
--- a/spec/beaker/dsl/install_utils/windows_utils_spec.rb
+++ b/spec/beaker/dsl/install_utils/windows_utils_spec.rb
@@ -15,7 +15,7 @@ describe ClassMixedWithDSLInstallUtils do
   let( :batch_path )        { '/fake/batch/path' }
   let(:msi_path)            { 'c:\\foo\\puppet.msi' }
   let(:winhost)             { make_host( 'winhost',
-                              { :platform => 'windows',
+                            { :platform => Beaker::Platform.new('windows-2008r2-64'),
                                 :pe_ver => '3.0',
                                 :working_dir => '/tmp',
                                 :is_cygwin => true} ) }
--- a/spec/beaker/host/windows/exec_spec.rb
+++ b/spec/beaker/host/windows/exec_spec.rb
@@ -60,5 +60,23 @@ module Beaker
         instance.reboot
       end
     end
+
+    describe '#cygwin_installed?' do
+      let (:response) { double( 'response' ) }
+
+      it 'uses cygcheck to see if cygwin is installed' do
+        expect( Beaker::Command ).to receive(:new).with("cygcheck --check-setup cygwin").and_return(:foo)
+        expect( instance ).to receive( :exec ).with(:foo, :accept_all_exit_codes => true).and_return(response)
+        expect( response ).to receive(:stdout).and_return('cygwin OK')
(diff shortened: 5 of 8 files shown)


tvpartytonight:
@puppetlabs/windows this should have some 👀  on please and thank-you.

kevpl:
These changes should be rebased / figured out after #1388 merges, as that work is necessary for Puppet 5 preparations.

tvpartytonight:
@kevpl doesn't look like there were any merge conflicts!

lib/beaker/host/windows/pkg.rb
```diff
@@ -101,15 +89,7 @@ def pe_puppet_agent_promoted_package_info( puppet_collection = nil, opts = {} )
 
   # @api private
   def identify_windows_architecture
-    arch = nil
-    execute("echo '' | wmic os get osarchitecture", :accept_all_exit_codes => true) do |result|
-      arch = if result.exit_code == 0
-        result.stdout =~ /64/ ? '64' : '32'
-      else
-        identify_windows_architecture_from_os_name_for_win2003
```
Iristyle: Should `identify_windows_architecture_from_os_name_for_win2003` be removed as well?

lib/beaker/host/windows/pkg.rb
```diff
@@ -101,15 +89,7 @@ def pe_puppet_agent_promoted_package_info( puppet_collection = nil, opts = {} )
...
-      end
-    end
-    arch
+    platform.arch =~ /64/ ? '64' : '32'
```
Iristyle: How does this work for 3rd parties that don't folllow the same naming conventions?

Seems more robust to just inspect the host.

lib/beaker/host/windows/pkg.rb
```diff
@@ -25,15 +22,6 @@ def install_package(name, cmdline_args = '')
       cygwin = "setup-x86.exe"
     end
 
-    if not check_for_command(cygwin)
```
Iristyle: Is this removed code landing in a new method... or does it already exist elsewhere already?

lib/beaker/host/windows/pkg.rb
tvpartytonight: 3rd parties can still name images whatever they like; however, platform strings supplied to beaker should still be split on the `variant`-`version`-`arch`.

In this case, if beaker relies on the robust method, you may end up testing the wrong architecture underneath and you would never know unless you were carefully inspecting your testing logs.

lib/beaker/host/windows/pkg.rb
tvpartytonight: Good catch, this should be removed as well.

lib/beaker/host/windows/pkg.rb
tvpartytonight: Good point @Iristyle. I think that this points to a guarantee that Windows platforms should have a check for cygwin and install it during initialization. I 'll submit another commit to this PR with those changes and correlating tests.

tvpartytonight:
@jcoconnor I'd be interested if you have any feedback for this PR.

lib/beaker/host/windows/pkg.rb
tvpartytonight: I've now added functionality to ensure that cygwin is installed using cygcheck during host validation.

tvpartytonight:
I have addressed the feedback and this is ready for further review @kevpl and @Iristyle

lib/beaker/host/windows/pkg.rb
kevpl: @tvpartytonight has this been done?

lib/beaker/host/windows/pkg.rb
```diff
@@ -101,23 +89,7 @@ def pe_puppet_agent_promoted_package_info( puppet_collection = nil, opts = {} )
...
-      else
-        identify_windows_architecture_from_os_name_for_win2003
-      end
-    end
-    arch
+    platform.arch =~ /64/ ? '64' : '32'
   end
 
-  # @api private
-  def identify_windows_architecture_from_os_name_for_win2003
```
tvpartytonight: @kevpl that method is deleted here.

lib/beaker/host/windows/pkg.rb
tvpartytonight: Yes, it has been removed.

lib/beaker/host/windows/exec.rb
```diff
@@ -108,4 +108,13 @@ def selinux_enabled?()
     false
   end
 
+  # Determin if cygwin is actually installed on the SUT.
+  #
+  # @return [Boolean]
+  def cygwin_installed?
+    output = exec(Beaker::Command.new('cygcheck -c cygwin'), :accept_all_exit_codes => true).stdout 
+    return true if output.match(/cygwin/) and output.match(/OK/)
```
Iristyle: This should probably be `&&` rather than `and`

lib/beaker/host/windows/exec.rb
```diff
@@ -108,4 +108,13 @@ def selinux_enabled?()
     false
   end
 
+  # Determin if cygwin is actually installed on the SUT.
+  #
+  # @return [Boolean]
+  def cygwin_installed?
+    output = exec(Beaker::Command.new('cygcheck -c cygwin'), :accept_all_exit_codes => true).stdout
```
Iristyle: I think `--check-setup` instead of `-c` is more self-documenting

lib/beaker/host/windows/pkg.rb
Iristyle: Ok, so that I'm clear, the platform string should have the requested arch in it (which might not match the actual image arch)?  i.e. you may request 32-bit, even though the image is 64-bit.

I asked because the deleted method ` identify_windows_architecture_from_os_name_for_win2003` appeared to be setting arch based on the host capabilities... was its approach flawed then?

lib/beaker/host_prebuilt_steps.rb
```diff
@@ -107,6 +107,7 @@ def validate_host host, opts
         when host['platform'] =~ /cumulus/
           check_and_install_packages_if_needed(host, CUMULUS_PACKAGES)
         when (host['platform'] =~ /windows/ and host.is_cygwin?)
```
Iristyle: It seems a little odd to have both `host.is_cygwin?` *and* `host.cygwin_installed?` defined (with a possibility of having `is_cygwin?` be true, but `cygwin_installed?` returning `false`). Do the docs clearly explain the difference?

lib/beaker/host/windows/exec.rb
tvpartytonight: This checks the content of the response because cygcheck still returns a `0` if the package is not found.

Beaker does capture the error level, and you could ostensibly say that by virtue of running `cygcheck` you can confidently say that cygwin itself is installed; I opted for this more verbose method to ensure that the cygwin installation was installed without errors.

tvpartytonight: oh, but you are right about using `&&`.

lib/beaker/host/windows/exec.rb
tvpartytonight: Updated to use `--check-setup`.

lib/beaker/host_prebuilt_steps.rb
tvpartytonight: Updated for a more verbose comment for `cygwin_installed?` and to point out that `is_cygwin?` is just a type check for `Windows::Host`.

lib/beaker/host/windows/pkg.rb
tvpartytonight: I think that approach was flawed for the reason stated above; a platform string must have an arch defined, and there seems to be no good reason to introspect the system for it when it is already an attribute of the host construct itself.

tvpartytonight:

tvpartytonight:
@Iristyle I think I have addressed your concerns; can you take another look?

lib/beaker/host/windows/exec.rb
```diff
@@ -108,4 +108,14 @@ def selinux_enabled?()
     false
   end
 
+  # Determine if cygwin is actually installed on the SUT. Differs from
+  # is_cygwin?, which is just a type check for a Windows::Host.
+  #
+  # @return [Boolean]
+  def cygwin_installed?
+    output = exec(Beaker::Command.new('cygcheck --check-setup cygwin'), :accept_all_exit_codes => true).stdout 
+    return true if output.match(/cygwin/) and output.match(/OK/)
```
Iristyle: Based on your comment I think you meant to change `and` to `&&`?

Re: exit codes.  You checked multiple Cygwin versions and they all respond the same way? When the check fails what if it responds with the string `NOT OK`?  I just figured the exit code check would insulate you from such differences.  If the binary is not found it should yield `127` (command not found), else the value from `cygcheck`

Iristyle: To make my suggestion more concrete:

```ruby
def cygwin_installed?
  response = exec(Beaker::Command.new('cygcheck --check-setup cygwin'), :accept_all_exit_codes => true)

  # command not found (127) or cygcheck failure code indicate not installed
  response.exit_code == 0
end
```

Iristyle: So it looks like `cygcheck` always returns `0`, regardless of whether or not it finds a package by the given name.  Oh well....

Iristyle:
:shipit:

PR #1416, opened 2017-06-26
(BKR-895) remove install_utils tests

kevpl:
- matching PR: https://github.com/puppetlabs/beaker-puppet/pull/5
- simple removal of all 3 files, no changes otherwise


ferglor:
Just checking - should `windows_utils_spec.rb` and `pe_defaults_spec.rb` still be under `spec/beaker/dsl/install_utils`?

kevpl:
@ferglor maybe not, but they shouldn't move to beaker-puppet either.

- pe stuff will go to beaker-pe soon, I've been in conversations about that
- the windows stuff is poisoned by host methods as well as installation stuff. that will be entangled separately from this work

So neither of these concerns should block this PR going in

PR #1417, opened 2017-06-28
(BKR-1149) Use last ip from the list when hypervisor is vagrant

rishijavia:


spec/beaker/host_prebuilt_steps_spec.rb
```diff
@@ -516,12 +516,20 @@
...
 
-      expect( Beaker::Command ).to receive( :new ).with( "ip a|awk '/global/{print$2}' | cut -d/ -f1 | head -1", [], {:prepend_cmds=>nil, :cmdexe=>false} ).once
+      expect( Beaker::Command ).to receive( :new ).with( "ip a | awk '/global/{print$2}' | cut -d/ -f1 | head -1", [], {:prepend_cmds=>nil, :cmdexe=>false} ).once
 
       expect( subject.get_ip( host ) ).to be === "192.168.2.130"
 
     end
 
+    it "can exec the get_ip command with tail with vagrant hypervisor" do
+      host = make_host('name', { :stdout => "192.168.2.131\n", :hypervisor => "vagrant" } )
```
nwolfe: Since the change you're making is about whether to get the first or last element of the list, I'd think we'd want to use a list that has more than one element.  That way we can detect if it actually grabbed the first or last.

lib/beaker/host/unix/exec.rb
```diff
@@ -26,7 +26,7 @@ def get_ip
     if self['platform'].include?('solaris') || self['platform'].include?('osx')
       execute("ifconfig -a inet| awk '/broadcast/ {print $2}' | cut -d/ -f1 | head -1").strip
     else
-      execute("ip a|awk '/global/{print$2}' | cut -d/ -f1 | head -1").strip
+      execute("ip a | awk '/global/{print$2}' | cut -d/ -f1 | #{self['hypervisor'] == 'vagrant' ? 'tail' : 'head'} -1").strip
```
nwolfe: I'm just kinda doing a drive-by review here, so I don't have any context, but....

Are we sure we want to *not* do this for solaris and osx platforms?

Maybe the answer is that solaris and osx aren't supported in the vagrant hypervisor so yes, we don't need to do this for them.

spec/beaker/host_prebuilt_steps_spec.rb
nwolfe: Oh but maybe we're not actually executing anything on this string, so it doesn't really matter?

nwolfe:
Also we should update the commits so they include the JIRA number

spec/beaker/host_prebuilt_steps_spec.rb
rishijavia: @nwolfe yes you're right we are not executing the command so it shouldn't matter

lib/beaker/host/unix/exec.rb
kevpl: @nwolfe yeah, we don't typically see usage of those OSes in this hypervisor, so we don't test them ourselves. If there is a more widespread need for this change, I expect a ticket will be filed, then we can move this to a case that includes those two. For now, I think this is fine where it is.

@rishijavia, I'm not a fan of the line length using the ternary, but that's a small nit that won't block merging.

kevpl:
@rishijavia I'll squash the commits for this, but in the future it would be better to prefix the commits with the (BKR-1149) so we can use the rebase & merge instead.

PR #1420, opened 2017-06-30
(BKR-1151) Add port 8142 for masters PXP/PCP

doug-rosser:
Orchestration (PXP/PCP) needs port 8142 opened on all masters, MoM and compile.

### DIFF
--- a/lib/beaker/hypervisor/ec2_helper.rb
+++ b/lib/beaker/hypervisor/ec2_helper.rb
@@ -19,6 +19,7 @@ module Beaker
 
       if roles.include? 'master'
         ports << 8140
+        ports << 8142
       end
 
       if roles.include? 'dashboard'


lib/beaker/hypervisor/ec2_helper.rb
```diff
@@ -19,6 +19,7 @@ def self.amiports(host)
 
       if roles.include? 'master'
         ports << 8140
+        ports << 8142
```
Sharpie: Might also want 8143 for API access.

https://docs.puppet.com/pe/2017.2/sys_req_sysconfig.html#for-split-installs

doug-rosser:
Looks like this PR needs Unit Test changes as well as logic to add 8143 during "split" (and only "split") installs.

kevpl:
@doug-rosser your last comment looks current :trollface:

rishijavia:
BKR-1151 resolved with #1433

PR #1423, opened 2017-07-06
(BKR-1154) place prepend cmds before env vars in commands

branan:
Previously, beaker put environment variables first, followed by the prepended command. This can break in a couple of ways:

* If the prepended command is a shell builtin, it will not work with `env`.
* Even if it could, the environment variables would be applied to the prepended command, not to the actual command.

An example of such a broken command is: `env LD_LIBRARY_PATH=/foo source /etc/profile; puppet --version`

Instead, beaker should put the prepended commands first. This allows it to work correctly if it's a shell builtin, and ensures that the environment variables are properly passed to the intended command.

An example of the fixed command is: `source /etc/profile; env LD_LIBRARY_PATH=foo puppet --version`

### DIFF
--- a/lib/beaker/command.rb
+++ b/lib/beaker/command.rb
@@ -78,7 +78,7 @@ module Beaker
       end
 
       # This will cause things like `puppet -t -v agent` which is maybe bad.
-      cmd_line_array = [env_string, prepend_commands, cmd, options_string, args_string]
+      cmd_line_array = [prepend_commands, env_string, cmd, options_string, args_string]
       cmd_line_array << append_command unless (cmd =~ /ntpdate/ && host[:platform] =~ /cisco_nexus/)
       cmd_line_array.compact.reject( &:empty? ).join( ' ' )
     end


LuvCurves:
👍  Looks Good to me

joshcooper:
Can we run some acceptance tests on different platforms to make sure this doesn't cause issues? I think all of our component Gemfiles support overriding which beaker to use via `export BEAKER_VERSION=git://github.com/...`

branan:
@joshcooper I ran it on a couple of platforms. I also did some searching in both beaker and a couple of acceptance suites for where the `prepend_commands` stuff gets used - AFAICT it's just for the cisco platforms.

joshcooper:
@branan It looks like the Windows Host class implements `prepend_commands`, so I think the changed line will affect every command that beaker executes on that platform (swapping the order)? I'm not sure if any other Host subclass does the same (in addition to Cisco)?

branan:
@joshcooper ugh, you're right. This breaks a couple of windows-y things, where we now try to call `env` from within cmd.exe, instead of the other way around.

I'm not sure there's a solid answer here besides making the prepend_commands interface more complex - we're essentially using it for two very different things between cisco and windows :/

PR #1425, opened 2017-07-07
(BKR-1155) Allow Configurable SSH Method

spjmurray:
The default method of connecting via SSH was updated from IP address to VM hostname.
This had the unfortunate effect of breaking tests on OpenStack hypervisors as where
it used to work first time via IP address it spins for a very long time trying to
connect to a totally made up VM hostname, causing timeouts and failed tests.

This reverts back to IP based addressing by default, however the vmpooler hypervisor
can now implicitly set the host's SSH connection type to vmhostname when not set
explicitly in the host/global configuration.

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -274,7 +274,7 @@ module Beaker
       # create new connection object if necessary
       @connection ||= SshConnection.connect( { :ip => self['ip'], :vmhostname => self['vmhostname'], :hostname => @name },
                                              self['user'],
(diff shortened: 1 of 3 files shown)


spjmurray:
It's up for debate as to whether the change which caused this (a80ab3ee7433f277a284c89b1273428939e37f12) needs reverting back to the old ordering so existing things don't break, and things with weird IP address-changing behaviour use the new configuration option.  I await your suggestions...

rishijavia:
Resolved in #1432

PR #1432, opened 2017-07-17
(BKR-1155) Allow configurable ssh connection preference

rishijavia:
This patch adds a new API method for hypervisors called connections_preference.
Hypervisor authors can overwrite this method to return an array of SSH
connection method preference of their choice when that hypervisor is used. This
option is then added to host_hash.

In host.rb file, we are sending ssh_connection.rb file an object of
:ssh_connection_pereference as an option. And we loop through each of the
element of the array in connect method of ssh_connection file.

The overall idea here is to create an API endpoint for hypervisor authors to set
their connection preference. As of now we dont want end user to set this in
their hosts config as there are no use case for it.

### DIFF
--- a/lib/beaker/hypervisor.rb
+++ b/lib/beaker/hypervisor.rb
@@ -48,7 +48,7 @@ module Beaker
 
       hypervisor = hyper_class.new(hosts_to_provision, options)
       hypervisor.provision if options[:provision]
-
+      self.set_ssh_connection_preference(hosts_to_provision, hypervisor)
       hypervisor
     end
 
@@ -65,6 +65,29 @@ module Beaker
     #Cleanup steps to be run for a given hypervisor.  Default is nil.
     def cleanup
       nil
+    end
+
+    DEFAULT_CONNECTION_PREFERENCE = ['ip', 'vmhostname', 'hostname']
+    #SSH connection method preference. Can be overwritten by hypervisor to change the order
+    def connection_preference
+      DEFAULT_CONNECTION_PREFERENCE
+    end
+
+    #Check if overriding method returns correct array with ip, vmhostname hostname as elements
+    def self.set_ssh_connection_preference(hosts_to_provision, hypervisor)
+      if hypervisor.connection_preference.sort == DEFAULT_CONNECTION_PREFERENCE.sort
+        hosts_to_provision.each{ |h| h[:ssh_connection_preference] = hypervisor.connection_preference}
+      else
+        raise ArgumentError, <<-HEREDOC
+Hypervisor's overriding connection_pereference method is not matching the API.
+
+Make sure your hypervisor's connection_preference returns an array
+containing the following elements in any order you prefer:
+"ip", "hostname", "vmhostname"
+
+Please check hypervisor.rb file's "self.connection_preference" method for an example
+        HEREDOC
+      end
     end
 
     #Proxy package managers on tests hosts created by this hypervisor, runs before validation and configuration.
--- a/lib/beaker/hypervisor.rb
+++ b/lib/beaker/hypervisor.rb
@@ -47,8 +47,8 @@ module Beaker
         end
 
       hypervisor = hyper_class.new(hosts_to_provision, options)
-      hypervisor.provision if options[:provision]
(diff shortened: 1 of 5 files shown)


rishijavia:
discuss this on [jira](https://tickets.puppetlabs.com/browse/BKR-1155)

lib/beaker/hypervisor.rb
```diff
@@ -87,6 +88,19 @@ def cleanup
       nil
     end
 
+    #SSH connection method preference. Can be overwritten by hypervisor to change the order
+    def self.connection_preference
+      ['ip', 'vmhostname', 'hostname']
+    end
+
+    def self.set_ssh_connection_perefence(hosts_to_provision, hypervisor)
+      if hypervisor.connection_preference.sort == self.connection_preference.sort
```
rishijavia: Using this to test if the array returned by the overriding method in hypervisor class matches the API requirement of having the same element name of `["ip", "vmhostname", "hostname"]`, order does not matter.

I know sorting would bring atleast `n log n` runtime debt, is there any better way to do the same?

kevpl: @rishijavia beaker by default does this twice in an empty run, and even in test suites with hundreds of tests, I don't see it getting called more than 10s of times, so I'm not worried about asymptotic behavior about this.

lib/beaker/hypervisor.rb
```diff
@@ -87,6 +88,19 @@ def cleanup
...
+        hosts_to_provision.each{ |h| h[:ssh_connection_preference] = hypervisor.connection_preference}
+      else
+        hosts_to_provision.each{ |h| h[:ssh_connection_preference] = self.connection_preference}
```
kevpl: @rishijavia rather than defaulting here, since it's hypervisor implementors who should be hitting this, not even beaker users, this should fail hard if the hypervisor implementor went to the trouble of specifying their own ordering rather than using our default, but then got it wrong.

lib/beaker/ssh_connection.rb
```diff
@@ -66,20 +67,12 @@ def connect_block host, user, ssh_opts
...
     def connect
       #try three ways to connect to host (vmhostname, ip, hostname)
-      methods = []
-      if @vmhostname
-        @ssh ||= connect_block(@vmhostname, @user, @ssh_opts)
-        methods << "vmhostname (#{@vmhostname})"
+      # Try each method in turn until we succeed
+      methods = @ssh_connection_preference.dup
+      while @ssh.nil? && (not methods.empty?) do
+        @ssh ||= connect_block(instance_variable_get("@#{methods.shift}"), @user, @ssh_opts)
```
kevpl: @rishijavia `methods` will be empty here, using the original `@ssh_connection_preference` would be better.

lib/beaker/hypervisor.rb
```diff
@@ -87,6 +88,19 @@ def cleanup
       nil
     end
 
+    #SSH connection method preference. Can be overwritten by hypervisor to change the order
+    def self.connection_preference
+      ['ip', 'vmhostname', 'hostname']
+    end
+
+    def self.set_ssh_connection_perefence(hosts_to_provision, hypervisor)
```
kevpl: @rishijavia preference spelling

lib/beaker/hypervisor.rb
```diff
@@ -68,7 +68,8 @@ def self.create(type, hosts_to_provision, options)
 
       hypervisor = hyper_class.new(hosts_to_provision, options)
       hypervisor.provision if options[:provision]
-
+      self.set_ssh_connection_perefence(hosts_to_provision, hypervisor)
```
kevpl: @rishijavia preference spelling

lib/beaker/ssh_connection.rb
rishijavia: @kevpl you meant this on line 76, and i did amended and changed `methods` to `@ssh_connection_preference`

rishijavia:

rishijavia:

rishijavia:

rishijavia:

PR #1433, opened 2017-07-18
(BKR-1151) Open port 8142 on master node and 8143 for split install

rishijavia:


lib/beaker/hypervisor/aws_sdk.rb
```diff
@@ -985,5 +986,16 @@ def load_fog_credentials(dot_fog = '.fog')
         :secret_key => default[:aws_secret_access_key],
       }
     end
+
+    # Adds port 8143 to host[:additional_ports]
+    # if master, database and dashboard are not on same instance
+    def test_split_install
+      @hosts.each do |host|
+        mono_roles = ['master', 'database', 'dashboard']
+        if (host[:roles] & mono_roles).size != 3 and (host[:roles] & mono_roles).any?
```
doug-rosser: Not sure if the test logic:

`(host[:roles] & mono_roles).size != 3`

is appropriate. The number of roles can very beyond 3, as shown at https://gist.github.com/doug-rosser/3c32a89af904d8f4dbb7bafc5103c548#file-example-aws-ec2-config-L46

kevpl: @rishijavia better for readability (& maintainability):
```ruby
roles_intersection = host[:roles] & mono_roles
if roles_intersection.size != 3 and roles_intersection.any?
```

rishijavia: @doug-rosser the bitwise `&` of `host[:roles]` with `mono_roles` will bring down the size of the array to 3 if master, dashboard and database is provided on the same host. And if that's the case, it's size will also be 3 which would fail the first part of the if statement and therefore `8143` will not be open on that host.

PR #1434, opened 2017-07-19
(BKR-437) Correctly detect public/private keys

hunner:
The private_key() method was previously wrong. It used the last key
found in a loop, but did not verify that both the public and private key
exists for the chosen key before continuing. This meant that the ssh key
specified in the hosts file was ignored.

### DIFF
--- a/lib/beaker/hypervisor/aws_sdk.rb
+++ b/lib/beaker/hypervisor/aws_sdk.rb
@@ -688,10 +688,9 @@ module Beaker
       keys = Array(@options[:ssh][:keys])
       keys << '~/.ssh/id_rsa'
       keys << '~/.ssh/id_dsa'
-      key_file = nil
-      keys.each do |key|
-        key_filename = File.expand_path(key + '.pub')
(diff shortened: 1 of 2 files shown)


lib/beaker/hypervisor/aws_sdk.rb
```diff
@@ -688,18 +688,21 @@ def public_key
...
       keys << '~/.ssh/id_rsa'
       keys << '~/.ssh/id_dsa'
-      key_file = nil
-      keys.each do |key|
-        key_filename = File.expand_path(key + '.pub')
-        key_file = key_filename if File.exists?(key_filename)
+      key_file = keys.find do |key|
+        key_pub = key + '.pub'
+        if File.exist?(File.expand_path(key_pub)) && File.exist?(File.expand_path(key))
+          true
```
rishijavia: @hunner, you can just use this if statement as a standalone. You don't need to return `true` or `false`.

hunner: Oh right...

PR #1436, opened 2017-07-20
(BKR-1084) Add --debug-errors option to enter pry on a test failing

james-stocks:
Adds a --debug-errors option. When used, beaker will drop the runner into a pry console if/when a test fails or errors.
This is intended for the beaker run subcommand.

### DIFF
--- a/beaker.gemspec
+++ b/beaker.gemspec
@@ -24,7 +24,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'rspec-its'
   s.add_development_dependency 'fakefs', '~> 0.6'
   s.add_development_dependency 'simplecov'
-  s.add_development_dependency 'pry', '~> 0.10'
   s.add_development_dependency 'rake', '~> 10.0'
 
   # Documentation dependencies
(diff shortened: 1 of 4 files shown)


lib/beaker/options/command_line_parser.rb
```diff
@@ -88,6 +88,12 @@ def initialize
             @cmd_options[:preserve_hosts] = mode || 'always'
           end
 
+          opts.on '--debug-errors',
```
james-stocks: While I only want this option to be on the _run_ subcommand; beaker will raise an exception if I pass a command through to beaker that's not declared here.
I could set the hide attribute on this option so it doesn't appear in beaker --help

james-stocks:
@kevpl - after reviewing with @johnduarte , we thought that if we had a debugger option like this, then we would not need the previously PR'ed --skip-last-teardown option - you can use this option instead to keep running beaker until it fails, and then inspect the ruby code and the SUT state at the time of failure.

I think this also means we don't _need_ the other PR to expand the debugging doc with byebug instructions; but we could still keep those doc changes as they should be useful general ruby debugging information

james-stocks:
Tests appear to be failing because we don't yet have pry-byebug on rubygems.delivery.puppetlabs.net - I can fix that if the general approach to this PR gets a thumbs up

james-stocks:
@puppetlabs-jenkins please re-test this

james-stocks:

james-stocks:
[Latest test failure](https://jenkins-beaker.delivery.puppetlabs.net/job/qe_beaker_intn-sys_vpool-beaker-rspec-support/layout=64mdac,platform=centos7/2529/console) is due to vmpooler capacity.

ferglor:

beaker.gemspec
```diff
@@ -33,6 +32,9 @@ Gem::Specification.new do |s|
   # Run time dependencies
   s.add_runtime_dependency 'minitest', '~> 5.4'
   s.add_runtime_dependency 'minitar', '~> 0.6'
+  s.add_runtime_dependency 'pry-byebug', '~> 3.4.2'
```
james-stocks: This needs to be runtime because the users of beaker will get it as a bundled gem ; and I don't know (and don't see obvious [documentation|http://bundler.io/v1.0/man/gemfile.5.html] for) whether bundler allows adding a gem with its dev dependencies.

PR #1442, opened 2017-07-28
(BKR-1155) Allow hypervisor to set ssh connection methods order

rishijavia:
There was ongoing conflict between difference ssh connection method preferences
among different hypervisors. VMPooler connects with vmhostname(dns) better than
ip address. Slice uses ip addresses only.

Rather than having one hardcoded connection methods preference (order), we
opened up beaker API to allow hypervisor authors to give their own preference.
Hypervisor authors just have to override connection_preference method. We now
also allow it to be set in hosts file.

Hypervisor's preference will take precedence over the default and
if specified in hosts file, that preference will take precedence over hypervisor.


kevpl:
I'm 👍  on this change as-is right now, but we shouldn't merge it until we can verify that what happened with [beaker-abs](https://github.com/puppetlabs/beaker-abs) won't take place during the next release, which will happen in the work ticketed in [BKR-1177](https://tickets.puppetlabs.com/browse/BKR-1177). This PR shouldn't be merged until that work is done.

kevpl:
Closing in favor of #1448

PR #1443, opened 2017-07-31
(BKR-1163) update minitar to newest version

DavidS:
minitar 0.5.4 has a security issue in some code beaker doesn't use. Still,
updating will avoid this to be flagged on security scans.


Iristyle:
@DavidS there's already a ticket for this - could you please update your commit accordingly? Thanks!

https://tickets.puppetlabs.com/browse/BKR-1163

DavidS:
Updated as requested.

PR #1444, opened 2017-07-31
(BKR-1132) adds concept of manual_test and manual_step

Andrew-Hayes:
Adds an option --exec-manual-tests

adds manual_tests. This allows use to do some automated setup for manual
test cases, but also skips them when run in CI

adds manual_step. This allows us to pass or fail a test after prompting
the tester

### DIFF
--- a/lib/beaker/dsl/structure.rb
+++ b/lib/beaker/dsl/structure.rb
@@ -54,6 +54,63 @@ module Beaker
         end
       end
 
+      # Provides a method to help manual tests. So we can use beaker to set up
+      # the environment, then prompt a user to manually check the setup.
+      # @param [String] step_name The name of the step to be logged.
+      def manual_step step_name
+        require 'readline'
+        logger.notify "\n* #{step_name}\n"
+        if(!@options.has_key?(:exec_manual_tests))
+          # if the option -exec-manual-tests is not set then this has executed outside of a manual tests
+          # so we raise an error to avoid issues
+          raise('--exec-manual-tests option not set, this means a manual_step was used outside a manual_test')
+        end
+
+        set_current_step_name(step_name)
+        # Here we prompt the user to tell us if the step passed or failed
+        loop do 
+          input = Readline.readline('Did this step pass, Y/n? ', true).squeeze(" ").strip.downcase
+          if %w(y yes).include?(input)
+            break
+          elsif %w(n no).include?(input)
+            # if the step failed, the user can enter a fail message.
+            # we loops to ensure they give use a fail message
+            fail_message = ''
+            loop do
+              fail_message = Readline.readline('What was the reason for failure? ', true).squeeze(" ").strip
+              if fail_message == ''
+                # if nothing is entered we tell the user to enter something
+                puts "No reason for failure given, please enter reason for failure."
+              else
+                break
+              end
+            end
+            raise Beaker::DSL::FailTest, fail_message
+          else
+            # if something other than Y or n is returned we ask them again
+            puts "Please enter Y or n."
+          end
+        end
+      end
+
+      # Provides a method to mark manual tests.
+      # If the --exec-manual-tests param is not set then we skip the test
+      # this is so manual tests do not execute by mistake
+      # @param [String] manual_test_name The name of the test to be logged.
+      # @param [Proc] block The actions to be performed during this test.
+      #
+      def manual_test manual_test_name, &block
+        if(@options.has_key?(:exec_manual_tests) && @options[:exec_manual_tests] == true)
+          # here the option is set so we run the test as normal
+          test_name manual_test_name, &block 
+        else
+          # here no option was set so we log the test name and skip it
+          test_name manual_test_name
+          raise( Beaker::DSL::SkipTest,
+                 '--exec-manual-tests option not set, so skipping manual test' )
+        end
+      end
+
       # Provides a method to name tests.
       #
       # @param [String] my_name The name of the test to be logged.
--- a/lib/beaker/options/command_line_parser.rb
+++ b/lib/beaker/options/command_line_parser.rb
@@ -94,6 +94,12 @@ module Beaker
             @cmd_options[:debug_errors] = bool
           end
 
+          opts.on '--exec-manual-tests',
+                  'Execute manual tests',
+                  '(default: false)' do |bool|
+            @cmd_options[:exec_manual_tests] = bool
+          end
+
           opts.on '--root-keys',
                   'Install puppetlabs pubkeys for superuser',
                   '(default: false)' do |bool|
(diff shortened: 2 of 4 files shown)


james-stocks:
LGTM, yardoc failures need fixed before this can be merged.

Two comments:
 * Should this be in subcommands and not the main beaker program? I think it should be in the main beaker execution - I think users of this will run an entire beaker suite from their workstation. They won't necessarily want to pick out particular subcommands.
 * This deserves a how_to document; but I think time should only be spent on that after the feature is approved

kevpl:
what kinds of things need manual verification, requiring a person to say `y/n` in a beaker test?

james-stocks:
@kevpl - we forwarded you an email with some internal discussion for context.
To record some of the context in this PR as well, a couple of categories of tests we'd use this for are:
 * Tests that are temporarily manual because they don't get priority to be implemented as automated tests before the product release they're first needed for.
 * Visual tests that need human eyes. These can eventually be automated, there's tools for this; but we have product releases we need to do before we can set that tooling up.

Andrew-Hayes:
The checks seem to have passed. Except for an unrelated windows issue.

ferglor:

lib/beaker/options/command_line_parser.rb
```diff
@@ -94,6 +94,12 @@ def initialize
             @cmd_options[:debug_errors] = bool
           end
 
+          opts.on '--exec-manual-tests',
```
tvpartytonight: In the current implementation of subcommands, this option will need to be copied over the `beaker test run` group.

lib/beaker/dsl/structure.rb
```diff
@@ -1,4 +1,5 @@
 require 'beaker/dsl/assertions'
+require 'readline'
```
kevpl: @Andrew-Hayes we don't want stray `require` calls including code that oftentimes won't be used on a particular run.

Would you move this require inside the `manual_step` method?

lib/beaker/dsl/structure.rb
```diff
@@ -54,6 +55,51 @@ def step step_name, &block
...
+        end
+
+        set_current_step_name(step_name)
+        # Here we prompt the user to tell us if the step passed or failed
+        loop do 
+          input = Readline.readline('did this step pass, Y/n?', true).squeeze(" ").strip
+          if input == 'Y'
+            break
+          elsif input == 'n'
+            raise Beaker::DSL::FailTest, fail_message
```
kevpl: @Andrew-Hayes we were debating whether or not having a `binding.pry` here would be better than this process, & I think we ended up being OK with using `readline` instead, but we're lacking some context to be confident of this.

It did seem to us that if you had a lot of testing to run, and some-odd number of manual tests to run as a part of that where you had to do visual comparison for verification, it could be difficult to juggle all of the reasons that things failed in your head as failures pile up. We were thinking a fail workflow like this might be better:
```
did this step pass, Y/n? n
default failure message: <fail_message argument>
to use this failure message, press <char>. Otherwise, enter a new failure message:
```
before, we'd talked about how juggling code to change vs code to test can be a pain for QA (I assume the `fail_message` parameter can fit in this boat).

Is this a case that comes up for you guys? We're curious about what your manual testing environment looks like, so that we can better match our reviews up with what you're really doing.

ferglor:
last comment still current

lib/beaker/dsl/structure.rb
```diff
@@ -54,6 +54,57 @@ def step step_name, &block
...
+        # Here we prompt the user to tell us if the step passed or failed
+        loop do 
+          input = Readline.readline('Did this step pass, Y/n?', true).squeeze(" ").strip
+          if input == 'Y'
+            break
+          elsif input == 'n'
+            # if the step failed, the user can enter a fail message.
+            fail_message = Readline.readline('What was the reason for failure? [Manual step failed]: ', true).squeeze(" ").strip
+            if fail_message == ""
+              fail_message = 'Manual step failed'
```
james-stocks: I'd lean towards making the tester enter something descriptive

lib/beaker/dsl/structure.rb
```diff
@@ -54,6 +54,57 @@ def step step_name, &block
...
+          # if the option -exec-manual-tests is not set then this has executed outside of a manual tests
+          # so we raise an error to avoid issues
+          raise('--exec-manual-tests option not set, this means a manual_step was used outside a manual_test')
+        end
+
+        set_current_step_name(step_name)
+        # Here we prompt the user to tell us if the step passed or failed
+        loop do 
+          input = Readline.readline('Did this step pass, Y/n?', true).squeeze(" ").strip
+          if input == 'Y'
```
james-stocks: Or 'y' and maybe allows begin_with?

lib/beaker/dsl/structure.rb
```diff
@@ -54,6 +54,57 @@ def step step_name, &block
         end
       end
 
+      # Provides a method to help manual tests. So we can use beaker to set up
+      # the environment, then prompt a user to manually check the setup.
+      # @param [String] step_name The name of the step to be logged.
+      # @param [String] fail_message The message to return if this manual step failed
```
james-stocks: delet this

lib/beaker/dsl/structure.rb
james-stocks: I think the environment/context for running these will be executing beaker on a laptop/workstation; using rake tasks that are as close as possible to the rake tasks that run the executed tests in CI (use the same vmpooler hosts, use the same package installation, run some or all of the automated tests as well as these special manual tests).

We haven't considered results storage - the workflow for manual testing today is that we generally do not store manual test results; we file tickets for anything that's not right; and everything else has implicitly passed. Maybe in future we would want a results store.

james-stocks: Regarding pry; that would be a good subsequent update to this. It wouldn't be useful in all cases; but it would be a handy option for immediate investigation.
You can probably also mix this new option with the --debug-errors options and mark this test failed then immediately drop into pry.

james-stocks:
@kevpl - one of your comments is hidden against an old commit, but is still valid.

You asked about the environment for running these:

I think the environment/context for running these will be executing beaker on a laptop/workstation; using rake tasks that are as close as possible to the rake tasks that run the executed tests in CI (use the same vmpooler hosts, use the same package installation, run some or all of the automated tests as well as these special manual tests).

We haven't considered results storage - the workflow for manual testing today is that we generally do not store manual test results; we file tickets for anything that's not right; and everything else has implicitly passed. Maybe in future we would want a results store.

james-stocks:
Regarding the suggestion to make pry available at this point: that would be a good subsequent update to this. It wouldn't be useful in all cases; but it would be a handy option for immediate investigation.
You can also mix this new option with the --debug-errors options and mark this test failed then immediately drop into pry. Though it may be nice to have the option to debug an _unexpected_ failure when you did not expect to have to set the --debug-errors option

lib/beaker/options/command_line_parser.rb
Andrew-Hayes: Done :-)

PR #1446, opened 2017-08-03
(BKR-1184) Remove Nokogiri from beaker

rishijavia:


er0ck:
fyi @zreichert, @samwoods1 we should probably evaluate this and ensure it's not gonna break qaelk.  last i checked (a while back) there wasn't much testing around junit.xml output and its schema

lib/beaker/logger_junit.rb
```diff
@@ -1,22 +1,7 @@
...
-    # writes the xml created in the block to the xml file given
-    #
-    # Note: Error Recovery should take place in the caller of this
-    # method in order to recover gracefully
-    #
-    # @param [String] xml_file Path to the xml file
-    # @param [String] stylesheet Path to the stylesheet file
-    # @param [Proc]   block XML message construction block
-    #
-    # @return nil
```
er0ck: why are the yarddocs being removed here?

lib/beaker/logger_junit.rb
```diff
@@ -72,51 +46,48 @@ def self.copy_stylesheet_into_xml_dir(stylesheet, xml_file)
...
+      else
+        #no existing file, create a new one
+        doc           = REXML::Document.new
+        doc << REXML::XMLDecl.new(version="1.0", encoding="UTF-8")
+        instruction_content = "type='text/xsl' href='#{File.basename(stylesheet)}'"
+        doc << REXML::Instruction.new(target="xml-stylesheet", content=instruction_content)
+      end
+      return doc
+    end
+
```
er0ck: why is this method being moved?  it makes the diff a bit hard for me to follow

lib/beaker/test_suite_result.rb
```diff
@@ -0,0 +1,246 @@
...
+            end
+          end
+        end
+      rescue Exception => e
+        @logger.error "failure in XML output: \n#{e.to_s}" + e.backtrace.join("\n")
+      end
+    end
+
+  end
+end
```
er0ck: i can see why we'd want to move this class for better file layout, etc.  

if it were in a separate commit it would make the diffs easier to read.

samwoods1:
Yeah, the specs are covering individual methods... It seems like an acceptance test that would compare the junit.xml from a simple test suite execution to a resource with known good junit.xml would be useful to validate this change, and protect against potential changes to the junit.xml output in the future.

rishijavia:
@er0ck, I made a different file to implement the functionality of `logger_junit.rb` using REXML. I started everything from scratch there and therefore missed documentation on some methods and disordered them. 

Updated my commit to reflect the order of methods to match what we have in the original file. And added documentation at missing spots too.

lib/beaker/test_suite_result.rb
rishijavia: @er0ck, yes there are bunch of changes that are made in `write_junit_xml` method. It was using nokogiri to create elements and attributes, but are is now changed to use REXML.

But there are no changes anywhere else apart from that one method.

er0ck: in thinking through this, for my own info, if the changes were made after the file move, in its own commit, then we could just view that single commit diff to see the changes and github wouldn't overlay the file move and not show us the individual char diffs.

but i can probably diff these locally myself

lib/beaker/test_suite_result.rb
```diff
@@ -0,0 +1,259 @@
...
+            item.add_attributes(
+              [
+                ['classname', File.dirname(test.path)],
+                ['name', File.basename(test.path)],
+                ['time', "%f" % test.runtime]
+              ])
+
+            test.exports.each do |export|
+              export.keys.each do |key|
+                item.add_attribute(key.to_s.tr(" ", "_"), export[key])
```
rishijavia: this is the only functionality change

Before this we were using export key as an xml attribute and export value as the value for that attribute. For example:
if export key is `middle earth` and value is `Elves`

we were creating xml like:
`<element middle earth="Elves"> </element>`

This is invalid XML. You cannot have whitespaces in attribute name. [W3 docs](https://www.w3.org/TR/2008/REC-xml-20081126/#NT-AttType). So we are now replacing spaces with `_`. So now the xml would look like:

`<element middle_earth="Elves"> </element>`

I'm open to other suggestions to create valid XML!

er0ck: i better be thumbs up with this.  you are fixing my change 👼 
https://github.com/puppetlabs/beaker/blame/master/lib/beaker/test_suite.rb#L213

rishijavia:
@er0ck @kevpl, changed commits to have better diffs

rishijavia:

rishijavia:
I also made a [repo](https://github.com/rishijavia/beaker-junitxml) with different folders. Each folder has a `command.md` file that shows the command ran to generate the `beaker_junit.xml` file. If you go look the the commit history, there is a commit that shows the file [generated by nokogiri](https://github.com/rishijavia/beaker-junitxml/commit/40a54b6767a91b8031b634861167fc1012bc26f3) and the same file [generated by REXML](https://github.com/rishijavia/beaker-junitxml/commit/57f59d6caf2840a641bc96dbdb2e58569a37adbe).

Each folder also has a `diff` file that shows difference between the two XML files.

kevpl:
@er0ck @samwoods1 it looks like @rishijavia has updated all of the conversations you were having. Please take a look at his comments & the diffs in the above comment, & let us know if you have other feedback or comments. Thanks!

lib/beaker/logger_junit.rb
```diff
@@ -102,21 +102,21 @@ def self.get_testsuites_from_doc(doc, name, already_existed)
...
+        doc           = REXML::Document.new
+        doc << REXML::XMLDecl.new(version="1.0", encoding="UTF-8")
+        instruction_content = "type='text/xsl' href='#{File.basename(stylesheet)}'"
+        doc << REXML::Instruction.new(target="xml-stylesheet", content=instruction_content)
       end
       return doc
     end
 
-    # Remove color codes and invalid XML characters from provided string
+ # Remove color codes and invalid XML characters from provided string
```
er0ck: @rishijavia indenting change, here, probably unintended.

er0ck:
yep.  i'm 👍 after the small indent change is fixed up.

rishijavia:
updated based on review :+1:

PR #1457, opened 2017-09-19
make apt-get install non-interactive

DarkGigaByte:
I've been waiting ages for install_puppet_agent_on to finish, until I realized that it had hung up on apt-get install because of an already existing config file which should be updated but had conflicts. It was stuck on the interactive dialog...

### DIFF
--- a/lib/beaker/host/unix/pkg.rb
+++ b/lib/beaker/host/unix/pkg.rb
@@ -92,7 +92,7 @@ module Unix::Pkg
           name = "#{name}=#{version}"
         end
         update_apt_if_needed
-        execute("apt-get install --force-yes #{cmdline_args} -y #{name}", opts)
+        execute("DEBIAN_FRONTEND=noninteractive apt-get install --force-yes #{cmdline_args} -y #{name}", opts)
       when /solaris-11/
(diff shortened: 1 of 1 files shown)


kevpl:
Hi @DarkGigaByte & thanks for the fix! If you would please add a spec test so that we can ensure we don't break this in the future & make the commit changes listed in our [CONTRIBUTING doc](https://github.com/puppetlabs/beaker/blob/master/CONTRIBUTING.md) (mostly just prefixing your git commit with `(MAINT)`), hopefully we'll be able to get this reviewed, merged, & released quickly!

mchllweeks:
Last comment still current.

DarkGigaByte:
Won't have the time in the next couple weeks. Maybe someone else will do it in the meantime?

ferglor:
Last comment still current.

kevpl:
last comment is still current. Two things to note:

- our Contributing standards have been updated, so this change will need a CHANGELOG entry as well
- if this isn't updated in a week, we'll close it as inactive

If you don't have the time to come back to this before then, feel free to close the PR yourself for now. It's not like there's any negative consequences to you either re-opening the PR later with tests or opening a new one for this.

PR #1458, opened 2017-09-19
Adding default to read fog credentials

cdenneen:
@tvpartytonight Added to presets to read from fog credentials in order to disable in beaker-aws

### DIFF
--- a/lib/beaker/options/presets.rb
+++ b/lib/beaker/options/presets.rb
@@ -193,6 +193,7 @@ module Beaker
           :cache_files_locally    => false,
           :aws_keyname_modifier   => rand(10 ** 10).to_s.rjust(10,'0'), # 10 digit random number string
           :run_in_parallel        => [],
+          :use_fog_credentials    => true,
           :ssh                    => {
                                      :config                => false,
                                      :paranoid              => false,


lib/beaker/options/presets.rb
```diff
@@ -38,6 +38,7 @@ class Presets
         :test_tag_or          => ['BEAKER_TEST_TAG_OR'],
         :test_tag_exclude     => ['BEAKER_EXCLUDE_TAG', 'BEAKER_TEST_TAG_EXCLUDE'],
         :run_in_parallel      => ['BEAKER_RUN_IN_PARALLEL'],
+        :use_fog_credentials  => 'true',
```
tvpartytonight: @cdenneen this should be added to the list of options in the `presets` method down below in this file. Additionally, it should be just `true`, not `'true'`. 

Thanks!

cdenneen:
@tvpartytonight I originally tried boolean on line 41 but that doesn't work (fails all rspec tests) I believe it's explicitly looking for strings for everything.
I can add to line 197 below and think the boolean will work there.
Feel free to try changing 41 to a boolean to see the errors that will occur.

tvpartytonight:
@cdenneen it's explicitly looking for strings because it is loading it from the environment; the presets method is a hash, so you should have no trouble loading it in as a `true` class.

tvpartytonight:
@cdenneen last comment still current.

kevpl:
last comment still current.

cdenneen:
@tvpartytonight @kevpl not following ... this is bool in latest commit.

lib/beaker/options/presets.rb
cdenneen: Done

kevpl: @cdenneen it looks like it's still a string here, is there a reason for that/

cdenneen: @kevpl ENV spec requires strings (doesn't work with bool).
The real preset is a bool:

https://github.com/cdenneen/beaker/blob/1b7ff68dab806dce9ebfce928e3e42f2dd8f1681/lib/beaker/options/presets.rb#L197

cdenneen: would it be better if I just removed it from there or should we be adding a BEAKER_USE_FOG ?

PR #1463, opened 2017-10-10
Allow nodeset YAML files to contain ERB

trevor-vaughan:
This allows users to embed ERB content into their nodeset YAML files.
This is particularly important when doing Docker testing on a shared
environment since the nodes cannot have overlapping names.

After thinking about a lot of different potential solutions, this one
appeared to be the most flexible and allow users to do what they needed
to in a language that is understood to Puppet users.

If no ERB content is present, there will be no change noticed by
current users.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,3 +1,4 @@
 --format documentation
 --color
 --tty
+--fail-fast
--- a/.rspec
+++ b/.rspec
@@ -1,4 +1,3 @@
 --format documentation
 --color
 --tty
---fail-fast
--- a/lib/beaker/version.rb
+++ b/lib/beaker/version.rb
@@ -1,5 +1,5 @@
 module Beaker
   module Version
-    STRING = '3.27.0'
+    STRING = '3.26.0'
   end
 end
--- a/lib/beaker/version.rb
+++ b/lib/beaker/version.rb
@@ -1,5 +1,5 @@
 module Beaker
   module Version
-    STRING = '3.26.0'
+    STRING = '3.27.0'
   end
 end
--- a/lib/beaker/options/hosts_file_parser.rb
+++ b/lib/beaker/options/hosts_file_parser.rb
@@ -14,12 +14,17 @@ module Beaker
       # @raise [ArgumentError] Raises if hosts_file_path is not a valid YAML file
       # @raise [Errno::ENOENT] File not found error: hosts_file doesn't exist
       def self.parse_hosts_file(hosts_file_path = nil)
+        require 'erb'
+
         host_options = new_host_options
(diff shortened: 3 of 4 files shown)


.rspec
```diff
@@ -1,3 +1,4 @@
 --format documentation
 --color
 --tty
+--fail-fast
```
kevpl: please remove `fail-fast` from this PR as it's not what we want when running rspec tests

lib/beaker/version.rb
```diff
@@ -1,5 +1,5 @@
 module Beaker
   module Version
-    STRING = '3.26.0'
+    STRING = '3.27.0'
```
kevpl: please remove version string changes from this PR. Those are handled in our releases

trevor-vaughan:
@kevpl The 'for instance' is pretty easy!

Go run a Beaker test with docker in GitLab...then go use another PR to run the same module in parallel on the same environment. 💥 Instant death.

Or have a Beaker run with docker fail 💥 instant death with manual cleanup needed. (I'd like my CI system to be able to continue doing CI even when things are a bit of a mess).

Another 'for instance': I want to use my AWS credentials the way I usually do (via environment variables) and I don't want to go hack the AWS backend to Beaker. Hack the `nodeset` to read those variables in the appropriate place and...instant win.

I'm not asking you to make it more complicated, I'm asking it to support being more complicated if the *user* needs it. Otherwise, we end up having to write crazy nodeset generator code and wrap beaker in other scripts just to twiddle some variables. The CI system gets more complicated and it makes more work for everyone.

I'm also not recommending that everyone use these capabilities in nodesets, I just want the ability to do so when it makes sense (like when running docker containers on a shared/CI system).

op-ct:
@trevor-vaughan @kevpl I can see a lot of handy use cases for ERB nodesets:

* keeping secrets out of the CONFIG sections
* nodesets that are portable between different environments
* ENV-based overrides

trevor-vaughan:
@kevpl Are the changes OK now?

trevor-vaughan:
As an additional note, I think we might be hitting this on Travis CI when using Beaker to drive Docker. It's evidencing as random test failures but the only way I've gotten it to repeat outside of Travis is to have a run fail in the middle and not clean up my container names properly. My guess is that they're not getting reaped right away and we're hitting a duplicate name error out on Travis.

If I could detect that in my nodeset that I'm using on Travis runs, I could smack a unique ID in there and work around the issue.

mchllweeks:
retest this please

kevpl:
@trevor-vaughan last message is still current. If no progress is made on this PR in a week, it will be closed due to inactivity.

PR #1464, opened 2017-10-11
(maint) Allow fine-grained, structured timing for the tests

ekinanp:
Previously, Beaker stored information for the duration of a single
test case. However, there was no way to obtain the duration of each
piece of the test (e.g. how long a specific step or sub-test took).
This commit adds this capability -- refer to the comments above the
"timed_node" instance in the TestCase class for more details on how
it works/is stored. Note that the timing specific information is
written to a single file in the same directory as the existing
summary.txt files under the basename #{suite}-times.txt.

This commit allows for the possibility of more fine-grained metrics
tracking (e.g. which step in a test failed, which sub-test failed,
etc.) in the future by extending the timed_node object to instead
represent an AST for the test case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ git logs & PR history.
 
 # [Unreleased](https://github.com/puppetlabs/beaker/compare/3.26.0...master)
 
+### Added
+
+- fine-grained, structured timing for tests
+
 # [3.26.0](https://github.com/puppetlabs/beaker/compare/3.25.0...3.26.0) - 2017-10-05
 
 ### Added
--- a/beaker.gemspec
+++ b/beaker.gemspec
@@ -45,6 +45,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'open_uri_redirections', '~> 0.2.1'
   s.add_runtime_dependency 'in-parallel', '~> 0.1'
   s.add_runtime_dependency 'thor', '~> 0.19'
+  s.add_runtime_dependency 'rubytree', '~> 0.9.2'
 
   # Run time dependencies that are Beaker libraries
   s.add_runtime_dependency 'stringify-hash', '~> 0.0'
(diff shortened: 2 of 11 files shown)


ekinanp:
I could not test this locally to see if the "#{name}-times.txt" file is actually produced because "bin/beaker" requires the beaker gem itself. If anyone knows of a way to do this, please let me know and I'll test it out else I'll try to do so later today.

MikaelSmith:
You can use a local install of beaker via https://github.com/puppetlabs/puppet/blob/master/acceptance/Gemfile#L8, i.e. set `GEM_SOURCE` to `file:///path/to/beaker/dir`.

ekinanp:
Ok, tested this locally. The right "#{name}-times.txt" file is generated with the expected structure, and none of this information is printed to STDOUT. So the changes here should be non-invasive to existing beaker workflows (outside of the "#{names}-times.txt" file being generated).

tvpartytonight:
@ekinanp thanks for your PR; one thing that I would like to see is a README documenting how to use the timing file created by this change. Additionally, beaker's contributing standards have changed, and we would like people to contribute to the [changelog](https://github.com/puppetlabs/beaker/blob/master/CHANGELOG.md) when adding new functionality in their PR.

ekinanp:
@tvpartytonight I added a README and updated the CHANGELOG

CHANGELOG.md
```diff
@@ -19,5 +19,7 @@ git logs & PR history.
 
 - concept of `manual_test` and `manual_step`
```
mchllweeks: Please put your changes below Unreleased with an Added section.  So it should look like:

# [Unreleased](https://github.com/puppetlabs/beaker/compare/3.26.0...master)

### Added

- fine-grained, structured timing for tests

tvpartytonight:
@ekinanp one stylistic issue with this PR is the usage of `test_name` nested in your examples; [beaker's style guide](https://github.com/puppetlabs/beaker/blob/master/docs/concepts/style_guide.md#test-naming) specifies that tests should have one `test_name` block per file. Could update your PR to be consistent with this style?

ekinanp:
@mchllweeks Updated the CHANGELOG.

@tvpartytonight I assume you mean the doc? If so, I have updated the examples there to conform to the beaker style guide.

ferglor:
retest this please

tvpartytonight:
@ekinanp and I discussed just leaving functionality to be implemented on a per step/test basis, instead of adding it to beaker directly. Closing in favor of that route for timing steps.

PR #1454, opened 2017-08-25
(BKR-1168) Fix the `beaker --version` undefined method `each_key' for nil:NilClass

raffs:
The issue was happing because beaker will run the function that needs
the HOSTS arrays, which is defined as null when running `beaker --version`,
therefore ruby thrown noMethod for nilClass.

This commit add an expression to return the code when using '--version'
option, instead of continuing running the beaker code.

PR #1465, opened 2017-10-17
(BKR-1168) beaker version causes error output

raffs:
The issue was happening because beaker will run the function that needs the HOSTS array, which is define as null when running beaker --version, therefore ruby thrown _noMethod for nilClass_.

This commit add an expression to return the code when using **--version** option, instead of continuing running the beaker code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,5 +26,9 @@ git logs & PR history.
 
 - concept of `manual_test` and `manual_step`
 
+### Bug fixes
+
+- Fixed corruption of `opts[:ignore]` when using `rsync`
+
 # [3.25.0](https://github.com/puppetlabs/beaker/compare/3.24.0...3.25.0) - 2017-09-26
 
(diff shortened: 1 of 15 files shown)


raffs:
@kevpl as we talked on the other PR, I create another branch and cherry-pick the commits from the previous PR related to this issue/ticket. Please let me know about anything else. 

Previous PR: https://github.com/puppetlabs/beaker/pull/1454

kevpl:
@raffs thanks for fixing this up! Please add a spec test to make sure that we don't break this in the future, & we'll make sure this gets in soon!

mchllweeks:
last comment still current

kevpl:
@raffs it looks like a bunch of commits have been added to this PR, I'm having trouble seeing if any of them are the tests we're looking for. Since that's the case, we'll treat this as if it hasn't been updated:

Last comment is still current. If no progress is made on this in a week, we'll close it due to inactivity.

ferglor:
Closing due to inactivity

PR #1466, opened 2017-10-17
(QENG-5106) Update Beaker to support amazon as a platform

mchllweeks:
We have converted two pipelines that require AWS resources to use ABS.  The pipelines previously used redhat as a platform when calling beaker.  Now that they will use ABS to acquire resources, the platform will be amazon which beaker did not yet support.


mchllweeks:
retest this please

mchllweeks:
retest this please

tvpartytonight:
👍

PR #1468, opened 2017-10-18
(BKR-1203) Add codenames for MacOS 10.13 and Ubuntu Artful

caseywilliams:


tvpartytonight:

johnduarte:
The Jenkins run failed to provision an Ubuntu 12.04 instance. This OS is EOL. Should it be removed from this test suite @tvpartytonight ?

tvpartytonight:
@johnduarte I've removed those platforms from testing, and since that was the only error, this is good to go.

PR #1470, opened 2017-11-07
(maint) Update the Ruby install of the pre-suite to force install bundler

ferglor:
When a new VM is checked out from the vmpooler, bundler is already installed.
This bundler installation causes a conflict when the beaker acceptance tests
attempt to install bundler, which in turn, causes the acceptance tests to fail.
This change adds a --force flag to the bundler install, ignoring the conflict


ferglor:
This is really just a quick fix to get the nightlies passing again, but I'd like to follow up with someone in SRE to figure out why bundler is preinstalled on the VMs.

PR #1476, opened 2017-12-19
(BKR-1268) Load project options from .beaker.yml

joshcooper:
If .beaker.yml is present in the current working directory, then load
the project's options. The overall precedence from lowest to highest is:

    defaults
    .beaker.yml
    ~/.beaker/subcommand_options.yaml
    ./beaker/subcommand_options.yaml
    etc

The project options (.beaker.yml) are essentially project defaults, so
they come after beaker defaults and before the subcommand options.

It is also possible for the project options to specify the options_file
to load so that it doesn't need to be specified on the command line.

Note this fixes a spec test that was stubbing part of the parse_args
method that it was trying to test.

### DIFF
--- a/lib/beaker/options/parser.rb
+++ b/lib/beaker/options/parser.rb
@@ -206,7 +206,8 @@ module Beaker
       #   4.  options file values
       #   5.  subcommand options, if executing beaker subcommands
       #   6.  subcommand options from $HOME/.beaker/subcommand_options.yaml
-      #   7.  default or preset values are given the lowest priority
+      #   7.  project values in .beaker.yml
+      #   8.  default or preset values are given the lowest priority
       #
       # @param [Array] args ARGV or a provided arguments array
       # @raise [ArgumentError] Raises error on bad input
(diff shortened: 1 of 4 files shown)


lib/beaker/options/parser.rb
```diff
@@ -217,6 +217,12 @@ def parse_args(args = ARGV)
         cmd_line_options[:command_line] = ([$0] + args).join(' ')
         @attribution = @attribution.merge(tag_sources(cmd_line_options, "flag"))
 
+        if File.exists?('.beaker.yml')
```
tvpartytonight: It seems like there needs to be an addition to the priority list in the yarddoc for this `parse_args` method. The interesting issue here is that the precedence listed here would indicate project => global => subcommand, subcommand winning out. It isn't linear, per se, but, it actually seems to be more usable to me. Was that intentional to do it that way?

I also think that you can eliminate this `if` block of code and just add `"project" => ".beaker.yml"` to the middle of the options_files just below this. That code handles the file detection and merging.

joshcooper: > It seems like there needs to be an addition to the priority list in the yarddoc for this parse_args method.

Ah yup, I'll update that.

The rest makes sense, let me update that.

tvpartytonight:

tvpartytonight:

tvpartytonight:
Thanks for the addition @joshcooper !

PR #1478, opened 2018-01-05
(BKR-1169) Ensure indentation is balanced

joshcooper:
Add `with_indent` method which indents the log for the duration of the
block. Use `ensure` so the indent level is always balanced, even if the
block raises an exception, e.g. due to skip_test.

Deprecate `step_in` and `step_out` since they can leave the indentation
unbalanced.


joshcooper:
I need to add a test that if the block raises, we still restore the original indent level.

joshcooper:
@kevpl this should be good to go /cc @tvpartytonight

kevpl:
🏆 🍹 thanks Josh!

PR #1481, opened 2018-01-10
(BKR-487) Remove AWS objects from host_preserved

RandellP:
This is a medium term fix for BKR-487.
The issue is that AWS objects get written to the host_preserved.yml file, but they can't be read in
because the AWS gem isn't loaded at read time.  And apparently even if it is... the initiate returns nil
which psych doens't like... (see the ticket for details)

This change is to convert those AWS objects to hashes just before writing the yml file.
This is a pretty poor fix overall.  But it will allow loading the host_preserved.yml file and running things
on the aws box.  But the later runs won't be able to kill the box or use any of the aws instance object methods...
But can't do that right now anyway, so it is a step forward. :)

### DIFF
--- a/lib/beaker/cli.rb
+++ b/lib/beaker/cli.rb
@@ -240,10 +240,56 @@ module Beaker
       hosts_yaml['CONFIG'] = hosts_yaml['CONFIG'].merge(@options.reject{ |k,v| k =~ dontpreserve })
       # remove copy of HOSTS information
       hosts_yaml['CONFIG']['provision'] = false
+      hosts_yaml['HOSTS'].each do |ip,host|
+        if ( ! host.key?(:instance) )
+          next
+        end
+        ec2_instance = host[:instance]
+        new_instance = ec2_instance_to_hash(ec2_instance)
+        host[:instance] = new_instance
+      end
+
       File.open(preserved_hosts_filename, 'w') do |file|
         YAML.dump(hosts_yaml, file)
       end
       @options[:hosts_preserved_yaml_file] = preserved_hosts_filename
+    end
+
+    # ec2_instance_to_hash
+    #
+    # converts an ec2 instance to a hash
+    # recurses into the instance variables looking for other AWS object to convert as well
+    # skips the configuration object because the AWS object doesn't write that to YAML
+    # This is specifically tailored to be a medium term solution to BKR-487
+    # It would be the wrong thing to use for anything else.
+    #
+    # @param obj
+    # @return [hash]
+    def ec2_instance_to_hash obj
+      h = Hash.new
+      if ( obj.class == AWS::Core::Configuration )
+        return obj
+      end
+      if ( obj.class.to_s =~ /^AWS/ )
+        obj.instance_variables.each do |var|
+          val = obj.instance_variable_get(var)
+          new_val = ec2_instance_to_hash(val)
+          if ( new_val.class != AWS::Core::Configuration )
+            h[var.to_s.delete("@")] = new_val
+          end
+        end
+      elsif ( obj.class == Hash )
+        obj.each do |key,val|
+          if ( val.class.to_s =~ /^AWS/ )
+            h[key] = ec2_instance_to_hash(val)
+          else
+            h[key] = val
+          end
+        end
+      else
+        return obj
+      end
+      return h
     end
 
     # Return a host_hash that is a merging of options host hashes with instance host objects


mchllweeks:

RandellP:

RandellP:

lib/beaker/cli.rb
```diff
@@ -240,12 +240,58 @@ def preserve_hosts_file
...
+    # This is specifically tailored to be a medium term solution to BKR-487
+    # It would be the wrong thing to use for anything else.
+    #
+    # @param obj
+    # @return [hash]
+    def ec2_instance_to_hash obj
+      h = Hash.new
+      if ( obj.class == AWS::Core::Configuration )
+        return obj
+      end
```
er0ck: should this conditional be after the one below as it's much less likely?

lib/beaker/cli.rb
```diff
@@ -240,12 +240,58 @@ def preserve_hosts_file
...
+      if ( obj.class.to_s =~ /^AWS/ )
+        obj.instance_variables.each do |var|
+          val = obj.instance_variable_get(var)
+          new_val = ec2_instance_to_hash(val)
```
er0ck: ooooh.  branching recursion.  bold

er0ck: i'd like to see this as a polymorphic conversion class.  but it's probably not worth the work for a temporary work around.

er0ck:
do we have AWS acceptance tests for beaker?  i'd like to see one for hosts_preserved, sub-commands, in general

kevpl:
@RandellP have you thought about using the hypervisor itself to make any changes that it might need, rather than storing more hypervisor-specific needs in beaker itself?

You should be able to access the Hypervisor objects themselves through accessors on `cli.rb` & `network_manager.rb`. Calling from `cli.rb`, you should be able to run:
```ruby
network_manager.hypervisors
```
to access them. If you do that, then you could do something like `responds_to?` your method name, executing it if so. Then you can provide this method in beaker-aws or -abs or wherever you need & things would be skipped just fine when not needed. This way, if anyone needs changes going forward like this use-case, they become much easier to add in the future, as people can just create their own hypervisor-specific version of this method in their hypervisor code.

kevpl:
@er0ck we test each of those things, but not against each other in combination.

lib/beaker/cli.rb
RandellP: no, because this condition is intended to stop execution from going into the conditional below.

lib/beaker/cli.rb
RandellP: lol

er0ck:
oooh.  i like kevin's hypervisors idea

RandellP:

RandellP:
@kevpl I did think about doing this in more interesting ways...  I actually started working in the beaker-aws repo, but couldn't solve the problem from there as there was nowhere to hook into to get my code called just before the yml is written.  And I can't do it any sooner because it would break an initial preserve-hosts=never run from deleting the instances at the end.

I could do more to make it generic such that other hypervisors could write similar customized code, but I don't think that is the right direction to go.  This kind of thing is barely a bandaid; as overall functionality is still broken, just a little less.  And the long term fix that we need in the near future would make the code something that should be deleted.  So I didn't think it was worth the effort to make it possible for other hypervisors to do the same thing.  Also, I think the long term fix will go in the same directory, or at least repo, so having this medium term fix there means only one PR to implement the long term fix and remove the medium term one.

One thing I could do is check if the AWS hypervisor is even loaded, and only do this if it is.  But a run would have to have like 1000 Hosts to save only a few seconds of run time... so didn't seem worth the complexity.

I kinda glazed over some things in the above that maybe you don't already know, and maybe there are things I don't know... so if anything doesn't seem to make sense or your opinion differs, lets talk face to face on it. :)  It's easier than typing out all the background.

RandellP:

RandellP:
the tests seem to be failing due to 504's from the forge.  And they move to different tests on different runs.  Is getting them all to pass a realistic expectation?

RandellP:

RandellP:

kevpl:
@RandellP and I discussed this, it sounds like using a development branch of beaker will be the chosen method for keeping this change around until it isn't needed anymore.

Shall we close the PR then, or would you like to keep it around for something?

RandellP:
Closing to move to a branch build

PR #1484, opened 2018-01-18
(maint) Add redhat_fips to the supported platforms regex

ScottGarman:
I *think* this is needed for beaker to work with the new redhat_fips platform definition, but please double-check to make sure that's the case.

### DIFF
--- a/lib/beaker/platform.rb
+++ b/lib/beaker/platform.rb
@@ -3,7 +3,7 @@ module Beaker
   # all String methods while adding several platform-specific use cases.
   class Platform < String
     # Supported platforms
-    PLATFORMS = /^(huaweios|cisco_nexus|cisco_ios_xr|(free|open)bsd|osx|centos|fedora|debian|oracle|redhat|scientific|sles|ubuntu|windows|solaris|aix|archlinux|el|eos|cumulus|f5|netscaler)\-.+\-.+$/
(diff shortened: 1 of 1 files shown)


ScottGarman:
Related BHG PR: https://github.com/puppetlabs/beaker-hostgenerator/pull/106

lib/beaker/platform.rb
```diff
@@ -3,7 +3,7 @@ module Beaker
   # all String methods while adding several platform-specific use cases.
   class Platform < String
     # Supported platforms
-    PLATFORMS = /^(huaweios|cisco_nexus|cisco_ios_xr|(free|open)bsd|osx|centos|fedora|debian|oracle|redhat|scientific|sles|ubuntu|windows|solaris|aix|archlinux|el|eos|cumulus|f5|netscaler)\-.+\-.+$/
+    PLATFORMS = /^(huaweios|cisco_nexus|cisco_ios_xr|(free|open)bsd|osx|centos|fedora|debian|oracle|redhat_fips|redhat|scientific|sles|ubuntu|windows|solaris|aix|archlinux|el|eos|cumulus|f5|netscaler)\-.+\-.+$/
```
kevpl: @ScottGarman is there a reason that `redhat_fips` needs to be a beaker platform?

Why can't it be another platform that beaker just understands as `redhat`?

mchllweeks: @ScottGarman I think this is what bhg should look like instead so that beaker just treats it as a redhat platform: https://github.com/puppetlabs/beaker-hostgenerator/pull/107.  Sorry I didn't catch this sooner before I merged your bhg changes.

ScottGarman:
Closing this as we ended up using el-7-x86_64 as the platform_name in BHG.

PR #1483, opened 2018-01-16
(maint) update yard dependency

steveax:
This commit updates yard to ~> 0.9.11 [1].

[1]: https://nvd.nist.gov/vuln/detail/CVE-2017-17042

PR #1487, opened 2018-02-07
(MAINT) update yard dependency

kevpl:
takes the work from #1483 & adds a fix to 
doc generation so that it can pass through
doc generation & spec testing.


kevpl:

PR #1489, opened 2018-02-09
(BKR-1422) add '--test-tag-exclude' as an option for beaker subcommands

melissa:


lib/beaker/subcommand.rb
```diff
@@ -53,6 +53,7 @@ def initialize(*args)
     class_option :'xml-time-order', :type => :boolean, :group => 'Beaker run'
     class_option :'debug-errors', :type => :boolean, :group => 'Beaker run'
     class_option :'exec_manual_tests', :type => :boolean, :group => 'Beaker run'
+    class_option :'test-tag-exclude', :type => :string, :group => 'Beaker run'
```
tvpartytonight: Does it make sense to restrict this option to `exec` only? You could add it specific to that subcommand using the `method_option`...method. There's an example of it for hosts in the `init` function.

melissa: I don't think so from the way we're invoking beaker when running subcommands. Additional options are passed into all subcommands regardless, so we'll need all subcommands to be able to parse the flag

melissa:
@tvpartytonight are there any additional tests or anything I need to add for this PR?

Also, I've tested this locally and this does fix our issue

lib/beaker/subcommand.rb
tvpartytonight: While we are in this section of code, should we update it to also accept `--test-tag-and` and `--test-tag-or`? Other than that, I'm a 👍 here.

kevpl:
retest this please

lib/beaker/subcommand.rb
melissa: I'm happy to add those in. I wasn't sure how much change I should introduce, but I'll go update this now

smcelmurry:
retest this please

smcelmurry:
retest this please

PR #1491, opened 2018-02-12
(BKR-1426) Fully qualify sles ssh restart cmd

ericwilliamson:
Previous to this commit, the sles SSH restart command was not fully
qualified. This lead to the command not being found if the user did not
have `/usr/sbin` in their path for some reason as we could not set the
path until after we modify the ssh service to permit user environments
(which requires a ssh restart).
This commit fully qualifies the sles `rcsshd` command. From testing
internally it appears the command exists at `/usr/sbin` for sles 11
32/64 bit and sles 12 64 bit.


ericwilliamson:
jenkins, retest this please

PR #1498, opened 2018-03-01
Fix `BEAKER_provision=no` workflow

joshuaspence:
Fixes BKR-1111 and BKR-494.

### DIFF
--- a/lib/beaker/hypervisor.rb
+++ b/lib/beaker/hypervisor.rb
@@ -38,7 +38,7 @@ module Beaker
 
       hypervisor = hyper_class.new(hosts_to_provision, options)
       self.set_ssh_connection_preference(hosts_to_provision, hypervisor)
-      hypervisor.provision if options[:provision]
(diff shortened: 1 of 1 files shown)


pinkypie:

mchllweeks:

kevpl:
@joshuaspence reading the tickets for this issue, it sounds like the real problem here is that the host preservation functionality between runs isn't putting the correct info into the file for the next run.

It seems that this solution takes away people's ability to run beaker without provisioning Systems Under Test (SUTs), which is a valid use-case, in order to allow your second run to work (although it has given up on not provisioning those SUTs). Am I mis-reading your work here? If not, then this isn't going to work for beaker as a valid solution to the issue.

mchllweeks:
Last comment still current.

joshuaspence:
I think I know a better fix.

joshuaspence:
See puppetlabs/beaker-vagrant#13 instead.

PR #1500, opened 2018-03-13
(BKR-1439) Make Solaris 11 use OpenCSW for package installation

ScottGarman:
Previously Solaris 11 used a built-in repo with the 'pkg' command to
install third-party packages, and the git package provided in that repo
does not support TLS 1.2 and is no longer compatible with GitHub https:
repos. This updates Solaris 11 to behave the same as Solaris 10, which
uses the OpenCSW repo which offers a wide variety of packages, including
a git package which is compatible with GitHub https: clones.


lib/beaker/host/unix/pkg.rb
```diff
@@ -93,14 +93,14 @@ def install_package(name, cmdline_args = '', version = nil, opts = {})
...
         update_apt_if_needed
         execute("apt-get install --force-yes #{cmdline_args} -y #{name}", opts)
-      when /solaris-11/
-        if opts[:acceptable_exit_codes]
-          opts[:acceptable_exit_codes] << 4
-        else
-          opts[:acceptable_exit_codes] = [0, 4] unless opts[:accept_all_exit_codes]
+      when /solaris-/
+        if self['platform'] =~ /solaris-11/
+          if opts[:acceptable_exit_codes]
```
ScottGarman: @kevpl Do you recall the reasoning behind adding 4 to the acceptable exit codes for Solaris 11? I've preserved that code here, but I couldn't find any explanation for what it was there for in the previous commit. If it was due to some oddity with the pkg command, then I could probably remove this entirely. However, I did take a look at the man page for pkg and it makes no reference to a return code of 4, so I'm not sure.

kevpl: @ScottGarman yeah, looking back at [the commit](https://github.com/puppetlabs/beaker/commit/a661dfe0dab92991808d7d4e9cbdee66afcf7022), I'm not able to find an explanation for it either. The way the code is formatted though, it would seem that it's really only there for the `pkg` command, so I think you should delete it since we're no longer using that command at all

kevpl:
👍 overall. Please add a test to the code & an entry to the CHANGELOG file, then it will be ready to push through the automation towards merging! Thanks for the fix!

mchllweeks:

kevpl:
@ScottGarman is this still on your radar? We'd prefer to not have PRs sitting as reminders in our projects. We'd rather close them & leave this space for things that are in progress.

Would you mind using the ticket to keep this in your mind instead & re-open the PR when you've got the other work ready?

PR #1512, opened 2018-05-30
(BKR-1007) Use AIX 6.1 packages everywhere for puppet6

caseywilliams:
Adjusts the AIX package selection criteria: puppet-agent will use AIX
6.1 packages for both AIX 7.1 and 7.2 in puppet 6; puppet 5 maintains
existing behavior around these platforms.

### DIFF
--- a/lib/beaker/host/unix/pkg.rb
+++ b/lib/beaker/host/unix/pkg.rb
@@ -417,8 +417,15 @@ module Unix::Pkg
         puppet_collection, puppet_agent_version, opts )
     when /^(sles|aix|el|centos|oracle|redhat|scientific)$/
       variant = 'el' if variant.match(/(?:el|centos|oracle|redhat|scientific)/)
-      arch = 'ppc' if variant == 'aix' && arch == 'power'
-      version = '7.1' if variant == 'aix' && version == '7.2'
+      if variant == 'aix'
+        arch = 'ppc' if arch == 'power'
+        version_x, version_y = /^(\d+)\.(\d+)/.match(puppet_agent_version).captures.map(&:to_i)
+        if version_x < 5 || version_x == 5 && version_y < 99 # 5.99.z indicates pre-release puppet6
+          version = '7.1' if version == '7.2'
+        else
+          version = '6.1'
+        end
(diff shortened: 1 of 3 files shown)


McdonaldSeanp:
jenkins test this please

caseywilliams:
jenkins test this please

smcclellan:
Blocked by master branch of beaker being broken, merge when that is fixed.

lib/beaker/host/unix/pkg.rb
```diff
@@ -417,8 +417,14 @@ def puppet_agent_dev_package_info( puppet_collection = nil, puppet_agent_version
...
+      if variant == 'aix'
+        arch = 'ppc' if arch == 'power'
+        if %w(PC1 puppet5).include? puppet_collection
+          version = '7.1' if version == '7.2'
+        else
+          version = '6.1'
+        end
+      end
       release_path_end = "#{variant}/#{version}/#{puppet_collection}/#{arch}"
       release_file = "puppet-agent-#{puppet_agent_version}-1.#{variant}#{version}.#{arch}.rpm"
```
caseywilliams: I actually think these paths will be incorrect -- it looks like all shipped builds still use 'PC1' in their URLs, regardless of the collection. I'll block this for now.

caseywilliams: Ok, I've updated this to rely on version numbers instead of collection names. This is now blocked on an update of the reported version in puppet-agent#master to 5.99.z -- I'll unblock as soon as that happens.

caseywilliams:
puppet-agent#master has gotten a 5.99 tag, so this is ready to merge from our perspective -- master beaker CI is still red, but I'm unblocking this

kevpl:
It would be better for everyone for this to be refactored out of core beaker into beaker-puppet, but I'm fine with this going into beaker as-is for now.

We're seeing unrelated failures right now, but I'd like to kick off one more time to see if we can get to the final stages of testing on this. If it doesn't get that far & the failures are still unrelated, I'm fine with merging it as-is.

Jenkins, test this please

PR #1513, opened 2018-06-05
(BKR-1472) Enhance Documentation

Dakta:
Enhances the Beaker documentation with formatting cleanup and additional information.

PR #1514, opened 2018-06-06
(BKR-1472) Unwrap Markdown

Dakta:
Hard wrapping on Markdown text was applied inconsistently between (and within) docs; all hard wrapping has been removed.

Split from #1513

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,22 +20,17 @@
 ````
     (BKR-1234) Make the example in CONTRIBUTING imperative and concrete
 
-    Without this patch applied the example commit message in the CONTRIBUTING
-    document is not a concrete example.  This is a problem because the
-    contributor is left to imagine what the commit message should look like
-    based on a description rather than an example.  This patch fixes the
-    problem by making the example concrete and imperative.
+    Without this patch applied the example commit message in the CONTRIBUTING document is not a concrete example.  This is a problem because the contributor is left to imagine what the commit message should look like based on a description rather than an example.  This patch fixes the problem by making the example concrete and imperative.
 
-    The first line is a real life imperative statement with a ticket number
-    from our issue tracker.  The body describes the behavior without the patch,
-    why this is a problem, and how the patch fixes the problem when applied.
-````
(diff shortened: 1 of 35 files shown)


kevpl:
What's the advantage to completely unwrapping as opposed to having a limit?

Other than that, I'm fine with this. It'll need to be rebased to be ready to merge though.

Dakta:
There are pitfalls to hard-wrapping Markdown in particular:

- Though uncommon, it can lead to unexpected formatting outcomes. Especially when modifying the wrapping, where block-defining syntax can get wrapped onto a new line. For example, if an inline hyphen is hard-wrapped onto a new line it will break the paragraph by starting a list. Same for a number of other block-level formatting characters.
- Changes to line content require re-wrapping. While software tools can accomplish this, see above. Single-word changes that result in re-wrapping show extraneous whitespace changes in wrapped adjacent lines. In a long paragraph, this can result in dozens of lines modified for a single word change.
- Hard-wrapping of indented blocks requires re-indenting (not necessarily handled properly by generic tools), and can lead again to surprise formatting outcomes.
- Hard-wrapping without a standard line-length results in inconsistent raw formatting across and within files. If the hard-wrapping is longer than the developer's editor wrapping length, it can result in hanging lines. If it matches perfectly, the developer must check the wrapping to match it, because their editor's soft-wrapping will visually match the document's hard-wrapping. If it's shorter, the developer may not notice; if they apply hard wrapping it may not match.

While I'm opposed to hard wrapping, my preferences don't drive this project and this is a big change just for whitespace; I understand. At the least we should adopt hard-wrapping guidelines as part of the Markdown formatting doc so that things are more consistent moving forward.

kevpl:
hehe, I preferred having a hard-wrap before this, but your simple argument in #1518 on it ([here unless it's moved around since then](https://github.com/puppetlabs/beaker/pull/1518/files#diff-9a67ecce54db708099f48cd19850f6d3R14)) convinced me, let alone all the points you make here.

I'm in favor of this, but it'll need to be rebased before it can be merged.

Dakta:
Rebased and ready for a clean, fast-forward merge.

PR #1519, opened 2018-06-06
(BKR-463) Rsync now fails with CommandFailed when destination path does not exist

Dakta:
Fixes BKR-463 as described, and includes improvement to spec mocking for rsync calls which should allow future rsync error handling improvements. Acceptance test coverage is modified to indicate fixed behavior.

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -502,6 +502,8 @@ module Beaker
     # @param to_path [String] The destination path on the host
     # @param opts [Hash{Symbol=>String}] Options to alter execution
     # @option opts [Array<String>] :ignore An array of file/dir paths that will not be copied to the host
+    # @raise [Beaker::Host::CommandFailure] With Rsync error (if available)
+    # @return [Rsync::Result] Rsync result with status code
     def do_rsync_to from_path, to_path, opts = {}
       ssh_opts = self['ssh']
       rsync_args = []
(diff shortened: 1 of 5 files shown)


lib/beaker/host.rb
```diff
@@ -571,7 +571,9 @@ def do_rsync_to from_path, to_path, opts = {}
       @logger.notify "rsync: localhost:#{from_path} to #{hostname_with_user}:#{to_path} {:ignore => #{opts[:ignore]}}"
       result = Rsync.run(from_path, to_path, rsync_args)
       @logger.debug("rsync returned #{result.inspect}")
-      result
+
+      return result if result.success?
+      raise Beaker::Host::CommandFailure, result.error
```
kevpl: @Dakta please add a `@return` & `@raise` yard statement to the method, so we'll be better informed as to the different return types depending on the success of rsync execution.

To what you mentioned about nested exceptions (although it sounds like `Rsync::Result` will fail without throwing an exception), some quick googling pointed me to [this article](http://blog.honeybadger.io/nested-errors-in-ruby-with-exception-cause/), which makes it sound like exceptions will automatically get chained together and can be inspected using the `exception.cause` method.

Dakta: Done.

Dakta:
Due to potential breaking behavior of throwing exceptions where previously would fail silently, this merge has been deferred to 4.0 release.

PR #1524, opened 2018-06-18
(MAINT) Adding some generic host helper utility methods

jsane:
PE-16486 required checking of file permissions of certain files and as part of
that work some generic utility type methods were identified. And reviewers suggested
they be made part of beaker for more broader use.
Pl refer to PR https://github.com/puppetlabs/pe_acceptance_tests/pull/1883 for additional background.

### DIFF
--- a/lib/beaker/dsl/helpers/host_helpers.rb
+++ b/lib/beaker/dsl/helpers/host_helpers.rb
@@ -606,6 +606,31 @@ module Beaker
             end
           end
         end
+
+        def ruby_command(host)
+          "env PATH=\"#{host['privatebindir']}:${PATH}\" ruby"
+        end
+
+        # Returns an array containing the owner, group and mode of
+        # the file specified by path. The returned mode is an integer
+        # value containing only the file mode, excluding the type, e.g
+        # S_IFDIR 0040000
+        def stat(host, path)
+          ruby = ruby_command(host)
+          owner = on(host, "#{ruby} -e 'require \"etc\"; puts (Etc.getpwuid(File.stat(\"#{path}\").uid).name)'").stdout.chomp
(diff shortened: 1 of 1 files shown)


lib/beaker/dsl/helpers/host_helpers.rb
```diff
@@ -606,6 +606,31 @@ def echo_on hosts, val
             end
           end
         end
+
+        def ruby_command(host)
+          "env PATH=\"#{host['privatebindir']}:${PATH}\" ruby"
```
kevpl: @jsane `host['privatebindir']` won't be available necessarily to beaker in the next major version (coming soon), as it will only be getting set from beaker-puppet or beaker-pe. Since that's the case, I don't believe beaker is the right place for these methods.

beaker-pe depends on beaker-puppet though, so I believe these methods should go into that repo. Let me know if you have any questions about this.

kevpl:
last comment still current

jsane:
Sorry for the delay in getting back. Am fine with adding these to beaker-pe instead. I was going by suggestions from @RandellP and @highb about adding them to beaker.

Dakta:
@jsane You should be able to address this issue by using a manual Bundler override for a local dependency. See [here](https://sourcediving.com/use-bundler-local-repository-override-when-developing-your-git-gems-7beeea3818c0). This sets a per-project override for a Gem dependency location. You'll want to set this in the top level project that you're running the `pe_acceptance_test`.

With the Beaker modules split, this issue of development dependency overrides needs a second look, since it's confusing to have two mechanisms for specifying local overrides, especially when one of them only works for the main Beaker Gem.

jsane:
@Dakta thanks. 
But for some reason that would not work for me. I had to explicitly add a line to the Gemfile and point it to my path. Joshua P gave that tip in https://github.com/puppetlabs/pe_acceptance_tests/pull/1883 though am still having some strange issues with using the new methods defined in beaker-puppet from within pe_acceptance_tests. In any case I will close this.

PR #1526, opened 2018-06-21
(BKR-1481) Introduce Beaker::Shared::FogFileParser.parse_fog_file()

Dakta:
Introduces a shared module for parsing .fog credential files, with descriptive error messages and spec coverage. Transparently supports Symbol or String type keys ([just like official fog-core does](https://github.com/fog/fog-core/blob/7865ef77ea990fd0d085e49c28e15957b7ce0d2b/spec/utils_spec.rb#L11)). Throws useful error messages:
[x] File exists
[x] File has contents
[x] Valid YAML
[x] Has a `default` section (not required?)
[x] Loads the correct credential group, may be [ENV-specified](https://www.rubydoc.info/gems/fog-core/1.42.0/Fog.credential)

Affected components (module updates will reference this PR):
- [x] beaker
    - [x] spec
    - [x] acceptance
- [x] beaker-vmware - puppetlabs/beaker-vmware#5
    - [x] spec
    - [x] acceptance
- [x] beaker-vcloud: depends on `Beaker::Hypervisor::VsphereHelper` (from `beaker-vmware`) so no changes, just spec check
    - [x] spec
- [x] beaker-aws - puppetlabs/beaker-aws#12
    - [x] spec
    - [x] acceptance
- [x] beaker-vmpooler - puppetlabs/beaker-vmpooler#11
    - [x] spec

### DIFF
--- a/lib/beaker/shared/fog_file_parser.rb
+++ b/lib/beaker/shared/fog_file_parser.rb
@@ -1,0 +1,41 @@
+require 'stringify-hash'
+
+module Beaker
+  module Shared
+    # A set of functions to read .fog files
+    module FogFileParser
+      # Read a .fog file
+      #
+      # @note Loaded .fog files may use symbols for keys.
+      #    Although not clearly documented, this *is* valid:
+      #    https://www.rubydoc.info/gems/fog-core/1.42.0/Fog#credential-class_method
+      #    https://github.com/fog/fog-core/blob/7865ef77ea990fd0d085e49c28e15957b7ce0d2b/spec/utils_spec.rb#L11
+      #
+      # @param fog_file_path [String] dot fog path
+      # @return [StringifyHash] credentials stored in fog_file_path
+      # @raise [ArgumentError] when the specified credential file is invalid, describing the reson
+      def parse_fog_file(fog_file_path = '.fog', credential = nil)
+        if !File.exist?(fog_file_path)
+          raise ArgumentError, ".fog file '#{fog_file_path}' does not exist"
+        end
+        begin
+          fog = YAML.load_file(fog_file_path)
+        rescue Psych::SyntaxError => e
+          raise ArgumentError, ".fog file '#{fog_file_path}' is not valid YAML:\n(#{e.class}) #{e.message}"
+        end
+        if fog == false # YAML.load => false for empty file
+          raise ArgumentError, ".fog file '#{fog_file_path}' is empty"
+        end
+        # transparently support symbols or strings for keys
+        fog = StringifyHash.new.merge!(fog)
+        # respect credential from env
+        # @note ENV must be a string, e.g. "default" not ":default"
+        credential ||= (ENV["FOG_CREDENTIAL"] && ENV["FOG_CREDENTIAL"].to_sym) || :default
+        if not fog[credential]
+          raise ArgumentError, ".fog file (#{fog_file_path}) is missing the required section: `#{credential}`"
+        end
+        fog[credential]
+      end
+    end
+  end
+end
--- a/lib/beaker/shared/fog_file_parser.rb
+++ b/lib/beaker/shared/fog_file_parser.rb
@@ -11,28 +11,35 @@ module Beaker
       #    https://www.rubydoc.info/gems/fog-core/1.42.0/Fog#credential-class_method
       #    https://github.com/fog/fog-core/blob/7865ef77ea990fd0d085e49c28e15957b7ce0d2b/spec/utils_spec.rb#L11
       #
-      # @param fog_file_path [String] dot fog path
+      # @param fog_file_path [String] dot fog path. Overridden by
       # @return [StringifyHash] credentials stored in fog_file_path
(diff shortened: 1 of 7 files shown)


kevpl:
@Dakta it looks like this is failing due to these changes.

```ruby
13:51:41 Failures:
13:51:41 
13:51:41   1) Beaker::Shared::FogFileParser#parse_fog_file returns the named credential section from ENV even when an argument is provided
13:51:41      Failure/Error: expect( parse_fog_file( fog_file_path = '/path/that/does/not/exist/.fog' )[:vmpooler_token], credential = :default ).to eq( "correct_token" )
13:51:41      
13:51:41      ArgumentError:
13:51:41        wrong number of arguments (2 for 0..1)
13:51:41      # ./spec/beaker/shared/fog_file_parser_spec.rb:81:in `block (3 levels) in <module:Shared>'
13:51:41 
```

Please take a look when you can.

lib/beaker/shared/fog_file_parser.rb
```diff
@@ -0,0 +1,48 @@
...
+      # @param fog_file_path [String] dot fog path. Overridden by
+      # @return [StringifyHash] credentials stored in fog_file_path
+      # @raise [ArgumentError] when the specified credential file is invalid, describing the reson
+      def parse_fog_file(fog_file_path = '~/.fog', credential = :default)
+        # respect credential location from env
+        if ENV["FOG_RC"]
+          fog_file_path = ENV["FOG_RC"]
+          from_env = ' set in ENV["FOG_RC"]'
+        end
+        if !File.exist?(fog_file_path)
```
ekinanp: I don't know if all this error handling is necessary. Shouldn't YAML.load output an appropriate error message if the fog file does not exist? And if the syntax is incorrect (so that the Psych::SyntaxError would not be necessary)?

lib/beaker/shared/fog_file_parser.rb
```diff
@@ -0,0 +1,48 @@
...
+      #
+      # @note Loaded .fog files may use symbols for keys.
+      #    Although not clearly documented, this *is* valid:
+      #    https://www.rubydoc.info/gems/fog-core/1.42.0/Fog#credential-class_method
+      #    https://github.com/fog/fog-core/blob/7865ef77ea990fd0d085e49c28e15957b7ce0d2b/spec/utils_spec.rb#L11
+      #
+      # @param fog_file_path [String] dot fog path. Overridden by
+      # @return [StringifyHash] credentials stored in fog_file_path
+      # @raise [ArgumentError] when the specified credential file is invalid, describing the reson
+      def parse_fog_file(fog_file_path = '~/.fog', credential = :default)
```
ekinanp: I think changing credential => credentials_key might make this more readable (b/c you're getting the credentials from that key). The function name should also be changed to `parse_credentials_from_fog_file` (or something similar), b/c you're parsing the credentials here, not the entire fog file.

lib/beaker/shared/fog_file_parser.rb
Dakta: Wrapping the full variety of typical failure here under ArgumentError is intended to reduce the duplication of error-handling code throughout all calling methods. Previously, consumers performed a variety of inconsistent error handling on their own ([example](https://github.com/puppetlabs/beaker-vmpooler/pull/11/files#diff-ca1d581f345cffd7fd28e3ffa1d50c0bL43)). But to them, it doesn't matter *why* parsing fails, as any way the credentials are not available. Moving error handling to here simplifies the spec of callers so that their tests do not have duplicate coverage of error conditions, which adding appropriate exception handling to all of them would necessitate. I tried to find a balance between modifying the behavior and spec tests of the callers and providing a uniform API for them to all consume. I guess I should prune the consumer error handling and spec tests down to reflect this.

lib/beaker/shared/fog_file_parser.rb
Dakta: I have [re-used the naming convention from fog-core](https://github.com/fog/fog-core/blob/f6ef6c3d9458bc235be57100acdced264e85e52a/lib/fog/core/credentials.rb#L35), but am not attached to it so the name change sounds good. Actually, looking at this again, this isn't so much a file parser as it is a generic fog credential provider (note the preference for environment variables, as described in the fog spec), so I'll rename it.

lib/beaker/shared/fog_file_parser.rb
ekinanp: Oh ok. Why not something like?
```
begin
  YAML.load_file(fog_file_path)
rescue Psych::SyntaxError, Errno::ENOENT => e
  raise ArgumentError.new("Failed getting credentials from fog file. Reason: #{e.message}")
end
```

Dakta: Sure, that's a nice refactor that simplifies the spec tests.

Dakta:
@kevpl Tests look good, minus one fail that I'm guessing is transient.

Dakta:
The dependent PRs (in the hypervisors) that are linked to this one are just waiting for this PR to be merged and released.

PR #1528, opened 2018-06-26
(BKR-1393) Update documentation on multiple arguments to subcommand exec

Dakta:


kevpl:
@Dakta please add a spec test to make sure we don't break this functionality in the future

Dakta:
Todo: update <docs/tutorials/subcommands.md>

Dakta:
@kevpl Looks like it's already covered in the spec: https://github.com/Dakta/beaker/blob/2c60362272ea07f9780e849c5b62400c0e8b08b2/spec/beaker/subcommand_spec.rb#L221

kevpl:
from our conversation, it looks like the TODO left here is to remove the functionality change & our talk about subcommand warnings when executed out of step will go into another ticket

Dakta:
@kevpl It is done.

PR #1529, opened 2018-06-26
(BKR-1469) Don't remove the hosts_file option in beaker init 

ekinanp:


kevpl:
@tvpartytonight I'm not as familiar with the subcommand code as you are. Any reason off the top of your head why we should be looking for an alternative solution rather than the one presented here?

tvpartytonight:
@ekinanp I think the title of the commit is a bit misleading, because you are actually removing the hosts_file key from being saved. Can you elaborate more on why this commit is needed?

Dakta:
@tvpartytonight That seems to be exactly backwards: this change removes the hosts_file key from *not* being saved. On the immediate next line is called `options_to_write.delete(key)`, which caused hosts_file to be deleted from the save list; thus this change causes hosts_file to be saved.

ekinanp:
@tvpartytonight I updated the commit message.

tvpartytonight:
You are right @Dakta I misread the PR. @ekinanp I think that the way this change is implemented will have unforseen effects when using subcommands because it will change the options resolution. Another approach to consider would be to change the `:hosts_file` option to `.beaker/subcommand_options.yaml` when using `beaker init`. I think that way will be cleaner because beaker's options resolution will be unchanged for subcommands.

Dakta:
Maybe relevant: https://tickets.puppetlabs.com/browse/BKR-1145

kevpl:
@ekinanp @tvpartytonight has this conversation resolved? The resolution of this doesn't seem clear.

ekinanp:
I think @Dakta was going to look into it since I have other things to work on for my team (originally thought it'd be a simple fix).

ekinanp:
@mchllweeks Yeah, it can wait until he returns from PTO. /cc @jcoconnor

kevpl:
I remember that it seemed that Dakta was going to pick this up and get it ready when he had the time, which it doesn't appear came up during his internship.

With that being the case, I'd default to @tvpartytonight as the SME for subcommand options execution & his comment above of a different way to do this.

We don't typically have time to pick up PRs & get them across the finish line, our interns are a special case that doesn't cross the rest of the team sadly. Between this & your comment that you won't have time to move this forward, my recommendation is that we close this PR.

If I'm wrong about the status of the PR, or if there's an argument as to why this is important enough that our team should schedule work on it, let me know, & we'll take another look at this. Thanks!

kevpl:
last comment still current

ekinanp:
@kevpl Oh I apologize, I thought I responded to this.

Yes, we can close this PR. We do have a workaround branch that uses this solution which we use in the imaging pipelines (https://github.com/ekinanp/beaker/tree/temp-BKR-1469). It would be nice if this could be fixed so we don't need that branch, but if there's other more pressing work then that workaround branch is OK for now.

PR #1531, opened 2018-07-02
(BKR-496) Fix create_tmpdir_on() for all platforms

Dakta:
Previously, `create_tmpdir_on()` relied on calling `$ getent passwd #{user}` on hosts. This caused the command to fail on host systems that don't support `getent`, which manifested on Windows and macOS. This is sub-optimal, because macOS is even a POSIX compliant system, so even using the standard POSIX method for checking if a user exists (`id -u username`, per: https://stackoverflow.com/a/14811915) would have been an improvement over `getent`.

This changes the call to use host-specific method `user_get()`, which calls the appropriate underlying system to check if the passed username exists on the host. Should make it much less brittle to addition of future OS support.

Fixed the group used in `chown` on the tmpdir to get an argument, check exists, and fall back to the host default if not provided.

Also fixed the `user_get()` and `group_get()` methods on all hosts to behavior in line with other `Host` methods: calls a back of passed otherwise returns `Result`.

Fixing `create_tmpdir_on()` has surfaced acceptance errors, which seem to be an issue with the hosts configuration not matching the actual host state.

~~~
Errored Tests Cases: (9)
  Test Case acceptance/tests/base/dsl/install_utils/clone_git_repo_on_test.rb reported: #<Beaker::Host::CommandFailure: Host 'xxx' exited with 2 running:
 getent group pe-puppet
Last 10 lines of output were:
>
    Test line: acceptance/tests/base/dsl/install_utils/clone_git_repo_on_test.rb:42:in `block (3 levels) in run_test'
  Test Case acceptance/tests/base/dsl/helpers/host_helpers/backup_the_file_test.rb reported: #<Beaker::Host::CommandFailure: Host 'xxx' exited with 2 running:
 getent group pe-puppet
Last 10 lines of output were:
>
    Test line: acceptance/tests/base/dsl/helpers/host_helpers/backup_the_file_test.rb:6:in `block (3 levels) in run_test'
  Test Case acceptance/tests/base/dsl/helpers/host_helpers/create_remote_file_test.rb reported: #<Beaker::Host::CommandFailure: Host 'xxx' exited with 2 running:
 getent group pe-puppet
Last 10 lines of output were:
>
    Test line: acceptance/tests/base/dsl/helpers/host_helpers/create_remote_file_test.rb:38:in `block (3 levels) in run_test'
  Test Case acceptance/tests/base/dsl/helpers/host_helpers/create_tmpdir_on_test.rb reported: #<Beaker::Host::CommandFailure: Host 'xxx' exited with 2 running:
 getent group pe-puppet
Last 10 lines of output were:
>
    Test line: acceptance/tests/base/dsl/helpers/host_helpers/create_tmpdir_on_test.rb:33:in `block (4 levels) in run_test'
  Test Case acceptance/tests/base/dsl/helpers/host_helpers/curl_on_test.rb reported: #<Beaker::Host::CommandFailure: Host 'xxx' exited with 2 running:
 getent group pe-puppet
Last 10 lines of output were:
>
    Test line: acceptance/tests/base/dsl/helpers/host_helpers/curl_on_test.rb:20:in `block (3 levels) in run_test'
  Test Case acceptance/tests/base/dsl/helpers/host_helpers/retry_on_test.rb reported: #<Beaker::Host::CommandFailure: Host 'xxx' exited with 2 running:
 getent group pe-puppet
Last 10 lines of output were:
>
    Test line: acceptance/tests/base/dsl/helpers/host_helpers/retry_on_test.rb:7:in `block (3 levels) in run_test'
  Test Case acceptance/tests/base/dsl/helpers/host_helpers/rsync_to_test.rb reported: #<Beaker::Host::CommandFailure: Host 'xxx' exited with 2 running:
 getent group pe-puppet
Last 10 lines of output were:
>
    Test line: acceptance/tests/base/dsl/helpers/host_helpers/rsync_to_test.rb:37:in `block (4 levels) in run_test'
  Test Case acceptance/tests/base/dsl/helpers/host_helpers/scp_from_test.rb reported: #<Beaker::Host::CommandFailure: Host 'xxx' exited with 2 running:
 getent group pe-puppet
Last 10 lines of output were:
>
    Test line: acceptance/tests/base/dsl/helpers/host_helpers/scp_from_test.rb:25:in `block (4 levels) in run_test'
  Test Case acceptance/tests/base/dsl/helpers/host_helpers/scp_to_test.rb reported: #<Beaker::Host::CommandFailure: Host 'xxx' exited with 2 running:
 getent group pe-puppet
Last 10 lines of output were:
>
    Test line: acceptance/tests/base/dsl/helpers/host_helpers/scp_to_test.rb:5:in `block (3 levels) in run_test'
~~~


Dakta:
Currently have a problem with Windows:

Host.tmpdir outputs a path in `cygpath -m`. (Aside: [why does it do this](https://github.com/puppetlabs/beaker/commit/54b807f937554c81bb10e68dda35aaa38cfe4853)? it "can be passed to puppet and still be a valid fully qualified path on Windows.") Frustratingly, this is not compatible with Cygwin's `chown`:

~~~console
# create a test user and tmpdir
Administrator@host ~ $ net user /add "tmpdirtestuser"
The command completed successfully.

Administrator@host ~ $ cygpath -m $(mktemp -td .XXXXXX)
C:/cygwin64/tmp/.U6cLrY

# try to chown
Administrator@host ~ $ chown tmpdirtestuser C:/cygwin64/tmp/.U6cLrY

# failure
Administrator@host ~ $ ls -ld C:/cygwin64/tmp/.U6cLrY
drwxr-xr-x 1 Administrator None 0 Jul  9 23:28 C:/cygwin64/tmp/.U6cLrY

# POSIX paths work fine
Administrator@host ~ $ chown tmpdirtestuser /tmp/.U6cLrY

# success
Administrator@host ~ $ ls -ld /tmp/.U6cLrY
drwx------+ 1 tmpdirtestuser None 0 Jul  9 23:28 .
~~~

So either a) create_tmpdir_on needs to become host platform-aware (check for Windows and use `cygpath` to clean the path), b) Host.tmpdir needs to output POSIX-style paths, or c) we need a platform-independent abstraction for `chown` (that plays nice with all path formats on Windows).

a) sounds brittle and indesirable,

b) would be simplest, but will probably break something for someone (@joshcooper can you lend some insight?). How much platform-awareness do we want leaking out to userspace? Existing examples, such as `Mac.user_get()`, and the use of Cygwin, make it seem like we want to obscure all hosts behind a façade of POSIX (nay, Linux) normalcy. Fortunately, this can be accomplished with an optional parameter on `Host.tmpdir()` to return POSIX format, which will safely preserve existing functionality; we could deprecate this later. This seems least intrusive, if merely delaying the inevitable.

c) would help lay the groundwork for efforts to make cross-platform testing smoother. I envision a Host-provided method `chown` and `chgrp` which would just wrap the host's native methods for this. Everything except Windows would probably just inherit straight up `chown` from Unix host, and Windows could just wrap its `chown` call with a `cygpath` call to sanitize input.

kevpl:
Historically, we've done a lot of the a) option (examples: [1](https://github.com/puppetlabs/beaker/blob/7470a0d31a255c1ac1ec633d8f069491551108f8/acceptance/tests/base/dsl/helpers/host_helpers/curl_on_test.rb#L5-L8), [2](https://github.com/puppetlabs/beaker/blob/7439f35a4e24fc910502f32a2d3bbd9f98f07278/lib/beaker/host/windows.rb#L29), [3](https://github.com/puppetlabs/beaker/blob/786899647523f2baf9f76113193396fb5a111060/lib/beaker/dsl/helpers/host_helpers.rb#L218-L221)), but it's led us down a bad path of a lot of variants on a theme that aren't standardized & when break are typically fixed separately.

I agree that c) would be the better option here, but what I'd like to see even more than that would be a standardized version of the path cleaning operation for use by all of these other commands.

Dakta:
The problem with path cleaning is that I don't think we can necessarily do it without calling cygpath on a Windows host. Windows paths have a large amount of variability, and it's not just a matter of stripping the drive identifier or cygwin prefix (which in theory isn't even a fixed quantity) and swapping path separators: valid paths can also have DOS-style shortnames (e.g. `C:\PROGRA~1\`). We can't recover from a DOS-style name without passing it through cygpath on the machine that generated it... I suppose we could just ignore that edge case and naïvely convert paths on the beaker coordinator, but it would be a lot nicer if we just forced Windows hosts to deal with it. Windows hosts are the outlier, so they can have the burden of dealing with their own behavior differences.

Or, thinking along other lines... Host-output/generated paths probably shouldn't ever cross hosts. So the problem here is that cygwin's chown doesn't support Windows-style paths, not that Unix hosts should ever have to support them.

Dakta:
Per discussion:

> The "problem" is that Cygwin's `chown` doesn't support Windows-style paths (paths generated with --windows or --mixed from `cygpath`), even though many other commands do such as `ls`. I think the solution is to create a `chown` method on hosts, just like there is are `mv` and `rm_rf` methods. On the Windows hosts, they would for now pass the path to cygpath. E.g. `$ chown User:Group '$(cygpath PATH)'`

So upcoming, `Windows::Host.chown()` and `Windows::Host.chgrp()` methods that wrap cygwin's tools with a call to cygpath to sanitize their path arguments. With matching Host methods on other platforms.

Dakta:
Jenkins, retest this please.

Dakta:
Needs a corresponding ticket for create_tmpdir_on in beaker-puppet.

PR #1537, opened 2018-07-19
(BKR-1492) Remove PEDefaults and #configure_type_defaults_on

Dakta:

PR #1538, opened 2018-07-20
(BKR-462) Fixes rsync issues around rsync not installed on remote host

Dakta:
Fixes an issue with Solaris hosts (which do not have rsync installed by default), fixes sloppy testing of `#install_package` so that it doesn't leave behind rsync on hosts that don't have rsync installed by default, adds an `#uninstall_package` host helper, and checks SSH key files for existence before passing them to rsync.

This should fix the acceptance test failure on Solaris in #1537.


lib/beaker/dsl/helpers/host_helpers.rb
```diff
@@ -394,6 +394,16 @@ def install_package host, package_name, package_version = nil
           host.install_package package_name, '', package_version
         end
 
+        # Uninstall a package on a host
+        #
+        # @param [Host] host             A host object
+        # @param [String] package_name   Name of the package to uninstall
+        #
+        # @return [Result]   An object representing the outcome of *uninstall command*.
+        def uninstall_package host, package_name
```
kevpl: @Dakta I'm curious as to why you've added a host helper method here, I thought that you were in the camp that these methods shouldn't exist, we should be using host methods themselves rather than having these DSL proxy methods?

Dakta: I may not like them, but at least they should be consistent. I can't very well deprecate `#install_package`, and it felt awkward to not have it match. So I chose to write a matching one-line wrapper method. It'll be just as easy to decommission as `#install_package` since they both operate on a single host.

Dakta:
Waiting on fixup for solaris package installation failure.

Dakta:
Gem::InstallError: net-telnet requires Ruby version >= 2.3.0.

Calling this unrelated.

Dakta:
Other failures down the pipe seem to be due to loss of implicit include of beaker-puppet, e.g.:

PR #1547, opened 2018-10-09
Introduce BEAKER_HYPERVISOR

ekohl:
Beaker already knows how to generate hosts on the fly using beaker-hostgenrator. The problem is that there's no way to choose the default hypervisor. This means you always specify the long form. Using an environment variable we can simplify configs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ git logs & PR history.
 
 # [Unreleased](https://github.com/puppetlabs/beaker/compare/4.1.0...master)
 
+### Added
+
+- `BEAKER_HYPERVISOR` environment variable to choose the beaker-hostgenerator hypervisor
+
 # [4.1.0](https://github.com/puppetlabs/beaker/compare/4.0.0...4.1.0) - 2018.10.25
 
 ### Added
--- a/docs/tutorials/lets_write_a_test.md
+++ b/docs/tutorials/lets_write_a_test.md
@@ -89,7 +89,27 @@ end
 You can now run this with
 
 ```console
-    $ beaker --host redhat7-64ma.yaml --pre-suite install.rb --tests mytest.rb
+$ beaker --host redhat7-64ma.yaml --pre-suite install.rb --tests mytest.rb
+```
+
+## Creating the host configuration on at runtime
+
+When you don't want to store a static file, you can also let [beaker-hostgenerator](https://github.com/puppetlabs/beaker-hostgenerator) generate it on the fly. The filename is used as a host specification. A simple example is:
+
+```console
+$ beaker --host centos7-64 --pre-suite install.rb --tests mytest.rb
+```
+
+It's also possible to set the hypervisor and hostname:
+
+```console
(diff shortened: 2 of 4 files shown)


docs/tutorials/lets_write_a_test.md
```diff
@@ -26,6 +26,8 @@ What needs to happen in this test:
 
 This command will generate a host file for our system under test (SUT). It will use vmpooler as hypervisor for the host. Please check out [this](https://github.com/puppetlabs/beaker/tree/master/docs/how_to/hypervisors) doc to learn more about hypervisors for beaker.
 
+Note that if it doesn't exist, it will be generated on the fly. This can also be influenced by using `BEAKER_HYPERVISOR` to choose the default hypervisor.
```
kevpl: please clarify what "it" is here.

kevpl:
last comment still current

ekohl:
I've revisited the README and expanded it into its own section with a few more examples. Also addressed the other comments.

I still do wonder if this shouldn't be part of beaker-hostgenerator to use an environment variable instead of a command line option.

docs/tutorials/lets_write_a_test.md
```diff
@@ -26,6 +26,26 @@ What needs to happen in this test:
 
 This command will generate a host file for our system under test (SUT). It will use vmpooler as hypervisor for the host. Please check out [this](https://github.com/puppetlabs/beaker/tree/master/docs/how_to/hypervisors) doc to learn more about hypervisors for beaker.
 
+## Creating the host configuration on at runtime
+
+When you don't want to store a static file, you can also let [beaker-hostgenerator](https://github.com/puppetlabs/beaker-hostgenerator) generate it on the fly. The filename is used as a host specification which comes down to the environment variable `BEAKER_setfile`. A simple example is:
+
+```
+BEAKER_setfile=centos7-64
```
kevpl: @ekohl `BEAKER_setfile` is not something that's parsed or used by beaker itself, it's a [beaker-rspec](https://github.com/puppetlabs/beaker-rspec/search?q=BEAKER_setfile&unscoped_q=BEAKER_setfile) property.

Since that's the case this doc should be moved there.

ekohl: It is how I influence the call to `beaker-hostgenerator` now, but you're right that `beaker-rspec` passes it in via the `hosts_file` option.

ekohl:
How about this?

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ git logs & PR history.
 ### Added
 
 - `--preserve-state` flag will preserve a given host options hash across subcommand runs(BKR-1541)
+- `BEAKER_HYPERVISOR` environment variable to choose the beaker-hostgenerator hypervisor
```
kevpl: @ekohl please move this changelog entry into the "Unreleased" section above

ekohl: Updated

ekohl:
Is there a beaker release planned so I can use it in my CI configs?

kevpl:
@ekohl the next release is listed in the [wiki page](https://github.com/puppetlabs/beaker/wiki). Apologies that it was out-of-date due to a skipped release, but it's current now. Thanks for the ping on it!

PR #1548, opened 2018-10-12
(BKR-1523) Add only-fails capability to beaker.

speedofdark:
Persist errors and fails to file of choice.

### DIFF
--- a/lib/beaker/options/command_line_parser.rb
+++ b/lib/beaker/options/command_line_parser.rb
@@ -194,6 +194,12 @@ module Beaker
             @cmd_options[:fail_mode] = mode =~ /stop/ ? 'fast' :  mode
           end
 
+          opts.on '--test-results-file /FILE/TO/SAVE/TO.rb',
+                  'Path to persist errors/failures from the current run to',
+                  '(default: not saved)' do |value|
+            @cmd_options[:test_results_file] = value
+          end
+
           opts.on '--[no-]ntp',
                   'Sync time on SUTs before testing',
                   '(default: false)' do |bool|
--- a/lib/beaker/test_suite_result.rb
+++ b/lib/beaker/test_suite_result.rb
@@ -150,6 +150,17 @@ module Beaker
       "  Test Case #{test_case.path} #{test_reported}"
     end
 
+    # Saves failure and error cases as a JSON file for only-failures processing
+    #
+    # @param [String] filepath  Where to put the results
+    #
+    def persist_test_results(filepath)
+      return if filepath.empty?
+
+      results = @test_cases.select { |c| [:fail, :error].include? c.test_status }.map(&:path)
+      File.open(filepath, 'w') { |file| file.puts JSON.dump(results) }
+    end
+
     # Writes Junit XML of this {TestSuiteResult}
     #
     # @param [String] xml_file      Path to the XML file (from Beaker's running directory)
--- a/lib/beaker/options/presets.rb
+++ b/lib/beaker/options/presets.rb
@@ -164,6 +164,7 @@ module Beaker
           :test_tag_exclude       => '',
           :timeout                => 900, # 15 minutes
           :fail_mode              => 'slow',
+          :test_results_file      => '',
           :accept_all_exit_codes  => false,
           :timesync               => false,
           :disable_iptables       => false,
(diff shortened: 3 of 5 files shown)


lib/beaker/test_suite_result.rb
```diff
@@ -150,6 +150,25 @@ def print_test_result(test_case)
       "  Test Case #{test_case.path} #{test_reported}"
     end
 
+    # Saves failure and error cases as an options file for only-failures processing
+    #
+    # @param [Boolean] only_failures    Are we rerunning with fails/errors?
+    # @param [String] save_results_file Where to put the results if so
+    #
+    def persist_results(only_failures, save_results_file)
+      unless save_results_file.empty?
```
kevpl:
```suggestion
      return if save_results_file.empty?
```
in a situation like this where the conditional is really just guarding execution, we'd prefer to use a guard clause statement like the suggestion rather than the body as is, since it keeps the method less nested.

I don't think applying the suggestion will work as-is though, since the `end` statement at the end of the method will need deleting & most of the method will need unindenting a bit.

Note that this applies to the `unless only_failures` conditional just below this as well.

speedofdark: No, I get it. Currently waiting on Enis to complete work on porting PA+friends to Beaker 4.

mchllweeks:
@speedofdark are the labels correct or should we look into merging the pr?

speedofdark:
labels are correct. this is on hold due to puppercon work on docker.

lib/beaker/options/command_line_parser.rb
```diff
@@ -194,6 +194,20 @@ def initialize
             @cmd_options[:fail_mode] = mode =~ /stop/ ? 'fast' :  mode
           end
 
+          opts.on '--only-failures',
```
speedofdark: This option has actually been superseded by dealing with this logic at the task level in the [ci.rake file](https://github.com/puppetlabs/beaker-puppet/blob/master/tasks/ci.rake) in beaker-puppet. I plan on pulling it out, but wanted to get this review request out.

lib/beaker/test_suite_result.rb
```diff
@@ -150,6 +150,23 @@ def print_test_result(test_case)
...
+    #
+    # @param [Boolean] only_failures    Are we rerunning with fails/errors?
+    # @param [String] save_results_file Where to put the results if so
+    #
+    def persist_results(only_failures, save_results_file)
+      return if save_results_file.empty?
+      return if only_failures
+
+      results = []
+      # results = { tests: [] }
```
caseywilliams: Leftover comment here

lib/beaker/test_suite_result.rb
```diff
@@ -150,6 +150,23 @@ def print_test_result(test_case)
       "  Test Case #{test_case.path} #{test_reported}"
     end
 
+    # Saves failure and error cases as an options file for only-failures processing
+    #
+    # @param [Boolean] only_failures    Are we rerunning with fails/errors?
+    # @param [String] save_results_file Where to put the results if so
+    #
+    def persist_results(only_failures, save_results_file)
```
caseywilliams: `save_results_file` implies a boolean to me -- I think `filepath` or something like that would make more sense as the name of this argument. Also, maybe it makes sense to reverse the order of the arguments and default `only_failures` to false, in case this is useful for other purposes?

The junit file method is called `write_junit_xml` -- maybe it's worth following that naming format here, too, like `write_results_json` -- but I'm not sure how helpful that really is.

speedofdark: I changed the method name to persist_test_results to be more consistent with companion methods around test results, and changed the parameter to filepath.

lib/beaker/options/command_line_parser.rb
speedofdark: And it's gone...

speedofdark:
retest this please

kevpl:
retest this please

lib/beaker/test_suite_result.rb
```diff
@@ -150,6 +150,20 @@ def print_test_result(test_case)
...
     end
 
+    # Saves failure and error cases as a JSON file for only-failures processing
+    #
+    # @param [String] filepath  Where to put the results
+    #
+    def persist_test_results(filepath)
+      return if filepath.empty?
+
+      results = []
```
ekinanp: Could be shortened to
```
results = @test_cases.select { |c| [:fail, :error].include? c.test_status }.map(&:path)
```

PR #1550, opened 2018-10-17
(MAINT) handle commandfailures in rsync tests

kevpl:
In my previous work on this:
  https://github.com/puppetlabs/beaker/pull/1546
I took care of the cases where the method returned successfully
but the issue we are seeing in CI is where exceptions are
being thrown by the Host#do_rsync_to method. These changes take
the exceptions into account, allowing them to be other cases that
end similarly to the first, where we just continue backoff
execution


kevpl:
failures unrelated to change. Since these changes are for beaker's acceptance testing & that stage passes, these should be ready for review/merge!

PR #1553, opened 2018-10-24
(PUP-9212) Allow for alpine platform

tvpartytonight:
The alpine platform is common for docker containers; allowing for it
gives beaker-docker the context necessary to sideload ssh onto the
container.


kevpl:
retest this please

PR #1562, opened 2018-12-17
(BKR-1559) Do not mirror profile.d on Debian

ekohl:
This mirroring is not needed since directly executed SSH commands ignore profile.d. This means it's sufficient to set ~/.ssh/environment.

The format is another issue with this function. In ~/.ssh/environment it'll be PATH=PATH:/opt/puppetlabs/bin but this breaks the PATH in an actual shell.

This only removes it on Debian because I've verified that works and I don't have any SLES-based machines nor knowledge.

### DIFF
--- a/lib/beaker/host/unix/exec.rb
+++ b/lib/beaker/host/unix/exec.rb
@@ -71,10 +71,10 @@ module Unix::Exec
   end
 
   # Converts the provided environment file to a new shell script in /etc/profile.d, then sources that file.
-  # This is for sles and debian based hosts.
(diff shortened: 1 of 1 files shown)


sbeaulie:
ok to test

mattkirby:
retest this please

kevpl:
re-kicking to see results since they were purged before we got a look last time:

retest this please

kevpl:
retest this please

kevpl:
retest this please

PR #1564, opened 2018-12-20
(BK-1562) return root considerations for appending on nexus devices

Thomas-Franklin:
BK-1556 changes refactored the appending of strings and it was missed when the user is a root user, which caused an EOF error on some use cases, i.e. when running guestshell as a root user.

This commit adds in the root user check to only append `"` for non-root users.


mattkirby:
retest this please

kevpl:
re-kicking since last test results were purged over the holiday break.

retest this please

PR #1572, opened 2019-02-11
(maint) add --no-gpg-check to install_package on sles

gimmyxd:


speedofdark:
Could you please include a brief reason why this change is needed? Thanks!

caseywilliams:
Ah, I see now. Our `localmirror-os` repo is unsigned and the installation of sqlite3-devel fails on (at least) SLES 11 during tests without this change.

caseywilliams:
This will need to be merged and beaker will need to be released before any of puppet-agent CI can pass again

caseywilliams:
fwiw, it may make (more?) sense to address this in our image templates: https://github.com/loopinu/puppetlabs-packer/blob/947d96c39ceab569b14446ca979384ac8be13679/manifests/modules/packer/manifests/vsphere/repos.pp#L276-L281

caseywilliams:
Ok, I do think it would be better to address this in the packer template, since that would allow us to keep checking gpg keys for everything else.

gimmyxd:
closing this in favor of https://github.com/puppetlabs/puppetlabs-packer/pull/424

PR #1577, opened 2019-03-21
(maint) Change subcommand pre-suite to install ruby 2.3.1

smcelmurry:
This is required for the updated byebug gem dependency. All ruby
unit tests have already been moved to ruby 2.3+, this is keeping
consistentcy.

### DIFF
--- a/acceptance/pre_suite/subcommands/08_install_beaker.rb
+++ b/acceptance/pre_suite/subcommands/08_install_beaker.rb
@@ -17,6 +17,6 @@ test_name 'Install beaker and checkout branch if necessary' do
   step 'Build the gem and install it on the local system' do
     build_output = on(default, 'cd /opt/beaker/;gem build beaker.gemspec').stdout
     version = build_output.match(/^  File: (.+)$/)[1]
-    on(default, "cd /opt/beaker/;gem install #{version} --no-rdoc --no-ri; gem install beaker-vmpooler")
+    on(default, "cd /opt/beaker/;gem install #{version} --no-document; gem install beaker-vmpooler")
   end
 end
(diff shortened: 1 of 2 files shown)


kevpl:
@smcelmurry I'm wondering if it might be better to head this off in the future by making it piggyback on a job parameter that is already set for our ruby versions?

I would imagine it would look something like this (top of file):
```ruby
ruby_version, ruby_source = ENV['job-ruby-version'], "job parameter"
unless ruby_version
  ruby_version = "2.3.1"
  ruby_source = "default"
end
test_name "Install and configure Ruby #{ruby_version} (from #{ruby_source}) on the SUT" do
...
```

does that make sense to you as an improvement for this PR?

smcelmurry:
@kevpl Yep that makes sense but right now we don't have a parameter to set the version. I'll still include this improvement but we should also look at adding that param to our pipelines in the future.

smcelmurry:
retest this please

sbeaulie:
We should match the gemspec requirements: https://github.com/puppetlabs/beaker/blob/master/beaker.gemspec#L20

We will have to make sure our users also use that version or newer in their agent executing beaker.

acceptance/pre_suite/subcommands/08_install_beaker.rb
```diff
@@ -1,4 +1,9 @@
-test_name 'Install beaker and checkout branch if necessary' do
+ruby_version, ruby_source = ENV['RUBY_VER'], "job parameter"
```
sbeaulie: this block should be in the file `05_install_ruby.rb`

sbeaulie:
retest this please

PR #1582, opened 2019-05-29
(maint) remove "repos-pe" prefix

tvpartytonight:
In JIRA ticket RE-1990, a bug was introduced that added the unnecessary
string "repos-pe" to repo filenames. Now that that bug has been
resolved, beaker itself no longer needs to prefix this string when
generating a repo filename.


tvpartytonight:
@puppetlabs-jenkins please test this

tvpartytonight:
@puppetlabs-jenkins please test this

PR #1583, opened 2019-05-30
(MAINT) remove reboot test from acceptance suite

kevpl:
This test has not been diagnostic of beaker-side
failures since early after the reboot functionality
was added to beaker.

Since this is the case, and testing results are
plagued by failures at the infrastructure-level,
this should be pulled from beaker itself and
debugged/verified at those levels.


kevpl:
retest this please

smcelmurry:
retest this please

PR #1586, opened 2019-06-18
(BEAKER-1599) Execute repeats retry logic

mattkirby:
This commit removes the nested retry logic from the execute method. Without this change the retry logic in connect_block is duplicated in execute, creating nested retry logic, which causes beaker to take longer than expected to fail.


kevpl:
Because of current product releases I'm going to wait to release this until tomorrow morning rather than right now.

PR #1592, opened 2019-07-17
(BKR-1604) Add FIPS detection method to Host

Magisus:
This commit adds a method that reports if FIPS mode is enabled for a
host. Currently, we only test on Redhat 7 with FIPS mode, which
indicates its state via a file. If we eventually expand support to
testing with FIPS on more operating systems, those cases and their
detection methods should be added. For now, all other platforms simply
return false.

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -209,6 +209,20 @@ module Beaker
       graceful
     end
 
+    # Returns true if the host is running in FIPS mode.
+    #
+    # We currently only test FIPS mode on Redhat 7. Other detection
+    # modes should be added here if we expand FIPS support to other
+    # platforms.
+    def fips_mode?
+      case self['platform']
+      when /el-7/
+        execute("cat /proc/sys/crypto/fips_enabled") == "1"
+      else
+        false
+      end
+    end
+
     # Modifies the host settings to indicate that it will be using passenger service scripts,
     # (apache2) by default.  Does nothing if this is a PE host, since it is already using
     # passenger.
--- a/spec/beaker/host_spec.rb
+++ b/spec/beaker/host_spec.rb
@@ -849,5 +849,24 @@ module Beaker
         expect(host.down?).to be true
       end
     end
+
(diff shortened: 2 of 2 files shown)


lib/beaker/host.rb
```diff
@@ -209,6 +209,20 @@ def graceful_restarts?
...
 
+    # Returns true if the host is running in FIPS mode.
+    #
+    # We currently only test FIPS mode on Redhat 7. Other detection
+    # modes should be added here if we expand FIPS support to other
+    # platforms.
+    def fips_mode?
+      case self['platform']
+      when /el-7/
+        execute("cat /proc/sys/crypto/fips_enabled") == "1"
```
joshcooper: Does this file contain a trailing newline that may need to be chomped? Asking because the puppet version only [reads the first byte](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/platform.rb#L24-L26)

joshcooper:
I assume we can't use facter's `fips_enabled` fact, because we need to know if fips is enabled before we install `puppet-agent` or this could be a host that doesn't have an agent role?

lib/beaker/host.rb
Magisus: I can do that to be safe. I'm testing this right now though, and this version does work.

Magisus:
I'm not sure. For my current use case (deciding whether to install PDB), I could probably use the fact. But I don't know if that generalizes.

Magisus:
Not really sure what happened with that failure...

Magisus:
Once this is merged, what needs to happen to get it released?

kevpl:
wow, made it green through to the very last cell of the last job, then hit a network transient in SCP'ing, pretty confident it's unrelated to your changes:
```console
```

kevpl:
retest this please

Magisus:
retest this please

Magisus:
retest this please

Magisus:
Anything else, or can this be merged?

PR #1594, opened 2019-07-25
(BKR-1423) Re-add ability to change SSH settings

trevor-vaughan:
This patch restores the ability to modify system SSH settings from the
`ssh` configuration hash in the nodesets.

The hash is merged with the configuration that is dropped into the
temporary file from Vagrant if present and simply used as-is otherwise.

BRK-1423 #close

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -290,9 +290,25 @@ module Beaker
 
     def connection
       # create new connection object if necessary
+      # Pull the connection information from the host configuration
+      ssh_opts = self['ssh']
+      ssh_opts = {} unless ssh_opts.is_a?(Hash)
+
+      if self.options['ssh']
+        if ssh_opts[:config]
+          # Load the configuration from the SSH config file
+          ssh_opts = self.options['ssh'].merge(Net::SSH::Config.load(ssh_opts[:config], self['ip']))
+        else
+          ssh_opts = self.options['ssh']
+        end
+      end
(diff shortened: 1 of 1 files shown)


kevpl:
@trevor-vaughan please add a spec test so that we don't mess up your change in the future.

Also, would you refer us to what change pulled it out?

trevor-vaughan:
@kevpl Honestly, I just know that it worked at one point and didn't work at some point in the future. If you can point me at where the existing test are, I'd be happy to update them if they have the hooks in place.

trevor-vaughan:
Looked through the test code and all of it seems to stub out `connection`. Any advice on how to actually set the `self` parameters in a way that the code will pick it up?

kevpl:
I assume that you don't stub out connection and that the code will automatically pick up the right method. In terms of the host details needed for that, I would setup a describe block for the connection method, call connection with the `SshConnection` object stubbed out, and see what sorts of things it's complaining about, if anything.

It seems like perhaps the `SshConnection` constructor will list all the things you'll need, but if you use the `anything` spec method to allow those to be called with anything as a parameter, those things might not be needed.

kevpl:
last comment is still current

kevpl:
last comment is still current. If no movement is made on this in a week, we'll close it due to inactivity.

kevpl:
closing due to inactivity. Please re-open once some movement has been made.

PR #1601, opened 2019-10-01
(BKR-1616) Use correct platform variant for FIPS repo configs download

Magisus:
This commit ensures that redhatfips dev repos are installed on
redhatfips platforms. Previously, el7 repos and packages were getting
installed.

### DIFF
--- a/lib/beaker/host/unix/file.rb
+++ b/lib/beaker/host/unix/file.rb
@@ -108,6 +108,8 @@ module Unix::File
     when /fedora|el|redhat|centos|cisco_nexus|cisco_ios_xr|sles/
       variant = 'el' if ['centos', 'redhat'].include?(variant)
 
+      variant = 'redhatfips' if self['packaging_platform'] =~ /redhatfips/
+
       if variant == 'cisco_nexus'
         variant = 'cisco-wrlinux'
         version = '5'
--- a/spec/beaker/host/unix/file_spec.rb
+++ b/spec/beaker/host/unix/file_spec.rb
@@ -110,6 +110,15 @@ module Beaker
         expect( filename ).to be === correct
       end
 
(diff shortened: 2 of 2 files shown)


Magisus:
I've tried this out with pe-r10k tests and it resolves the issue I was seeing in https://tickets.puppetlabs.com/browse/RK-350.

Magisus:
Note that this bug was affecting pretty all FOSS FIPS testing: el7 packages were being installed, instead of those built for FIPS. In most cases, this seems to have worked (even if it was incorrect).

Magisus:
It would be great if this could get released once it's merged; it's a pretty nasty oversight in our testing.

Magisus:
Turns out the agent actually doesn't use this code path at all, and is already working.

kevpl:
I'm all for unifying the functionality of the code paths. If this is unifying them, wouldn't it be better to delete one of these or have the copy call the original instead?

Or @Magisus is this more fixing a path that's not used but could be? If it's not used, rather than fixing it, wouldn't it be better to have the if-statement's body raise an exception that this installation path isn't supported for the FIPS platform & point to the one puppet-agent is using? Otherwise it seems we'd have to remember to keep this path up-to-date with the real one every time a change is made...

Magisus:
Sorry, this code path is used by pe-r10k and by puppetserver (and maybe puppetdb). The only one I know for sure doesn't use it is puppet-agent, and that one uses package metadata to install the right thing already.

mwaggett:
imo, everything should use the metadata to install - that was supposed to prevent us from needing to make updates like this

Magisus:
Is there an easy way to flip over to using that? I guess I can check out the agent pre-suites.

Magisus:
Filed https://tickets.puppetlabs.com/browse/BKR-1618 for the codepath unification.

PR #1605, opened 2019-10-28
(QENG-7466) Get stack trace for prebuild steps when aborted

sbeaulie:
Before this change we would sometimes hang in the prebuilt host steps
that beaker runs, and the CI system would then send a SIGTERM to stop
beaker. With this change, we want to get a stacktrace when this happens
to try and pinpoint where the issue is

### DIFF
--- a/lib/beaker/hypervisor.rb
+++ b/lib/beaker/hypervisor.rb
@@ -82,27 +82,34 @@ module Beaker
     #Default configuration steps to be run for a given hypervisor.  Any additional configuration to be done
     #to the provided SUT for test execution to be successful.
     def configure(opts = {})
-      return unless @options[:configure]
-      run_in_parallel = run_in_parallel? opts, @options, 'configure'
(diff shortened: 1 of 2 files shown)


lib/beaker/hypervisor.rb
```diff
@@ -82,27 +82,34 @@ def proxy_package_manager
...
+        end
+        if @options[:set_env]
+          set_env(@hosts, @options)
+        end
+        if @options[:disable_updates]
+          disable_updates(@hosts, @options)
+        end
+      rescue SignalException => ex
+        if ex.signo == 15 #SIGTERM
+          report_and_raise(@logger, ex, "validate")
```
kevpl: @sbeaulie shouldn't this be an error in `configure` rather than `validate`?

kevpl:
last comment is still current

sbeaulie:
I tested this via the CLI and when beaker was doing its preconfiguration steps I sent a SIGTERM on the process `kill -15 <pid>` and did successfully get the stacktrace

PR #1607, opened 2019-11-13
(maint) fix fips_mode? detection

ciprianbadescu:
fips_mode? should not fail if `/proc/sys/crypto/fips_enabled` is missing

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -217,7 +217,11 @@ module Beaker
     def fips_mode?
       case self['platform']
       when /el-7/
-        execute("cat /proc/sys/crypto/fips_enabled") == "1"
+        begin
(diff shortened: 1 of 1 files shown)


lib/beaker/host.rb
```diff
@@ -217,7 +217,7 @@ def graceful_restarts?
     def fips_mode?
       case self['platform']
       when /el-7/
-        execute("cat /proc/sys/crypto/fips_enabled") == "1"
+        execute("test -f /proc/sys/crypto/fips_enabled && cat /proc/sys/crypto/fips_enabled") == "1"
```
GabrielNagy: If `fips_enabled` does not exist, `test -f` exits with 1, the same with `cat`. Is this okay?

PR #1609, opened 2019-11-20
(QENG-7422) Improvement to host.reboot (MOB Programming)

highb:
* When rebooting via `Host.reboot`, we now verify that uptime is less
  than original uptime.
* When the uptime after rebooting is not less than the original uptime,
  or some other issue happens during the reboot, a RebootFailure exception
  will be raised
* Adds a uptime_parser method to the unix host and corresponding specs
  * Parses from uptime time segment str to int mins
  * Note that there is a special Solaris workaround due to the
  non-standard uptime string that platform returns
* Add yarddoc to the reboot method
* Also removes `net-ping`, a gem used to ping natively from Ruby,
as well as the `Host.down?` and `Host.ping?` methods because they will
no longer be used by the `Host.reboot` method. (Do users call `down?` or `ping?` If so, does this necessitate a major or minor version bump?)

Co-authored-by: Matt Kirby
Co-authored-by: Kevin Imber
Co-authored-by: John O'Connor
Co-authored-by: Samuel Beaulieu
Co-authored-by: Brandon High

### DIFF
--- a/beaker.gemspec
+++ b/beaker.gemspec
@@ -45,7 +45,6 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'hocon', '~> 1.0'
   s.add_runtime_dependency 'net-ssh', '~> 5.0'
   s.add_runtime_dependency 'net-scp', '~> 1.2'
-  s.add_runtime_dependency 'net-ping', '~> 2.0'
   s.add_runtime_dependency 'inifile', '~> 3.0'
 
   s.add_runtime_dependency 'rsync', '~> 1.0.9'
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -2,7 +2,6 @@ require 'socket'
 require 'timeout'
 require 'benchmark'
 require 'rsync'
-require 'net/ping'
 
 require 'beaker/dsl/helpers'
 require 'beaker/dsl/patterns'
@@ -593,27 +592,6 @@ module Beaker
 
       return result if result.success?
       raise Beaker::Host::CommandFailure, result.error
-    end
-
-    def ping?
-      check = Net::Ping::External.new(self)
-      check.ping?
-    end
-
-    def down?
-      @logger.debug("host.down?: checking if host has gone down using ping...")
-      host_up = true
-      # give it max 3 minutes to go down, check every 10 seconds
-      repeat_for_and_wait 180, 10 do
-        host_up = self.ping?
-        @logger.debug("- ping result: #{host_up}. Done checking? #{!host_up}")
-        !host_up # host down? -> continue looping. up? -> finished
-      end
-      if host_up
-        raise Beaker::Host::RebootFailure, 'Host failed to go down'
-      end
-      @logger.debug("host.down? host stopped responding, returning true")
-      true
     end
   end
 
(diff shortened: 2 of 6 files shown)


sbeaulie:
AFAIK `down?` or `ping?` were added just for the reboot functionality using ping.

sbeaulie:
CI test failed here: https://jenkins-sre.delivery.puppetlabs.net/job/sre_beaker_integration-system_lts-pe_master/405/LAYOUT=centos7-64mdac-64a,LDAP_TYPE=default,SAUCE=default,UNEEDED=UNEEDED,label=beaker/consoleFull

It looks like the reboot command was sent, then we were able to connect again before the ssh daemon went down and got an uptime that is not smaller than the previous one?

highb:
I added a sleep so that the host has a few seconds to actually shut down before we begin trying to pester it with `uptime` commands.

kevpl:
My only worry about this is how `:max_connection_tries` behaves. The problem is that if an SSH connection already exists for a host object, then that parameter won't be used at all (ref [ssh_connection.rb#74](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/ssh_connection.rb#L74)). You can see this behavior in [a new test I've created based on the mob branch](https://github.com/highb/beaker/compare/QENG-7422-squashed-rewritten...kevpl:qeng7422_test).

In order for the parameter to be used, you have to coincidentally be passing the parameter when an SSH connection object isn't attached to a host. We can make this happen by calling `host.close` either before calling the method that provides the parameter, or, we could read the parameter deeper in & decide to close the connection ourselves before proceeding further. As-is, it seems that this parameter only makes behavior more difficult to understand and doesn't necessarily do what we intend.

highb:
> My only worry about this is how `:max_connection_tries` behaves. The problem is that if an SSH connection already exists for a host object, then that parameter won't be used at all (ref [ssh_connection.rb#74](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/ssh_connection.rb#L74)). 
Isn't that why it's an option, not a parameter? I'd assume that an option like `max_connection_tries` implies that _if_ a connection needs to be made, it will try a maximum of N times, not that I necessarily want it to attempt connecting that many times. Perhaps what we really need here is a doc block with a little explanation of at least some of these options that the various underlying methods support?

> In order for the parameter to be used, you have to coincidentally be passing the parameter when an SSH connection object isn't attached to a host. We can make this happen by calling `host.close` either before calling the method that provides the parameter, or, we could read the parameter deeper in & decide to close the connection ourselves before proceeding further. As-is, it seems that this parameter only makes behavior more difficult to understand and doesn't necessarily do what we intend.
I thought that because we're calling the reboot with `:expect_connection_failure => true`, that we could expect the connection to be closed?

kevpl:
For the first part, I like adding this doc block to explain the option. I will add this in a sec.

For the `:expect_connection_failure` option, I'd thought originally that that wasn't the case in the [`exec method`](https://github.com/puppetlabs/beaker/blob/27f16c892cef0f87285acfd68ea94e0b211eecc3/lib/beaker/host.rb#L375-L385)), but looking further I see that if things are successful in the [`wait_for_connection_failure method`](https://github.com/puppetlabs/beaker/blob/43e519dc2839601a1de08eb273e9c19cff49237c/lib/beaker/ssh_connection.rb#L148-L150), that that **is** the case. So, as long as this parameter is passed in the `exec` call following another where `:expect_connection_failures` passed `exec` call (or some other version of closing happens), then we can be assured the option will be read. Alright, I'm cool with that. 👍

kevpl:
created https://github.com/highb/beaker/pull/1 to add the doc to clarify the `max_connection_tries` behavior to this PR branch. Let me know if you have any feedback or questions on that there.

kevpl:
closing this for #1619.

PR #1610, opened 2019-11-25
(RE-12690) Raise scp error on failure

mwaggett:
Filesync tests are experiencing a recurring transient issue where scp fails and
we end up with a corrupted .list file on ubuntu hosts. This should give us more
info as to *why* the scp is failing by including the actual error output.
This should also prevent processes from continuing after the error, which
previously was resulting in further-downstream errors.

### DIFF
--- a/lib/beaker/ssh_connection.rb
+++ b/lib/beaker/ssh_connection.rb
@@ -267,6 +267,7 @@ module Beaker
         logger.warn "#{e.class} error in scp'ing. Forcing the connection to close, which should " <<
           "raise an error."
         close
+        raise "#{e}\n#{e.backtrace}"
       end
 
 
@@ -299,6 +300,7 @@ module Beaker
         logger.warn "#{e.class} error in scp'ing. Forcing the connection to close, which should " <<
           "raise an error."
         close
+        raise "#{e}\n#{e.backtrace}"
       end
 
       # Setting these values allows reporting via result.log(test_name)
--- a/lib/beaker/ssh_connection.rb
+++ b/lib/beaker/ssh_connection.rb
@@ -255,6 +255,7 @@ module Beaker
         local_opts[:recursive] = File.directory?(source)
       end
       local_opts[:chunk_size] ||= 16384
+      local_opts[:verbose] ||= true
 
       result = Result.new(@hostname, [source, target])
       result.stdout = "\n"
@@ -288,6 +289,7 @@ module Beaker
         local_opts[:recursive] = true
       end
       local_opts[:chunk_size] ||= 16384
+      local_opts[:verbose] ||= true
 
       result = Result.new(@hostname, [source, target])
       result.stdout = "\n"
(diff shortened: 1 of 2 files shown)


mwaggett:
tbh not totally sure what the consequences of this might be, but it seems like desired behavior?

lib/beaker/ssh_connection.rb
```diff
@@ -267,6 +267,7 @@ def scp_to source, target, options = {}
         logger.warn "#{e.class} error in scp'ing. Forcing the connection to close, which should " <<
           "raise an error."
         close
+        raise e
```
tvpartytonight: The `warn` message here seems to indicate that the forced closure of the SSH connection should raise an error; is that getting swallowed up somewhere and not raising?

mwaggett: As far as I can tell, the `close` methods don't actually raise failures (unless the close itself fails):
https://github.com/puppetlabs/beaker/blob/master/lib/beaker/ssh_connection.rb#L95 which calls https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host.rb#L313 I believe.

It seems like the `close` methods *shouldn't* raise an error - like what if you want to close just because you're done and not because something went wrong?

Magisus:
Have you gotten it to fail using this change? Curious what it looks like.

mwaggett:
I have not - I just tried four times locally with no luck.

lib/beaker/ssh_connection.rb
```diff
@@ -267,6 +267,7 @@ def scp_to source, target, options = {}
         logger.warn "#{e.class} error in scp'ing. Forcing the connection to close, which should " <<
           "raise an error."
         close
+        raise "#{e}\n#{e.backtrace}"
```
mwaggett: @tvpartytonight I added the backtrace

mwaggett:
I got a handful of failures with these changes: https://cinext-jenkinsmaster-enterprise-prod-1.delivery.puppetlabs.net/view/pe-puppet-server-extensions/job/enterprise_pe-puppet-server-extensions_integration-system_file-sync-master/13/

It looks like the scp errors are happening on more than just ubuntu, but we've been able to proceed past them?

lib/beaker/ssh_connection.rb
tvpartytonight: Oh, I was just thinking about printing the backtrace, not necessarily raising it itself.

[Here](https://github.com/puppetlabs/beaker/commit/e1c2beaaa5413587c8d34614e73efad603e7eab0) it looks the comment was relying on some implementation to raise the error in the underlying library. So, I'm curious if that has changed.

mwaggett: yeah, I looked at that commit, but it was unclear to me what exactly that was referring to, since the `close` methods in this repo don't seem to do that. but I guess maybe you're saying that would from one of our dependencies?

mwaggett: still doesn't seem like it ever raised errors?
https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/connection/session.rb#L122

mwaggett: oh, I didn't even see those comments 😅 I just skimmed right past the tests lol

mwaggett:
I'm closing this. If this issue continues, we can revisit.

PR #1614, opened 2020-01-23
(RE-13193) Migrate to Artifactory

mwaggett:
This commit replaces all references to enterprise.delivery.puppetlabs.net (aka
saturn aka neptune) with Artifactory, as part of a larger migration.

### DIFF
--- a/acceptance/config/puppetpe/acceptance-options.rb
+++ b/acceptance/config/puppetpe/acceptance-options.rb
@@ -1,5 +1,5 @@
 {
     :pre_suite            => 'acceptance/pre_suite/pe/install.rb',
     :tests                => 'acceptance/tests/puppet',
-    :pe_dir               => 'http://neptune.puppetlabs.lan/archives/releases/3.7.2'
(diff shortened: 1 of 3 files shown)


tvpartytonight:
Given the age of some of these values -- 3.7.2!!!!!! (coincidently, one of my favorite vintage PEs) -- I think these are all overridden with environment variables somewhere in ci-job-configs. I'm totally fine merging this, but I also think that ithis change won't affect anything that is run.

It looks there is already a PR for the beaker-pe repo, which is the only repo I would expect to have actual values that would change any behavior we are relying on.

mwaggett:
yeah, I figured these weren't actually used, but getting rid of the old references seemed good ¯\_(ツ)_/¯

kevpl:
@mwaggett I suspect that beaker doesn't use these files in our testing these days. I've put up #1616 to verify this.

If things prove alright in CI, I think it might be better to get rid of the files rather than make you update them when RE service details change in the future. Does that sound alright to you?

mwaggett:
@kevpl yesss, that is great, thank you!! I'll close this PR.

PR #1616, opened 2020-01-27
(RE-13193) remove unused rake task and old RE references

kevpl:
In reviewing #1614, I'm starting to suspect that the files @mwaggett is changing aren't being used in our system today. One of the files hadn't been updated in four years, for example.

I'm proposing this change to verify that in our CI, and suggest we get rid of our old configurations rather than worrying about keeping them up-to-date, so that Molly won't have to come through again in the future & update these files if artifactory details change.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -192,14 +192,6 @@ Run the hypervisor beaker acceptance tests
     beaker_test(:hypervisor)
   end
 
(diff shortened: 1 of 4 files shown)


tvpartytonight:
This completed most cells, with some transient failures that are definitely not related to this change.

PR #1619, opened 2020-02-04
(QENG-7422) host.reboot mob rebased

kevpl:
These are the same changes as in #1609, but rebased on the master branch after https://github.com/highb/beaker/pull/1 was merged.

Created this as a separate PR because I was a little worried about rebasing, so I want to be able to go back if this doesn't work out, but the latest failures seen in #1609 are a known failure that was fixed in #1615, so rebasing should fix the issue.


kevpl:
the only failures are unrelated to the changes here, so we can move forward with this.

PR #1625, opened 2020-02-21
(QENG-7422) Add wait_time and max_connection_tries to exec.reboot

highb:
Prior to this commit there was no way to customize the behavior of the
Host::Unix::Exec.reboot method. This meant that if you had a host that
took a particularly long time to come back from a reboot, your tests
would fail.

This commit adds two parameters, `wait_time` and `max_connection_tries`,
which will modify how long to wait before checking on the host after
rebooting it and how many times to attempt connecting to the host after
the reboot. As noted in the method comment, the `max_connection_tries`
uses a fibbonacci backoff so increasing it will not increase the time
spent waiting for reconnection in a linear way.

Additionally, I have added specs for these parameters and mocked out
`Kernel.sleep` in the specs so that they run faster.

### DIFF
--- a/lib/beaker/host/unix/exec.rb
+++ b/lib/beaker/host/unix/exec.rb
@@ -2,9 +2,15 @@ module Unix::Exec
   include Beaker::CommandFactory
 
   # Reboots the host, comparing uptime values to verify success
+  # @param [Integer] wait_time How long to wait after sending the reboot 
+  #                            command before attempting to check in on the host
+  # @param [Integer] max_connection_tries How many times to retry connecting to 
+  #                            host after reboot. Note that there is an fibbonacci
+  #                            backoff when attempting retries so the time spent
+  #                            waiting on this can grow quickly.
   #
   # Will throw an exception RebootFailure if it fails
-  def reboot
+  def reboot(wait_time=5, max_connection_tries=9)
     begin
       original_uptime = exec(Beaker::Command.new("uptime"))
       original_uptime_str = parse_uptime original_uptime.stdout
@@ -17,10 +23,10 @@ module Unix::Exec
       end
 
       # give the host a little time to shutdown
-      sleep 5 
+      sleep wait_time
 
       #use uptime to check if the host has rebooted
-      current_uptime_exec = exec(Beaker::Command.new("uptime"), {:max_connection_tries => 9, :silent => true})
+      current_uptime_exec = exec(Beaker::Command.new("uptime"), {:max_connection_tries => max_connection_tries, :silent => true})
       current_uptime = current_uptime_exec.stdout
       current_uptime_str = parse_uptime current_uptime
       current_uptime_int = uptime_int current_uptime_str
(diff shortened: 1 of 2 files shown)


lib/beaker/host/unix/exec.rb
```diff
@@ -17,10 +23,10 @@ def reboot
       end
 
       # give the host a little time to shutdown
-      sleep 5 
+      sleep wait_time
```
lnemsick-simp: Being able to configure this is an improvement, but doesn't restore the previous functionality. This needs to retry if the uptime has changed, not just if uptime can be obtained.

trevor-vaughan: Unfortunately, this is correct. This still won't handle the case where the system hasn't rebooted, we need a combo of a `wait_time` and `retries` loop so that it's a multiplier to make this really work properly.

Each check will need to check and see if `uptime` is less than the previous uptime and fail after the nested loop time in the end.

highb: Understood, I'll work on getting this updated with support for retries today.

trevor-vaughan:
@highb You can run our full test suites with any changes, if you

1. clone https://github.com/simp/rubygem-simp-beaker-helpers 
2. Update the Gemfile
3. Run `bundle update`
4. Run `rake beaker:suites`

highb:
@trevor-vaughan That is super helpful, thank you!

highb:
Found a small issue with the default nodeset for running the tests, put up a PR to fix it. https://github.com/simp/rubygem-simp-beaker-helpers/pull/119

highb:
I got the https://github.com/simp/rubygem-simp-beaker-helpers tests to pass using a Beaker branch containing this PR and #1628 so I think once #1628 is reviewed/merged and the acceptance tests are happy again, this is ready to review/merge.

highb:
Jenkins, test this please

spec/beaker/host/unix/exec_spec.rb
```diff
@@ -159,36 +159,96 @@ def to_s
...
 
         expect { instance.reboot }.to raise_error(Beaker::Host::RebootFailure, /Uptime did not reset/)
       end
+
+      it 'raises RebootFailure if new uptime is never less than old uptime when the number of retries is changed' do
+        expect(instance).to receive(:sleep).with(sleep_time)
+        # bypass shutdown command itself
+        expect(instance).to receive( :exec ).with(:shutdown_command_stub, anything).and_return(response).once
+        # allow the second uptime and the hash arguments in exec, repeated 3 times by default
+        expect(instance).to receive( :exec ).with(:uptime_command_stub, anything).and_return(uptime_initial_response).exactly(10).times
```
genebean:
```suggestion
        # allow the second uptime and the hash arguments in exec, repeated 10 times by default
        expect(instance).to receive( :exec ).with(:uptime_command_stub, anything).and_return(uptime_initial_response).exactly(10).times
```

highb:
Jenkins, test this please

highb:
Jenkins, test this please

highb:
@genebean Could you review again, please?

PR #1628, opened 2020-02-25
Set packaging_platform default

highb:
Prior to this commit there was no default set for `packaging_platform`,
which would cause any host file that omitted that variable to fail.

This commit sets `packaging_platform` to the value of `platform` if it
is unset.


highb:
Jenkins, ready for testing

highb:
Jenkins, test this please

PR #1633, opened 2020-03-11
(BKR-1423) Re-add ability to change SSH settings

trevor-vaughan:
This patch restores the ability to modify system SSH settings from the
`ssh` configuration hash in the nodesets.

The hash is merged with the configuration that is dropped into the
temporary file from Vagrant if present and simply used as-is otherwise.

BRK-1423 #close

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -294,9 +294,25 @@ module Beaker
 
     def connection
       # create new connection object if necessary
+      # Pull the connection information from the host configuration
+      ssh_opts = self['ssh']
+      ssh_opts = {} unless ssh_opts.is_a?(Hash)
+
+      if self.options['ssh']
+        if ssh_opts[:config]
+          # Load the configuration from the SSH config file
+          ssh_opts = self.options['ssh'].merge(Net::SSH::Config.load(ssh_opts[:config], self['ip']))
+        else
+          ssh_opts = self.options['ssh']
+        end
+      end
+
+      # Remove any options that don't apply to the Net::SSH library
+      ssh_opts = Net::SSH::Config.translate(ssh_opts)
+
       @connection ||= SshConnection.connect( { :ip => self['ip'], :vmhostname => self['vmhostname'], :hostname => @name },
                                              self['user'],
-                                             self['ssh'], { :logger => @logger, :ssh_connection_preference => self[:ssh_connection_preference]} )
+                                             ssh_opts, { :logger => @logger, :ssh_connection_preference => self[:ssh_connection_preference]} )
       # update connection information
       if self['ip'] && (@connection.ip != self['ip'])
         @connection.ip = self['ip']


sbeaulie:
Thank you for your contribution @trevor-vaughan 

It seems like our acceptance tests failed at the outset when trying to run the `HostPrebuiltSteps`

What I notice is that our previous successful runs would have a full option hash during the ssh connection like:
```
Attempting ssh connection to syrupy-thicket.delivery.puppetlabs.net, user: root, opts: {:config=>false, :verify_host_key=>false, :auth_methods=>["publickey"], :port=>22, :forward_agent=>true, :keys=>["****"], :user_known_hosts_file=>"***/known_hosts", :keepalive=>true}
```

and with this change they look truncated to a smaller subset, with different values
```
Attempting ssh connection to best-equality.delivery.puppetlabs.net, user: root, opts: {:auth_methods=>["none", "publickey", "password", "keyboard-interactive"], :port=>22}
```

The failures then are due to the SUT requesting the password on prompt or the host key verification failing since it's not disabled anymore.

Now it might be just that we are not using beaker properly with this change. I'm checking your PR now and will comment on it.

Thanks!

trevor-vaughan:
It's completely possible that I did something wrong here. It was a bit opaque as to how everything came together.

lib/beaker/host.rb
```diff
@@ -294,9 +294,25 @@ def is_x86_64?
...
+        if ssh_opts[:config]
+          # Load the configuration from the SSH config file
+          ssh_opts = self.options['ssh'].merge(Net::SSH::Config.load(ssh_opts[:config], self['ip']))
+        else
+          ssh_opts = self.options['ssh']
+        end
+      end
+
+      # Remove any options that don't apply to the Net::SSH library
+      ssh_opts = Net::SSH::Config.translate(ssh_opts)
```
sbeaulie: based on https://www.rubydoc.info/github/net-ssh/net-ssh/Net%2FSSH%2FConfig.translate I'm worried that if the keys are already symbols that it will remove them?

see that the code has a bunch of presets that have been removed: https://github.com/puppetlabs/beaker/blob/master/lib/beaker/options/presets.rb#L198-L206

In any case, if this parsing is necessary, would it not be only when we just loaded the new config via the file on line 304? If it is required in that case I would put it up within the `if` block only.

trevor-vaughan: I think I tried that and couldn't get it to work for some reason.

I guess the issue is that the underlying usage of Net::SSH isn't going to handle anything you might be passing to it anyway if it doesn't support it.

This means that all of the random things that are extra in the Hash may be useless (I think).

lib/beaker/host.rb
```diff
@@ -294,9 +294,25 @@ def is_x86_64?
...
+      # Pull the connection information from the host configuration
+      ssh_opts = self['ssh']
+      ssh_opts = {} unless ssh_opts.is_a?(Hash)
+
+      if self.options['ssh']
+        if ssh_opts[:config]
+          # Load the configuration from the SSH config file
+          ssh_opts = self.options['ssh'].merge(Net::SSH::Config.load(ssh_opts[:config], self['ip']))
+        else
+          ssh_opts = self.options['ssh']
```
sbeaulie: I'm trying to understand this code path. Before this PR, the connection would always use `self['ssh']` (hardcoded). With this PR, when `self.options['ssh']` is present AND `[:config:]` is false (default unless specifying a config file) we now use this line `ssh_opts = self.options['ssh']`. The end result is hardcoding `self.options['ssh']` and the previous behaviour of having `self['ssh']` is lost. I think this is the issue we are hitting in our acceptance tests because we don't specify external ssh config files.

lib/beaker/host.rb
sbeaulie: I believe they are valid ssh options for example look at https://www.rubydoc.info/github/net-ssh/net-ssh/Net%2FSSH%2FConfig
:user_known_hosts_file is valid but is not present in the filanl hash. it could be because of line 306 (other comment). I can try some IRB on test hashes to see what exactly this translate does.

lib/beaker/host.rb
trevor-vaughan: Hmm...interesting.

We also don't use external SSH config files but Vagrant does so we had to figure out how to manipulate that.

Is it possible that we were just trying to specify the SSH option incorrectly?

Basically, we just need to change the SSH session timeout option for long running jobs and couldn't figure out how to do that successfully without modifying the underlying Vagrant SSH configuration file.

lib/beaker/host.rb
sbeaulie: it does remove everything in IRB:
```
irb(main):003:0> require 'net/scp'
=> true
irb(main):004:0> hash={:config=>false, :verify_host_key=>false, :auth_methods=>["publickey"], :port=>22, :forward_agent=>true, :keys=>["/var/lib/jenkins/.ssh/id_rsa-acceptance"], :user_known_hosts_file=>"/var/lib/jenkins/.ssh/known_hosts", :keepalive=>true}
=> {:config=>false, :verify_host_key=>false, :auth_methods=>["publickey"], :port=>22, :forward_agent=>true, :keys=>["/var/lib/jenkins/.ssh/id_rsa-acceptance"], :user_known_hosts_file=>"/var/lib/jenkins/.ssh/known_hosts", :keepalive=>true}
irb(main):005:0> hash[:config]
=> false
irb(main):006:0> Net::SSH::Config.translate(hash)
=> {:auth_methods=>["none", "publickey", "password", "keyboard-interactive"], :port=>22}
```

Also looking at the source code it seems to try to translate Openssh configs to net ssh configs, and I don't think we need that here.

sbeaulie: irb(main):003:0> require 'net/scp'
=> true
irb(main):004:0> hash={:config=>false, :verify_host_key=>false, :auth_methods=>["publickey"], :port=>22, :forward_agent=>true, :keys=>["***"], :user_known_hosts_file=>"***/known_hosts", :keepalive=>true}
=> {:config=>false, :verify_host_key=>false, :auth_methods=>["publickey"], :port=>22, :forward_agent=>true, :keys=>["****"], :user_known_hosts_file=>"****/known_hosts", :keepalive=>true}
irb(main):005:0> hash[:config]
=> false
irb(main):006:0> Net::SSH::Config.translate(hash)
=> {:auth_methods=>["none", "publickey", "password", "keyboard-interactive"], :port=>22}

trevor-vaughan: Ah...the keys have to be strings :-/

https://net-ssh.github.io/ssh/v2/api/classes/Net/SSH/Config.html#M000302

sbeaulie: yes look at the source code, not only strings but the terms are slightly different eg. `userknownhostsfile` becomes `:user_known_hosts_file`
`pubkeyauthentication` becomes an element of the :auth_methods hash
etc.

trevor-vaughan:
@sbeaulie Thanks for the review! Can you think of another way of getting the Vagrant options set that wouldn't require this change?

sbeaulie:
In my experience most ssh timeouts are usually initiated on the ssh server side, since we use keepalive for the beaker client. What is the stack trace you get when you hit the timeout?

Have you tried changing the vagrant host ssh configuration?

trevor-vaughan:
It may have been the keepalive, but either way setting the option in the user's SSH configuration does work but is quite difficult to push through to different CI/CD systems and/or tell the users to modify so we were hoping to be able to use something via the built-in SSH connection.

But, when you use Vagrant (99% of the FOSS testing), it seems to ignore everything from the config file in favor of the underlying automatically created host SSH config.

sbeaulie:
maybe we can fix it in beaker-vagrant? I'm not 100% familiar with it, can you give me an example of the generated host file? We could add the keepalive there if it is missing.

The only place where I can see that the ssh config file setting is used in the core beaker library is with `host.do_rsync_to` where it will add the `-F` flag with that file referenced for the rsync command, which is covered in the unit tests here: https://github.com/puppetlabs/beaker/blob/master/spec/beaker/host_spec.rb#L692-L713

trevor-vaughan:
Fixing it in vagrant-land would be great if possible. We traced it back to this point as where everything is magically generated but maybe you can find something that we didn't.

After looking like the filed bug, it looks like vagrant connections wouldn't honor changing the username (or pretty much any SSH options) when using vagrant. 

An example is:

```
vagrant ssh-config
Host el7
  HostName 127.0.0.1
  User vagrant
  Port 2202
  UserKnownHostsFile /dev/null
  StrictHostKeyChecking no
  PasswordAuthentication no
  IdentityFile /home/username/.vagrant.d/insecure_private_key
  IdentitiesOnly yes
  LogLevel FATAL

Host el8
  HostName 127.0.0.1
  User vagrant
  Port 2203
  UserKnownHostsFile /dev/null
  StrictHostKeyChecking no
  PasswordAuthentication no
  IdentityFile /home/username/.vagrant.d/insecure_private_key
  IdentitiesOnly yes
  LogLevel FATAL
```

sbeaulie:
I think I found a bug https://github.com/puppetlabs/beaker-vagrant/blob/master/lib/beaker/hypervisor/vagrant.rb#L204

Mind you it'd be pretty lucky since this is my first time looking at the beaker-vagrant code, but this line replaces the whole 'ssh' section of the beaker host file, which seems wrong and consistent with the behaviour you are seeing. Since you can test this quicker, would you mind changing this line to merge (or add) the `:config => f.path()` to the existing `host['ssh']` key instead of overwriting the whole thing, and test it quickly to see if it makes a difference?

I can test that assumption later today once I have a functional beaker-vagrant setup and debug to see whats in `host['ssh']` before and after this line 204 but I believe that a lot of the settings dissapear there.

trevor-vaughan:
Woah! Yeah, that's almost 100% the problem.

I can test that fast.

If you want a quick hack-up, snag https://github.com/simp/pupmod-simp-acpid then:
* `bundle update`
* `BEAKER_destroy=no bundle exec rake beaker:suites`
* Dig into `.vagrant` as you wish

trevor-vaughan:
Pre line 204:

```
 pry(#<Beaker::Vagrant>)> host['ssh']
=> {:config=>false,
 :verify_host_key=>false,
 :auth_methods=>["publickey"],
 :port=>22,
 :forward_agent=>true,
 :keys=>["/home/username/.ssh/id_rsa"],
 :user_known_hosts_file=>"/home/username/.ssh/known_hosts",
 :keepalive=>true,
 :keepalive_interval=>1}
```

Post line 204:
```
 pry(#<Beaker::Vagrant>)> host['ssh']
=> {:config=>"/home/username/tmp/el720200316-3048142-16p0ol5"}
```

sbeaulie:
That's it, just set `host['ssh'][:config] = f.path()` and leave the rest. Can you rebuild the library on that change and test your long running tests again?

trevor-vaughan:
I can check and see if it fixes the original bug report (which would pretty much validate everything).

trevor-vaughan:
Ok, I now remember why I had to do it here.

The issue is that the config file is overridden by any and all options in the options hash.

This means that most of the default options are incorrect. I think that we tried the workaround here because it would solve the issue for people that pointed to an alternate configuration file. However, moving something like this into `beaker-vagrant` might make more sense just to fix this one case and figure that it probably won't pop up elsewhere.

trevor-vaughan:
@sbeaulie I think I just figured it out. I'll patch this when I get a chance and ping you. Should be some time later this evening.

trevor-vaughan:
Closing, this needs to be pushed down to `beaker-vagrant`. Will open another PR there.

PR #1634, opened 2020-03-18
(BKR-1423) Pick up Vagrant config for rsync

trevor-vaughan:
Ensure that rsync connections use the on-disk Vagrant SSH configuraiton
if an alternate physical SSH configuration has not been specified.

Correlates directly with puppetlabs/beaker-vagrant#28

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -547,8 +547,17 @@ module Beaker
 
       # vagrant uses temporary ssh configs in order to use dynamic keys
       # without this config option using ssh may prompt for password
-      if ssh_opts[:config] and File.exists?(ssh_opts[:config])
-        ssh_args << "-F #{ssh_opts[:config]}"
+      #
+      # We still want any user-set SSH config to win though
+      filesystem_ssh_config = nil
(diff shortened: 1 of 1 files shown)


highb:
I'm looking into the beaker failure now.

highb:
Okay, I have a PR up to fix the issue with Beaker so we can get a good test run on this. Hopefully that will be merged soon.

trevor-vaughan:
@highb Does it make sense, and did you get a chance to look at puppetlabs/beaker-vagrant#28?

highb:
It looks like @sbeaulie is already reviewing the other PR so I'll let him give more feedback there. Given the assumption that `self[:vagrant_ssh_config]` gets set by beaker-vagrant, this PR makes sense to me. 👍 

As an aside, this all seems like a bit of an over-use of hashes where perhaps real objects should have been used instead but that's a path we've already walked down with Beaker so I'm not going to fret about it. 😆

trevor-vaughan:
@highb Heh, couldn't agree more about the Objects but just keeping things going for now.

sbeaulie:
retest this please

PR #1635, opened 2020-03-24
(BKR-1637) fix for Ubuntu 20.04 ssh cmd environment

ciprianbadescu:
In Ubuntu 20.04 openssh server was updated to `OpenSSH_8.1p1 Ubuntu-5`
and the way the environment is handled was changed: if a variable
is defined in `/etc/environment` it cannot be overwritten from
`~/.ssh/envirnment`.

With this change we empty `/etc/environment` on Ubuntu 20.04 to keep
the previous behaviour.


ciprianbadescu:
@puppetlabs/night-s-watch for review

PR #1640, opened 2020-05-06
(BKR-1651) Remove el-5 support

highb:
This commit removes support for adding the EPEL repos via Beaker.
This is being done because the EPEL mirrors provided by RedHat have
changed their structure significantly and we do not have time to update
Beaker's code to match.

Previously #1639 cleaned this up for the external resource test. This removes the remaining references.


highb:
Jenkins test this please

highb:
Jenkins test this please

PR #1642, opened 2020-05-08
(hotfix) Remove single quotes from Unix host file ops

highb:
Prior to this commit single quotes were added around the `mkdir_p`
`rm_rf` and `mv` commands in `Beaker::Unix::Exec`. Refer to https://github.com/puppetlabs/beaker/commit/a60451cfa546126972f6588024552ca95c4ec9b5
These added single quotes caused interpolation of variables such as `~` for
the user HOME to stop working.

This commit removes the single quotes added to those command string in
the mentioned commit. We can consider adding double quote in the future,
although folks may be reliant on shell globbing and other features of
not having quotes so we may need to add a parameter to control the
behavior.


highb:
Jenkins, test this please

PR #1643, opened 2020-05-13
[BKR-1653] Fix reboot time check

trevor-vaughan:
* Replace the check for 'uptime' to compare the actual boot times of the
  system between what should have been a reboot.
* Use a single command to reboot the system falling back from systemd to
  reboot to shutdown -r which should cover all of the usual suspects
  * The previous 'shutdown -r' would not always actually trigger a
    reboot on systemd-enabled systems

BKR-1653 #close

### DIFF
--- a/lib/beaker/host/unix/exec.rb
+++ b/lib/beaker/host/unix/exec.rb
@@ -12,20 +12,23 @@ module Unix::Exec
   #
   # Will throw an exception RebootFailure if it fails
   def reboot(wait_time=10, max_connection_tries=9, uptime_retries=18)
-    begin
-      original_uptime = exec(Beaker::Command.new("uptime"))
-      original_uptime_str = parse_uptime original_uptime.stdout
-      original_uptime_int = uptime_int original_uptime_str
+    require 'time'
 
-      if self['platform'] =~ /solaris/
-        exec(Beaker::Command.new("reboot"), :expect_connection_failure => true)
-      else
-        exec(Beaker::Command.new("/sbin/shutdown -r now"), :expect_connection_failure => true)
-      end
+    begin
+      original_boot_time_str = exec(Beaker::Command.new('who -b'), {:max_connection_tries => max_connection_tries, :silent => true}).stdout
+      original_boot_time_line = original_boot_time_str.lines.grep(/boot/).first
+
+      raise Beaker::Host::RebootFailure, "Could not find system boot time using 'who -b': #{original_boot_time_str}" unless original_boot_time_line
+
+      original_boot_time = Time.parse(original_boot_time_line)
+
+      exec(Beaker::Command.new('/bin/systemctl reboot -i || reboot || /sbin/shutdown -r now'), :expect_connection_failure => true)
     rescue Beaker::Host::CommandFailure => e
       raise Beaker::Host::RebootFailure, "Command failed when attempting to reboot: #{e.message}"
     rescue RuntimeError => e
       raise Beaker::Host::RebootFailure, "Unexpected exception in reboot: #{e.message}"
+    rescue ArgumentError => e
+      raise Beaker::Host::RebootFailure, "Unable to parse time: #{e.message}"
     end
 
     attempts = 0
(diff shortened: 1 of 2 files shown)


lib/beaker/host/unix/exec.rb
```diff
@@ -12,16 +12,17 @@ module Unix::Exec
...
   # Will throw an exception RebootFailure if it fails
   def reboot(wait_time=10, max_connection_tries=9, uptime_retries=18)
+    require 'time'
+
     begin
-      original_uptime = exec(Beaker::Command.new("uptime"))
-      original_uptime_str = parse_uptime original_uptime.stdout
-      original_uptime_int = uptime_int original_uptime_str
+      original_boot_time_str = exec(Beaker::Command.new('who -b')).stdout
+      original_boot_time_line = original_boot_time_str.lines.grep(/system boot/).first
```
genebean, line 19: Not sure if this is relevant, but the output on macOS is different:

```bash
# my local mac
╔ ☕️  gene:~
╚ᐅ who -b
reboot   ~        May  7 08:58 

# a test CentOS box
╔ ☕️  gene:~
╚ᐅ ssh gene-demobox.ops.puppetlabs.net who -b                                                       
         system boot  Dec 12 00:05
```

genebean:
Initial test failure seems to be a Jenkins configuration issue. I have asked about it internally

trevor-vaughan:
Bah, of course it is. 

Well, I can change it to just check and see if something was output and roll with it :shrug:

lib/beaker/host/unix/exec.rb
trevor-vaughan: Relaxed the regex check to just `boot`. Hopefully, that catches everything.

highb:
It's not a Jenkins configuration issue; the specs are failing because you change the behavior of the method. exec_spec.rb needs to be updated to match this change.

trevor-vaughan:
@highb Well, this makes sense :facepalm:

highb:
@trevor-vaughan You should be able to run the specs locally with the following:
```
bundle install
bundle exec rake spec
```

trevor-vaughan:
Yeah, for some reason I just forgot to run the tests

genebean:
My bad @highb - I thought I saw an issue related to labels and Jenkins workers

trevor-vaughan:
Well, fixing the spec tests is going to take longer than the fix (as per usual).

The nest of stubs and mocks hasn't been easy to untangle and possibly was the reason that the previous version seemed to be correct (maybe). I think some of the logic and/or stubbing might be backwards but I'm not quite sure yet.

lib/beaker/host/unix/exec.rb
```diff
@@ -12,16 +12,17 @@ module Unix::Exec
   #
   # Will throw an exception RebootFailure if it fails
   def reboot(wait_time=10, max_connection_tries=9, uptime_retries=18)
+    require 'time'
+
     begin
-      original_uptime = exec(Beaker::Command.new("uptime"))
-      original_uptime_str = parse_uptime original_uptime.stdout
```
highb, line 17: I'll have to remember to yank out `parse_uptime` once this is merged, since I'm pretty sure it and `uptime_time` are both dead code now.

trevor-vaughan: I wasn't sure if nobody else was using them so I left them out there.

highb: I'm pretty sure they were only added as part of the rework we did to make reboot use `uptime`, so I *think* they should be safe to remove. https://github.com/puppetlabs/beaker/pull/1619/files#diff-adecec78dcc971d847136c54308d7766R37

lib/beaker/host/unix/exec.rb
```diff
@@ -12,16 +12,17 @@ module Unix::Exec
...
+      original_boot_time_line = original_boot_time_str.lines.grep(/boot/).first
 
-      if self['platform'] =~ /solaris/
-        exec(Beaker::Command.new("reboot"), :expect_connection_failure => true)
-      else
-        exec(Beaker::Command.new("/sbin/shutdown -r now"), :expect_connection_failure => true)
-      end
+      raise Beaker::Host::RebootFailure, "Could not find system boot time using 'who -b': #{original_boot_time_str}" unless original_boot_time_line
+
+      original_boot_time = Time.parse(original_boot_time_line)
```
sbeaulie, line 23: we should handle the case where Time.parse fails and also raise a `Beaker::Host::RebootFailure`

lib/beaker/host/unix/exec.rb
```diff
@@ -12,16 +12,17 @@ module Unix::Exec
...
+
+      exec(Beaker::Command.new('/bin/systemctl reboot -i || reboot || /sbin/shutdown -r now'), :expect_connection_failure => true)
```
sbeaulie, line 25: Out of curiousity which system didnt take `/sbin/shutdown -r now` but would accept `/bin/systemctl reboot` ?
We could also add 
```suggestion
      exec(Beaker::Command.new('/bin/systemctl reboot -i || reboot || /sbin/shutdown -r now || init 6'), :expect_connection_failure => true)
```

lib/beaker/host/unix/exec.rb
highb: That's a good idea. It looks like it should [throw an `ArgumentError`](https://ruby-doc.org/stdlib-2.4.1/libdoc/time/rdoc/Time.html#method-c-parse).

highb:
@sbeaulie [This is the `hosts.yaml`](https://gist.github.com/highb/ae3530621fa260b4b913743b935ac1b3) that I ran against to test.

highb:
@trevor-vaughan I think once you've addressed @sbeaulie 's feedback, this will be ready to be merged. Thanks for the PR!

lib/beaker/host/unix/exec.rb
trevor-vaughan: @sbeaulie EL7+.

`shutdown -r now` only works if no other user/daemon/whatever is actively logged in or otherwise blocking.

Since we're testing....whatever....there's really no way to guarantee that one of the offending activities is not occurring.

trevor-vaughan:
@highb and @sbeaulie Updates complete. Thanks for the review!

highb:
Checking out what happened to Jenkins. Hopefully, it's something silly that will do away on a re-run.

highb:
Ha yep, it's the flaky IIS module. Re-running should fix it.

highb:
Green test run! Merging now.

PR #1644, opened 2020-05-14
(FACT-2580) Added rm_rf, mkdir_p and mv for windows

sebastian-miclea:
These are required on Windows with cygwin, directories do not get created/removed without single quotes.
Initially they were added on unix, but that failed when using `~\path` because it was not expanded correctly.

### DIFF
--- a/lib/beaker/host/unix/exec.rb
+++ b/lib/beaker/host/unix/exec.rb
@@ -127,7 +127,7 @@ module Unix::Exec
   # @param [String] dir The directory structure to create on the host
   # @return [Boolean] True, if directory construction succeeded, otherwise False
   def mkdir_p dir
-    cmd = "mkdir -p #{dir}"
+    cmd = "mkdir -p \"#{dir}\""
     result = exec(Beaker::Command.new(cmd), :acceptable_exit_codes => [0, 1])
     result.exit_code == 0
   end
@@ -144,7 +144,7 @@ module Unix::Exec
   # @param [Boolean] rm Remove the destination prior to move
   def mv orig, dest, rm=true
     rm_rf dest unless !rm
-    execute("mv #{orig} #{dest}")
+    execute("mv \"#{orig}\" \"#{dest}\"")
   end
 
   # Attempt to ping the provided target hostname
--- a/spec/beaker/host/unix/exec_spec.rb
+++ b/spec/beaker/host/unix/exec_spec.rb
@@ -40,13 +40,13 @@ module Beaker
 
       it 'rm first' do
         expect( instance ).to receive(:execute).with("rm -rf #{destination}").and_return(0)
-        expect( instance ).to receive(:execute).with("mv #{origin} #{destination}").and_return(0)
+        expect( instance ).to receive(:execute).with("mv \"#{origin}\" \"#{destination}\"").and_return(0)
         expect( instance.mv(origin, destination) ).to be === 0
 
       end
 
       it 'does not rm' do
-         expect( instance ).to receive(:execute).with("mv #{origin} #{destination}").and_return(0)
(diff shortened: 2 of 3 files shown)


genebean:
jenkins retest this please

lib/beaker/host/windows/exec.rb
```diff
@@ -119,6 +119,30 @@ def selinux_enabled?()
     false
   end
 
+  # Create the provided directory structure on the host
+  # @param [String] dir The directory structure to create on the host
+  # @return [Boolean] True, if directory construction succeeded, otherwise False
+  def mkdir_p dir
+    cmd = "mkdir -p '#{dir}'"
```
highb, line 126: This is inconsistent with the Unix behavior due to us reverting the single quoting change that was made previously since it broke interpolated characters in the path such as `~`, which has widespread usage. If you want to prevent globbing, you may be able to use double quotes, instead?

lib/beaker/host/windows/exec.rb
```diff
@@ -119,6 +119,30 @@ def selinux_enabled?()
...
+    execute("rm -rf #{path}")
+  end
+
+  # Move the origin to destination. The destination is removed prior to moving.
+  # @param [String] orig The origin path
+  # @param [String] dest the destination path
+  # @param [Boolean] rm Remove the destination prior to move
+  def mv orig, dest, rm=true
+    rm_rf dest unless !rm
+    execute("mv '#{orig}' '#{dest}'")
```
highb, line 143: Same suggestion as above.

lib/beaker/host/windows/exec.rb
sebastian-miclea: Thanks for your suggestion, I tested out and works great!

lib/beaker/host/windows/exec.rb
```diff
@@ -119,6 +119,30 @@ def selinux_enabled?()
...
+  def mkdir_p dir
+    cmd = "mkdir -p \"#{dir}\""
+    result = exec(Beaker::Command.new(cmd), :acceptable_exit_codes => [0, 1])
+    result.exit_code == 0
+  end
+
+  # Recursively remove the path provided
+  # @param [String] path The path to remove
+  def rm_rf path
+    execute("rm -rf #{path}")
```
highb, line 134: Perhaps this should be double-quoted to prevent globbing, as well?

lib/beaker/host/windows/exec.rb
highb: Could you add the double quoting on Unix hosts so that this behavior is consistent between platforms? I can imagine it being pretty frustrating if the behavior varies. 😄

sebastian-miclea: Updated the unix methods so windows with cygwin will inherit them.

lib/beaker/host/windows/exec.rb
sebastian-miclea: This is a particular case, we have some tests that use `rm -rf /path/*` which does not delete anything if quoted or double quoted.

ekohl:
I think this should be marked as a breaking change. From https://travis-ci.org/github/voxpupuli/puppet-kafka/jobs/689172715

```
debian8-64-1 09:39:06$ mkdir -p "~/.ssh"
debian8-64-1 executed in 0.05 seconds
debian8-64-1 09:39:06$ chmod 0600 ~/.ssh
  chmod: cannot access ‘/root/.ssh’: No such file or directory
```

Note this is also exactly what broke in 4.22.0

A local example:
```
ekohl@wisse /tmp/example $ mkdir -p "~/example"
ekohl@wisse /tmp/example $ ls 
'~'
ekohl@wisse /tmp/example $ tree
.
└── ~
    └── example

2 directories, 0 files
```

PR #1654, opened 2020-05-29
(BKR-1656) Fixes for cat and file_exist? on PSWindows

sebastian-miclea:
These changes are needed in order to have the same output when these methods are used on different platforms. The implementation for unix is slightly different and had a different return values. 
Having the same return values for both unix and windows implementation will reduce the need to check which os you're running on when writing beaker tests.

**file_exist?** had a different output compared to the unix implementation (which returns bool). It actually returned the position of where it found the the match.
```
2.5.3 :001 > "true" =~ /true/
 => 0
2.5.3 :002 > "true" =~ /rue/
 => 1
```
**cat** had different output due to the fact that `execute` does not wrap the output in an `Response` object, but `exec` does.

### DIFF
--- a/lib/beaker/host/pswindows/file.rb
+++ b/lib/beaker/host/pswindows/file.rb
@@ -21,11 +21,11 @@ module PSWindows::File
   end
 
   def cat(path)
-    exec(powershell("type #{path}"))
+    exec(powershell("type #{path}")).stdout
   end
 
   def file_exist?(path)
-    result = exec(Beaker::Command.new("if exist #{path} echo true"), :acceptable_exit_codes => [0, 1])
-    result.stdout =~ /true/
+    result = exec(Beaker::Command.new("if exist #{path} echo true"), accept_all_exit_codes: true)
+    result.stdout.strip == 'true'
   end
 end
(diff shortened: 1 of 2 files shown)


mihaibuzgau:
Not taking into consideration the exit code of `if exists` seems that it does not impact the functionality.
Tests:
With directory that exists:
```
C:\Users\mitza>if exist C:\users echo true
true

C:\Users\mitza>echo %errorlevel%
0

```
With file that exists:
```
C:\Users\mitza>if exist C:\users\desktop.ini echo true
true

C:\Users\mitza>echo %errorlevel%
0

```
With directory or file that does not exist:
```
C:\Users\mitza>if exist C:\notexists echo true

C:\Users\mitza>echo %errorlevel%
0
```
relative path:
```
C:\Users\mitza>if exist notexists echo true

C:\Users\mitza>echo %errorlevel%
0
```

Characters that are not allowed in windows paths:
```
C:\Users\mitza>if exist <invalidwinpath echo true
< was unexpected at this time.

C:\Users\mitza>echo %errorlevel%
0
```

PR #1655, opened 2020-06-02
[BRK-1655] Better strict_host_key_checking fix

trevor-vaughan:
* Take into account that :paranoid is deprecated
* Take into account that true/false is deprecated for :verify_host_key

### DIFF
--- a/lib/beaker/ssh_connection.rb
+++ b/lib/beaker/ssh_connection.rb
@@ -73,10 +73,8 @@ module Beaker
          if ssh_opts.include?(:strict_host_key_checking) && (Net::SSH::Version::CURRENT.major > 5)
            strict_host_key_checking = ssh_opts.delete(:strict_host_key_checking)
 
-           if ssh_opts[:verify_host_key].nil?
-             ssh_opts[:verify_host_key] = strict_host_key_checking ? :always : :never
-           else
-             ssh_opts[:verify_host_key] = (ssh_opts[:verify_host_key] ? :always : :never) unless ssh_opts[:verify_host_key].is_a?(Symbol)
+           unless ssh_opts[:verify_host_key].is_a?(Symbol)
+             ssh_opts[:verify_host_key] ||= strict_host_key_checking ? :always : :never
            end
          end
 
--- a/lib/beaker/ssh_connection.rb
+++ b/lib/beaker/ssh_connection.rb
@@ -69,8 +69,15 @@ module Beaker
       begin
          @logger.debug "Attempting ssh connection to #{host}, user: #{user}, opts: #{ssh_opts}"
 
+         # Work around net-ssh 6+ incompatibilities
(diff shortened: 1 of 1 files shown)


lib/beaker/ssh_connection.rb
```diff
@@ -69,8 +69,15 @@ def connect_block host, user, ssh_opts, options
...
+         # Work around net-ssh 6+ incompatibilities
          if ssh_opts.include?(:strict_host_key_checking) && (Net::SSH::Version::CURRENT.major > 5)
-           ssh_opts[:paranoid] = ssh_opts.delete(:strict_host_key_checking)
+           strict_host_key_checking = ssh_opts.delete(:strict_host_key_checking)
+
+           if ssh_opts[:verify_host_key].nil?
+             ssh_opts[:verify_host_key] = strict_host_key_checking ? :always : :never
+           else
+             ssh_opts[:verify_host_key] = (ssh_opts[:verify_host_key] ? :always : :never) unless ssh_opts[:verify_host_key].is_a?(Symbol)
+           end
```
highb, line 80: I believe this should do the same thing in a more succinct manner?
```suggestion
           ssh_opts[:verify_host_key] ||= strict_host_key_checking ? :always : :never
```

trevor-vaughan: @highb The value needs to be preserved if it's already a `Symbol`

highb:
```suggestion
           unless ssh_opts[:verify_host_key].is_a?(Symbol)
             ssh_opts[:verify_host_key] ||= strict_host_key_checking ? :always : :never
           end
```
How about this?

trevor-vaughan: @highb Seems reasonable to me.

highb: `nil` has `#is_a?` so it should work:
```
[1] pry(main)> foo = nil
=> nil
[2] pry(main)> foo.is_a?(Symbol)
=> false
[3] pry(main)> foo = :banana
=> :banana
[4] pry(main)> foo.is_a?(Symbol)
=> true
```

highb:
If you want to get rid of that `verify_host_key: false is deprecated, use :never` message while you're modifying this, I believe either https://github.com/puppetlabs/beaker/blob/master/lib/beaker/options/presets.rb#L199 and https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host/cisco.rb#L16 will also need an update or this should also handle the case where `verify_host_key` is a `Boolean`.

highb:
This should prevent it from completely failing, though so it's fine if you don't want to deal with that at this moment, too! 😃

trevor-vaughan:
Well, I don't use the Cisco part and I haven't seen any issue outside of that so I'm good

highb:
Hm, looks like the tests failed because a platform is missing. I'll investigate tomorrow.

highb:
This has been released in 4.26.0.

PR #1656, opened 2020-06-03
[BKR-1653] Fix boot time race condition

trevor-vaughan:
In some cases, the initial check for the boot time appears to happen
before the system can correctly respond.

This adds the timeout check to the intial boot time check.


trevor-vaughan:
Yep, no issues here.

PR #1659, opened 2020-06-11
(maint) Improve reboot code

sbeaulie:
Before this change, the 'who -b' command was resilient when run before the reboot
Now making it retry too if it cannot find the /boot/ line after a reboot too.
We have seen one instance of a sles-12 machine rebooting, and the first request
for 'who -b' after the reboot failed the connection, which made the search for the line
be nil, and returning a stack trace when trying to convert that into Time with a
TypeError: no implicit conversion of nil into String. Now protects against that
and will retry as expected.

### DIFF
--- a/lib/beaker/host/unix/exec.rb
+++ b/lib/beaker/host/unix/exec.rb
@@ -50,7 +50,11 @@ module Unix::Exec
       sleep wait_time
 
       current_boot_time_str = exec(Beaker::Command.new('who -b'), {:max_connection_tries => max_connection_tries, :silent => true}).stdout
-      current_boot_time = Time.parse(current_boot_time_str.lines.grep(/boot/).first)
+      current_boot_time_line = current_boot_time_str.lines.grep(/boot/).first
+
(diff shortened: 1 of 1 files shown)


lib/beaker/host/unix/exec.rb
```diff
@@ -50,7 +50,11 @@ def reboot(wait_time=10, max_connection_tries=9, uptime_retries=18)
       sleep wait_time
 
       current_boot_time_str = exec(Beaker::Command.new('who -b'), {:max_connection_tries => max_connection_tries, :silent => true}).stdout
-      current_boot_time = Time.parse(current_boot_time_str.lines.grep(/boot/).first)
```
sbeaulie, line 53: we already do this for the pre-reboot code here https://github.com/puppetlabs/beaker/pull/1659/files#diff-adecec78dcc971d847136c54308d7766R23

highb: 👍  It should probably be abstracted into a method but this is fine for now.

PR #1660, opened 2020-06-23
(maint) Add Beaker gem release Github action

lucywyman:
This adds a new Github action workflow to release Beaker to
rubygems.org. It's triggered when a new tag is created, and uses the
`RUBYGEMS_AUTH_TOKEN` secret on this repository to push the built gem to
rubygems.org.

NOTE: This WON'T WORK and will be really annoying to see errors on before this repo has the auth token secret for VoxPupuli rubygems owner. This PR can be transfered when the repo is transfered, so it's probably best to leave this open until after the transfer is complete and Vox has added their auth token.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
+    if: github.repository == 'voxpupuli/beaker'
     steps:
       - uses: actions/checkout@v2
         with:
@@ -21,4 +22,3 @@ jobs:
           RUBYGEMS_AUTH_TOKEN: ${{ secrets.RUBYGEMS_AUTH_TOKEN }}
         with:
           args: push *.gem
-
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,24 @@
+name: Release the gem
+
+on:
+  create:
+    ref_type: tag
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: '0'
+      - name: Build gem
+        uses: scarhand/actions-ruby@master
+        with:
+          args: build *.gemspec
+      - name: Publish gem
+        uses: scarhand/actions-ruby@master
+        env:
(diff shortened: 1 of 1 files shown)


highb:
👍  Looking forward to seeing this tested out when Vox gets their secret added.

.github/workflows/release.yml
```diff
@@ -0,0 +1,24 @@
+name: Release the gem
+
+on:
+  create:
+    ref_type: tag
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
```
ekohl, line 9: Does it make sense to limit this to voxpupuli/beaker?
```suggestion
    runs-on: ubuntu-latest
    if: github.repository == 'voxpupuli/beaker'
```

lucywyman: Can't hurt!

.github/workflows/release.yml
```diff
@@ -0,0 +1,24 @@
...
+      - name: Build gem
+        uses: scarhand/actions-ruby@master
+        with:
+          args: build *.gemspec
+      - name: Publish gem
+        uses: scarhand/actions-ruby@master
+        env:
+          RUBYGEMS_AUTH_TOKEN: ${{ secrets.RUBYGEMS_AUTH_TOKEN }}
+        with:
+          args: push *.gem
```
ekohl, line 24: Given how minimal this container image is, does it make sense to just use the raw ruby container and execute the shell commands manually? The reason I'm asking is mostly security: this container could be modified to leak the auth token and it would be hard to detect this.

Untested, but perhaps the following gives more flexibility?

```suggestion
      - name: Install Ruby 2.6
        uses: ruby/setup-ruby@v1
        with:
          ruby-version: '2.6'
      - name: Build gem
        shell: gem build *.gemspec
      - name: Setup credentials
        shell: |
          mkdir -p ~/.gem
          echo -e "---\n:rubygems_api_key: ${{ secrets.RUBYGEMS_AUTH_TOKEN }}" > ~/.gem/credentials
          chmod 0600 ~/.gem/credentials
      - name: Publish gem
        shell: gem push *.gem
```

ekohl:
I created https://github.com/voxpupuli/gem-workflow-test/ to play around with this. Right now I'm very happy with how that works.

PR #1661, opened 2020-06-29
(BKR-1675) Add net-scp 3.x compatibility

lucaskanashiro:
Not sure if specs cover this update but there is no test failure after updating `net-scp` to the latest version.


ekohl:
https://github.com/voxpupuli/beaker/pull/1663 changes it to also allow the older versions. I'm leaning to preferring that version. Thoughts?

ghoneycutt:
This functionality was added in https://github.com/voxpupuli/beaker/pull/1663 Thanks @lucaskanashiro @ekohl

PR #1675, opened 2020-08-11
Fix UNIX reboot check granularity

### LINKED ISSUE #1674: The beaker reboot needs to cross minute boundaries based on the method used
Currently the reboot command for Unix uses `who -b` to detect a reboot. Unfortunately, this means that there is a race condition that may result in the system being unable to detect a successful reboot if it doesn't cross the minute boundary.

The `last -F reboot` command does not exist on all systems, but if it works, it can get second-level granularity. If not, we can fall back to `who -b` and sleep past the minute boundary before proceeding.

I'm working on a patch for this now.

trevor-vaughan:
Tries a more granular method for getting the seconds on a reboot.

Failing that, falls back to 'who -b' and sleeps for the remaining
seconds up to the next minute plus one to ensure that the reboot time
gets updated sufficiently.

Fixes #1674

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,3 +1,4 @@
 --format documentation
 --color
 --tty
+--fail-fast
--- a/lib/beaker/host/unix/exec.rb
+++ b/lib/beaker/host/unix/exec.rb
@@ -16,15 +16,37 @@ module Unix::Exec
 
     attempts = 0
 
+    # Some systems don't support 'last -F reboot' but it has second granularity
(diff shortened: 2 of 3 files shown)


trevor-vaughan:
Note: I've checked this with the `simp-beaker-helpers` rubygem as well as the SIMP `auditd` module.

alexjfisher:
In future, could we allow more time for interested parties to take a look before merging PRs?

PR #1677, opened 2020-08-25
Handle systems going back in time after reboot

### LINKED ISSUE #1674: The beaker reboot needs to cross minute boundaries based on the method used

trevor-vaughan:
Updates the UNIX reboot method to handle systems that go back in time
after reboot.

This can happen often on virtual machines that are running on a heavily
loaded system.

The error handling had to be loosened to allow the loop to handle
whatever bizarre circumstances get thrown at it by the underlying
system.

Fixes #1674

### DIFF
--- a/lib/beaker/host/unix/exec.rb
+++ b/lib/beaker/host/unix/exec.rb
@@ -26,17 +26,18 @@ module Unix::Exec
     original_boot_time_str = nil
     original_boot_time_line = nil
     begin
+      attempts += 1
       # Number of seconds to sleep before rebooting.
       reboot_sleep = 1
 
-      original_boot_time_str = exec(Beaker::Command.new(boot_time_cmd), {:max_connection_tries => max_connection_tries, :silent => true, :accept_all_exit_codes => true}).stdout
+      original_boot_time_str = exec(Beaker::Command.new(boot_time_cmd), {:max_connection_tries => max_connection_tries, :silent => true}).stdout
       original_boot_time_line = original_boot_time_str.lines.grep(/boot/).first
 
-      raise Beaker::Host::RebootFailure, "Could not find system boot time using '#{boot_time_cmd}': '#{original_boot_time_str}'" unless original_boot_time_line
+      raise Beaker::Host::RebootWarning, "Could not find system boot time using '#{boot_time_cmd}': '#{original_boot_time_str}'" unless original_boot_time_line
 
       original_boot_time_matches = original_boot_time_line.scan(boot_time_regex).last
 
-      raise Beaker::Host::RebootFailure, "Found no valid times in '#{original_boot_time_line}'" unless original_boot_time_matches
+      raise Beaker::Host::RebootWarning, "Found no valid times in '#{original_boot_time_line}'" unless original_boot_time_matches
 
       original_boot_time = Time.parse(original_boot_time_matches.first)
 
@@ -49,60 +50,65 @@ module Unix::Exec
       sleep(reboot_sleep)
 
       exec(Beaker::Command.new('/bin/systemctl reboot -i || reboot || /sbin/shutdown -r now'), :expect_connection_failure => true)
-    rescue RuntimeError => e
-      raise Beaker::Host::RebootFailure, "Unexpected exception in reboot: #{e.message}"
     rescue ArgumentError => e
       raise Beaker::Host::RebootFailure, "Unable to parse time: #{e.message}"
-    rescue => e
-      # Yes, this isn't best practice but rebooting is weird and we really want
-      # to make sure we retry if something went wrong (no matter what it was).
-      attempts += 1
-      if attempts < uptime_retries
-        @logger.warn("Could not get initial boot time. Will retry #{uptime_retries - attempts} more times.")
-        retry
-      else
-        raise
-      end
+    rescue Beaker::Host::RebootWarning => e
+      raise if attempts > uptime_retries
+      @logger.warn(e.message)
+      @logger.warn("Retrying #{uptime_retries - attempts} more times.")
+      retry
+    rescue StandardError => e
+      raise if attempts > uptime_retries
+      @logger.warn("Unexpected Exception: #{e.message}")
+      @logger.warn("Retrying #{uptime_retries - attempts} more times.")
+      @logger.warn(e.backtrace[0,3].join("\n"))
+      @logger.debug(e.backtrace.join("\n"))
+      retry
     end
 
     attempts = 0
     begin
+      attempts += 1
+
       # give the host a little time to shutdown
       @logger.debug("Waiting #{wait_time} for host to shut down.")
       sleep wait_time
 
+      # Accept all exit codes because this may fail due to the parallel nature of systemd
       current_boot_time_str = exec(Beaker::Command.new(boot_time_cmd), {:max_connection_tries => max_connection_tries, :silent => true, :accept_all_exit_codes => true}).stdout
       current_boot_time_line = current_boot_time_str.lines.grep(/boot/).first
 
-      raise Beaker::Host::RebootFailure, "Could not find system boot time using '#{boot_time_cmd}': '#{current_boot_time_str}'" unless current_boot_time_line
+      raise Beaker::Host::RebootWarning, "Could not find system boot time using '#{boot_time_cmd}': '#{current_boot_time_str}'" unless current_boot_time_line
 
       current_boot_time_matches = current_boot_time_line.scan(boot_time_regex).last
 
-      raise Beaker::Host::RebootFailure, "Found no valid times in '#{current_boot_time_line}'" unless current_boot_time_matches
+      raise Beaker::Host::RebootWarning, "Found no valid times in '#{current_boot_time_line}'" unless current_boot_time_matches
 
       current_boot_time = Time.parse(current_boot_time_matches.first)
 
       @logger.debug("Original Boot Time: #{original_boot_time}")
       @logger.debug("Current Boot Time: #{current_boot_time}")
 
-      # This actually may fail a few times before succeeding because virtual
-      # machines can skew backwards in time so we need to let the retry happen.
-      unless current_boot_time > original_boot_time
+      # If this is *exactly* the same then there is really no good way to detect a reboot
+      if current_boot_time == original_boot_time
         raise Beaker::Host::RebootFailure, "Boot time did not reset. Reboot appears to have failed."
       end
     rescue ArgumentError => e
       raise Beaker::Host::RebootFailure, "Unable to parse time: #{e.message}"
-    rescue => e
-      # Yes, this isn't best practice but rebooting is weird and we really want
-      # to make sure we retry if something went wrong (no matter what it was,
-      # including hosts going back in time).
-      attempts += 1
-      if attempts < uptime_retries
-        @logger.debug("Boot time did not reset. Will retry #{uptime_retries - attempts} more times.")
-        retry
-      else
-        raise
-      end
+    rescue Beaker::Host::RebootFailure => e
+      raise
+    rescue Beaker::Host::RebootWarning => e
+      raise if attempts > uptime_retries
+      @logger.warn(e.message)
+      @logger.warn("Retrying #{uptime_retries - attempts} more times.")
+      retry
+    rescue StandardError => e
+      raise if attempts > uptime_retries
+      @logger.warn("Unexpected Exception: #{e.message}")
+      @logger.warn("Retrying #{uptime_retries - attempts} more times.")
+      @logger.warn(e.backtrace[0,3].join("\n"))
+      @logger.debug(e.backtrace.join("\n"))
+      retry
     end
   end
 
--- a/lib/beaker/host/unix/exec.rb
+++ b/lib/beaker/host/unix/exec.rb
@@ -29,7 +29,7 @@ module Unix::Exec
       # Number of seconds to sleep before rebooting.
       reboot_sleep = 1
 
-      original_boot_time_str = exec(Beaker::Command.new(boot_time_cmd), {:max_connection_tries => max_connection_tries, :silent => true, :accept_all_exit_codes => true}).stdout.lines.first
+      original_boot_time_str = exec(Beaker::Command.new(boot_time_cmd), {:max_connection_tries => max_connection_tries, :silent => true, :accept_all_exit_codes => true}).stdout
       original_boot_time_line = original_boot_time_str.lines.grep(/boot/).first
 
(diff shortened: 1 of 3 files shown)


lib/beaker/host/unix/exec.rb
```diff
@@ -85,23 +85,24 @@ def reboot(wait_time=10, max_connection_tries=9, uptime_retries=18)
...
       @logger.debug("Current Boot Time: #{current_boot_time}")
 
+      # This actually may fail a few times before succeeding because virtual
+      # machines can skew backwards in time so we need to let the retry happen.
       unless current_boot_time > original_boot_time
         raise Beaker::Host::RebootFailure, "Boot time did not reset. Reboot appears to have failed."
       end
-    rescue Beaker::Host::RebootFailure => e
+    rescue ArgumentError => e
+      raise Beaker::Host::RebootFailure, "Unable to parse time: #{e.message}"
```
ekohl, line 94: Is this really fatal? Where does it even originate from?

ekohl (commented):
It's a bit hard to follow, but verall I think this looks good.

ekohl (approved):

lib/beaker/host/unix/exec.rb
trevor-vaughan: @ekohl This gets thrown if the regex match on the time doesn't work. If we get something that's absolute nonsense, we can't compare times.

lib/beaker/host/unix/exec.rb
```diff
@@ -85,23 +85,24 @@ def reboot(wait_time=10, max_connection_tries=9, uptime_retries=18)
...
+    rescue => e
+      # Yes, this isn't best practice but rebooting is weird and we really want
+      # to make sure we retry if something went wrong (no matter what it was,
+      # including hosts going back in time).
       attempts += 1
```
igalic, line 99: are we instructing beaker to reboot the system again, in case we can't tell it was rebooted?

igalic (commented):
🙋🏻‍♀️

lib/beaker/host/unix/exec.rb
```diff
@@ -49,20 +49,20 @@ def reboot(wait_time=10, max_connection_tries=9, uptime_retries=18)
...
       if attempts < uptime_retries
-        @logger.debug("Could not get initial boot time. Will retry #{uptime_retries - attempts} more times.")
+        @logger.warn("Could not get initial boot time. Will retry #{uptime_retries - attempts} more times.")
         retry
       else
         raise
       end
-    rescue Beaker::Host::CommandFailure => e
-      raise Beaker::Host::RebootFailure, "Command failed when attempting to reboot: #{e.message}"
```
op-ct, line 61: Do we want to stop explicitly rescuing `Beaker::Host::CommandFailure` exceptions?  The specificity of this message has been helpful when troubleshooting beaker failures.

lib/beaker/host/unix/exec.rb
```diff
@@ -49,20 +49,20 @@ def reboot(wait_time=10, max_connection_tries=9, uptime_retries=18)
...
 
       exec(Beaker::Command.new('/bin/systemctl reboot -i || reboot || /sbin/shutdown -r now'), :expect_connection_failure => true)
-    rescue Beaker::Host::RebootFailure => e
+    rescue RuntimeError => e
+      raise Beaker::Host::RebootFailure, "Unexpected exception in reboot: #{e.message}"
+    rescue ArgumentError => e
+      raise Beaker::Host::RebootFailure, "Unable to parse time: #{e.message}"
+    rescue => e
+      # Yes, this isn't best practice but rebooting is weird and we really want
+      # to make sure we retry if something went wrong (no matter what it was).
```
op-ct, line 58: Since this patch removes the explicit rescue for `Beaker::Host::RebootFailure` and this rescues anything that's left, the `Beaker::Host::RebootFailure` raised on line 34 will end up here.  Is that intentional?

(In case the `rescue RuntimeError` was meant to catch that: `Beaker::Host::RebootFailure` inherits from Ruby's `StandardError`.)

lib/beaker/host/unix/exec.rb
trevor-vaughan: @igalic  No, the system will try to reboot multiple times if there is some issue getting the original information from the system.

The second loop set will not try to reboot, it will just try to access the system until the count limit is reached.

lib/beaker/host/unix/exec.rb
trevor-vaughan: @op-ct Yeah, the error was getting caught earlier and missing the `retry`. When I ran through and hit some error cases, the resulting message was clear enough (gets passed through the `raise`) that it was easy to figure out what went wrong.

lib/beaker/host/unix/exec.rb
trevor-vaughan: @op-ct It wasn't more helpful than it was confusing in the end honestly. I won't lie, the whole thing probably needs a refactor but, right now, I've got things that are breaking and this implementation has worked across everything I've thrown at it.

op-ct: As far as I can see, the implementation would work just as well if the `Beaker::Host::CommandFailure` was still in place.  I disagree that the message was more confusing than helpful; at least you could narrow a failure down to an attempted host command.  

If there isn't a stronger reason to remove the rescue, I would prefer that it remains.

lib/beaker/host/unix/exec.rb
op-ct: @trevor-vaughan: Can you provide an example (or a way to make it fail in that manner)?  

I don't see how the earlier rescues here could manage to catch a `Beaker::Host::RebootFailure`; it doesn't inherit from either `RuntimeError` or `ArgumentError` (the lineage is [`RebootFailure < CommandFailure < StandardError`](https://github.com/voxpupuli/beaker/blob/90a837cff51b4d9a6487ebb3c961596201afd632/lib/beaker/host.rb#L20-L21)). That suggests that whatever error cases you encountered either didn't involve a `Beaker::Host::RebootFailure` or didn't occur in this block.

lib/beaker/host/unix/exec.rb
```diff
@@ -49,20 +49,20 @@ def reboot(wait_time=10, max_connection_tries=9, uptime_retries=18)
       sleep(reboot_sleep)
 
       exec(Beaker::Command.new('/bin/systemctl reboot -i || reboot || /sbin/shutdown -r now'), :expect_connection_failure => true)
-    rescue Beaker::Host::RebootFailure => e
+    rescue RuntimeError => e
+      raise Beaker::Host::RebootFailure, "Unexpected exception in reboot: #{e.message}"
```
op-ct, line 53: To summarize the previous discussions, I suggest these modifications:
```suggestion
    rescue Beaker::Host::RebootFailure, RuntimeError => e
      raise Beaker::Host::RebootFailure, "Unexpected exception in reboot: #{e.message}",
    rescue Beaker::Host::CommandFailure => e
      raise Beaker::Host::RebootFailure, "Command failed when attempting to reboot: #{e.message}"       
```

lib/beaker/host/unix/exec.rb
trevor-vaughan: @op-ct I'm 99% sure that the end rescue never gets hit during the command failure. This means that the `retry` never triggers if the SSH tries to hit the system but the system is taking a while to come back up. Honestly, I ended up hitting so many edge cases depending on system load, lag, etc..., that I basically narrowed it down to the 'safe ones' and 'everything else'.

lib/beaker/host/unix/exec.rb
trevor-vaughan: @op-ct Run the `simp-core` tests about 50 times over a week, watch them fail in random ways and have this be the only scenario that actually works (true story).

lib/beaker/host/unix/exec.rb
trevor-vaughan: @op-ct I actually *want* to retry if we trigger one of our exceptions. It turns out that things can fail for all sorts of reasons and somewhere between one and 5 tries later everything will "just work".

trevor-vaughan:
For what it's worth, I agree with the points that @op-ct is making. However, the only test case that I have that causes the 'back in time' issue is one that takes over an hour to run and the final solution that I have took WAY too much time to get in place and determine to be consistent.

I'm on board with finding a better way to detect reboots that works across platforms. In all honesty, creating something like a device file in `/dev` is probably a better solution but FreeBSD turned out to be a blocker on that one.

This truly should be an 'edge-case' occurrence and I'd like to come back to this later with a better solution if possible. Right now though, I just don't have a ton of time to keep prodding at individual exception stacks in the hopes that I manage to hit the appropriate (completely random) edge case again.

If anyone wants to watch their system melt, you can grab simp/simp-core and run the following (I suggest using RVM):

- Change the Gemfile to point `beaker` at this PR
- bundle install
- bundle exec rake beaker:suites
- watch your computer melt for 1+ hours

If it fails, debug, if it doesn't fail, run about 5 more times to be reasonably sure :-/

lib/beaker/host/unix/exec.rb
op-ct: @trevor-vaughan I could accept that in the abstract, but in this particular instance there's still a problem: it handles nonsensical times in two opposite ways.  I'm comparing:

1. Your reason for using `rescue ArgumentError => e` to avoid the `retry` (given to @ekohl earlier in the review):

   > "This gets thrown **if the regex match on the time doesn't work**. If we get something that's absolute nonsense, we can't compare times."

2.  The reasons for this block's two `Beaker::Host::RebootFailure`s (which don't avoid the `retry`) :

    * The first one is raised *when the regex match for the boot doesn't work*:

      ```ruby
      raise Beaker::Host::RebootFailure, "Could not find system boot time using '#{boot_time_cmd}': '#{original_boot_time_str}'" unless original_boot_time_line
      ```
    * The second one is raised  *when the regex match on the time doesn't work*: 
      ```ruby
      raise Beaker::Host::RebootFailure, "Found no valid times in '#{original_boot_time_line}'" unless original_boot_time_matches
      ```

If "the time regex is nonsense" is the reason to skip the reboot in the first instance, why are we handling it the opposite way in the second instance?

trevor-vaughan: @op-ct Oh, that!

Well, in the first instance, I seemed to have a case where this could happen a couple of times but then it would start working (no idea why). In this case, I needed to retry the reboot again because I'm not sure why the command wasn't catching. Note, this happened *extremely* rarely.

I the second instance, the host has just rebooted and literally anything goes! We may have gone back in time, rebooted into the exact same time, something else...? So, in that case, we just loop until things magically work (they usually do the first time but sometimes it takes a handful) or we eventually die and the `raise` throws the original exception up to the user.

op-ct: Thanks;  I think I see what's so weird about this now:

**(TL;DR: None of this does what you think it's doing)**

The are two separate begin/rescue blocks in the `reboot` function:
1. The first block (lines 28–66) records the "original" boot time from the SUT, then reboots it.
2. The second block (lines 69–106) checks the SUT's current boot time to verify that it has rebooted. 

The weirdness: None of what you've just mentioned happens here in this first block, and any exceptions in the second block can't `retry` the reboot:

* The reboot happens on the *final* line (line 51) of the first block. 
  *  That's the only place where the `reboot` function reboots—so if you're already in the second begin/rescue block, you've missed your chance to `retry` it.  
  * More importantly: it's impossible for any exception raised _before_ line 51 to `retry` a reboot, (because it didn't execute the first one).
* So: rescuing either of the `Beaker::Host::RebootFailure` exceptions we've been discussing could never `retry` a reboot, because the reboot happens _after_ they'd be raised.
* They also wouldn't raise in the circumstances you describe:  The purpose of `Beaker::Host::RebootFailure` exceptions (in both blocks) is to retry SUT commands when they don't provide a boot time at all—they don't raise when a boot time was returned, but it's earlier than what you expected.
* The just-rebooted, too-early/same-time problems can't happen here either, because those comparisons necessarily happen _after_ the reboot (in the second block).
* The second block is post-reboot and _does_ compare boot times—but it's impossible to `retry` the reboot from there, too—because the reboot happened in the first begin/rescue block.

So, it looks like the only way this code could actually `retry` to _re_-reboot an SUT is if the `Beaker::Command.new('/bin/systemctl reboot -i ...` at the end of the first block raised an exception internally, which got caught by the open-ended `rescue => e`.  
* Good news: if such an exception was raised, it's very likely to be the `Beaker::Host::CommandFailure` that this patch funnels into `rescue => e`
* Bad news: Even if that happened, it doesn't address your Back to the Future problems, because those are only relevant in the seconds block, and that _can't_ retry the reboot.

If any of this patch helps with your backwards-in-time or same-time scenario, it's a side effect of some kind.  This is why I asked to see an example or a way to make it fail; a lot of this is so unrelated that it would be good to take a step back and validate that the problem you _think_ you have is 1. your actual problem and 2. what you are fixing with these changes.  

Meanwhile, this patch erroneously kneecaps some useful functionality, and—worse yet—allows unrelated exceptions to cause unnecessary re-reboots, potentially several times.  

It would be a mistake to merge this, as it is right now.

op-ct: @trevor-vaughan I'd be happy to look into it.  But I haven't seen a characteristic example of a backwards-in-time failure, so I'm not sure I'd recognize one even if I reproduced it.

I know simp-core runs public acceptance tests in GitLab CI—have there been any failures over there with the problems that this patch is intended to fix?  If so, there should be a log over there we can look at.

trevor-vaughan:
Finally had time to re-read things and the new update should be in line with what @op-ct was talking about in the thread.

If nothing else, it's MUCH more explicit and clear about what errors are happening and the boot time check is less strict (because we just care that it rebooted not that it didn't do something weird with the time)

op-ct (approved):
:+1: LGTM; the clarity is much appreciated

trevor-vaughan:
@igalic and @ekohl it looks like @op-ct is OK with the changes now. Any chance of moving this forward?

trevor-vaughan:
@ghoneycutt You may be interested in this as well.

ekohl:
> @igalic and @ekohl it looks like @op-ct is OK with the changes now. Any chance of moving this forward?

I'm OK with it, but right now I'm swamped with other things so don't have the time to help push this forward.

PR #1678, opened 2020-08-26
Move the entire workflow to Github Actions

ekohl:
This removes Travis in favor of Github Actions. This also adds a release workflow via Actions. When a tag is created on the voxpupuli/beaker repo, a gem is pushed to Rubygems. This allows people to push tags to their forks without it attempting to push.

The gem release code has been tested via https://github.com/voxpupuli/gem-workflow-test and the only change is the exact repo name. The secret right now has not been added, but can be set outside of this PR. It replaces https://github.com/voxpupuli/beaker/pull/1660.


lucywyman (approved):

ghoneycutt (changes requested):
I do not think we should move individual projects to a different CI system without conversation on the mailing list. This seems like tool sprawl which makes it harder for the community to contribute. GitHub Actions might be the way to go and if it is, we should update modulesync_config and work on porting all of VP's projects over instead of a one off change.

ekohl:
> I do not think we should move individual projects to a different CI system without conversation on the mailing list.

That was indeed still on my agenda but rather busy at work so this slipped a bit.

> This seems like tool sprawl which makes it harder for the community to contribute. GitHub Actions might be the way to go and if it is, we should update modulesync_config and work on porting all of VP's projects over instead of a one off change.

Gems don't really have a modulesync today. That means it's already very mixed. However, it is my hope to unify.

ghoneycutt:
@ekohl noticed that with gems. Perhaps we need something like modulesync_config_for_gems that is very minimal. That would make it easier to port all the gems to GitHub Actions :)

ekohl:
There is https://github.com/voxpupuli/puppet-lint_modulesync_configs but as you can see, it's a bit out of date. Ideally we'd resurrect it and generalize it to any gems. We know we have a bunch of beaker gems as well which probably all behave roughly the same. IMHO this isn't hard, but for the next few weeks I have very little time.

ekohl:
I would now like to merge this. This change was done mostly via a modulesync of https://github.com/voxpupuli/puppet-lint_modulesync_configs, with the difference that this also runs yard.

bastelfreak (approved):

PR #1685, opened 2020-12-07
(maint) Show execute results on local connection

Filipovici-Andrei:
When executing commands through the local connection, their output is only transmitted as an output,
but it's not showed in the terminal.
As a fix, we now use the logger to print the result for us.

### DIFF
--- a/lib/beaker/local_connection.rb
+++ b/lib/beaker/local_connection.rb
@@ -53,9 +53,12 @@ module Beaker
           result.stdout << std_out
           result.stderr << std_err
           result.exit_code = status.exitstatus
+          @logger.info(result.stdout)
+          @logger.info(result.stderr)
         end
       rescue => e
         result.stderr << e.inspect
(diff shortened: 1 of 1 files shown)


lib/beaker/local_connection.rb
```diff
@@ -53,9 +53,12 @@ def execute command, options = {}, stdout_callback = nil, stderr_callback = stdo
           result.stdout << std_out
           result.stderr << std_err
           result.exit_code = status.exitstatus
+          @logger.info(result.stdout)
+          @logger.info(result.stderr)
```
ekohl, line 57: Given this may be used and printed in other ways, isn't `debug` more appropriate?

Filipovici-Andrei: Well with a normal beaker run (beaker provisions a VM), the command output is printed using the default info log level.
So I think that we should keep the same behaviour when running on localhost.

PR #1690, opened 2021-01-25
(maint) update ssh_preference example

steveax:
Prior to this commit, the provided example was invalid YAML. This
commit fixes the example host file so that it is parseable.

### DIFF
--- a/docs/how_to/ssh_connection_preference.md
+++ b/docs/how_to/ssh_connection_preference.md
@@ -30,7 +30,10 @@ HOSTS:
     hypervisor: vmpooler
     platform: ubuntu-16.04-amd64
     template: ubuntu-1604-x86_64
-    ssh_preference: [:vmhostname, :hostname, :ip]
+    ssh_preference:
+    - :vmhostname
+    - :hostname
+    - :ip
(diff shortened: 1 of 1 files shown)


docs/how_to/ssh_connection_preference.md
```diff
@@ -30,7 +30,10 @@ HOSTS:
     hypervisor: vmpooler
     platform: ubuntu-16.04-amd64
     template: ubuntu-1604-x86_64
-    ssh_preference: [:vmhostname, :hostname, :ip]
```
donoghuc, line 33: Alternately you could do:
```suggestion
    ssh_preference: '[:vmhostname, :hostname, :ip]'
```
But yeah this is some strange yaml parsing behavior. https://github.com/ruby/psych/issues/228

donoghuc: hm, actually no that suggestion does not work (the value is just a string). So yeah what you have here is probably best.

ekohl (approved):
Thanks!

ekohl:
I don't really see how this can break the tests.

PR #1691, opened 2021-02-02
(BKR-1690) Fix localhost logging

Filipovici-Andrei:
When running tests on localhost we don't see the output from the commands we run,
as beaker normally does, by default, on provisioned machines. 
As a fix I added logging messages.

### DIFF
--- a/lib/beaker/local_connection.rb
+++ b/lib/beaker/local_connection.rb
@@ -53,9 +53,12 @@ module Beaker
           result.stdout << std_out
           result.stderr << std_err
           result.exit_code = status.exitstatus
+          @logger.info(result.stdout) unless result.stdout.empty?
+          @logger.info(result.stderr) unless result.stderr.empty?
         end
       rescue => e
         result.stderr << e.inspect
+        @logger.info(result.stderr)
         result.exit_code = 1
       end
 


Filipovici-Andrei:
Could someone help me with the failing tests?

lib/beaker/host/unix/exec.rb
```diff
@@ -154,7 +154,7 @@ def mkdir_p dir
   # Recursively remove the path provided
   # @param [String] path The path to remove
   def rm_rf path
-    execute("rm -rf #{path}")
+    execute("rm -rf \"#{path}\"")
```
ekohl, line 157: I recall that this can break things. For example, you can no longer use `~` for a home directory.
```console
$ ls "~/bla"
ls: cannot access '~/bla': No such file or directory
```

That means that it's an API change, not a simple bugfix.

Filipovici-Andrei: Great catch, I didn't think of it. Escaping spaces, addresses the issue you mentioned.

ekohl:
This still doesn't feel like a good idea.

> Method rm_rf fails silently on windows platforms, when deleting paths with spaces.
Yet this is changing the unix provider. That doesn't really make sense to me

There was https://github.com/voxpupuli/beaker/commit/97363c95aa03491bd02cf1ad4801a03e5c84ecc9 already and doesn't that fix your problem?

Filipovici-Andrei:
That PR fixes `rm_rf` for localhost. I try to fix it for remote machines, to which beaker connects through cygwin.

ekohl:
Still, as a user of the API you are responsible for escaping. That means your fix can also change `a\ path\ with\ spaces` will be changed into `a\\ path\\ with\\ spaces`. There isn't really a way to do this right I think.

lib/beaker/host/unix/exec.rb
```diff
@@ -154,6 +154,9 @@ def mkdir_p dir
   # Recursively remove the path provided
   # @param [String] path The path to remove
   def rm_rf path
+    unless path.start_with?('"', '\'') && path.end_with?('"', '\'')
```
Filipovici-Andrei, line 157: This should do the trick. Probably it's enough to check if the string starts with quotes.

lib/beaker/host/unix/exec.rb
```diff
@@ -154,6 +154,9 @@ def mkdir_p dir
...
+      path.gsub!(/(?<!\\) /, '\ ')
```
ghoneycutt, line 158: Regular expressions should be commented with what they are attempting to match written in plain English.

Filipovici-Andrei: Added comments for that. If they are not enough, please let me know. Sorry for the long delay.

Filipovici-Andrei:
@ghoneycutt, @ekohl is there something else that should be done on this PR?

ghoneycutt:
@Filipovici-Andrei looks like you need to rebase. I'll leave to @ekohl to merge as I'm not familiar with this code.

Filipovici-Andrei:
@ghoneycutt, @ekohl, I rebased the PR.

ekohl (commented):
I really don't like this if I'm honest. It creates a really inconsistent API. For example, `mkdir_p` doesn't have any escaping.

I think there should at least be a helper method to deal with paths.

Also, another way to deal with this is to say that the caller of `rm_rf` is responsible for escaping. Then they can use `rm_rf(Shellwords.escape(path))`. We can even expose that as `rm_rf_escaped(path)`. That leaves the current API in place but introduces a "safe" API.

Filipovici-Andrei:
Agreed @ekohl. Reverted all changes except those related to logging. We need those because we don't write anywhere the run command output, as beaker normally does when it runs on a provisioned machine.

lib/beaker/local_connection.rb
```diff
@@ -53,9 +53,12 @@ def execute command, options = {}, stdout_callback = nil, stderr_callback = stdo
           result.stdout << std_out
           result.stderr << std_err
           result.exit_code = status.exitstatus
+          @logger.info(result.stdout)
+          @logger.info(result.stderr)
```
ekohl, line 57: How do you know what's what? Is that important?

Also, what happens if either is empty? Will it log empty lines?

ekohl (commented):
Thanks for removing the escaping part.

lib/beaker/local_connection.rb
Filipovici-Andrei: It's important to print them in the right order, as a command might right stuff on stdout and stderr.
And yes, if either is empty, an empty line is printed.
I'll add a check and print to log, only if there is something to print.

Filipovici-Andrei: @ekohl , @ghoneycutt are there any more issues with this PR?

ekohl (approved):

PR #1704, opened 2021-05-25
Fix vagrant matching in 'get_ip'

### LINKED ISSUE #1703: The get_ip method should match on any vagrant-backed hypervisor
The method currently has a match on `vagrant` but this misses `vagrant_libvirt` (and anything else that is added in the future)

trevor-vaughan:
Update the match in `get_ip` to match on any hypervisor matching
`vagrant`.

Fixes #1703

### DIFF
--- a/lib/beaker/host/unix/exec.rb
+++ b/lib/beaker/host/unix/exec.rb
@@ -138,7 +138,9 @@ module Unix::Exec
     if self['platform'].include?('solaris') || self['platform'].include?('osx')
       execute("ifconfig -a inet| awk '/broadcast/ {print $2}' | cut -d/ -f1 | head -1").strip
     else
-      execute("ip a | awk '/global/{print$2}' | cut -d/ -f1 | #{self['hypervisor'] == 'vagrant' ? 'tail' : 'head'} -1").strip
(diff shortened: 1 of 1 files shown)


bastelfreak (approved):

lib/beaker/host/unix/exec.rb
```diff
@@ -138,7 +138,9 @@ def get_ip
     if self['platform'].include?('solaris') || self['platform'].include?('osx')
       execute("ifconfig -a inet| awk '/broadcast/ {print $2}' | cut -d/ -f1 | head -1").strip
     else
-      execute("ip a | awk '/global/{print$2}' | cut -d/ -f1 | #{self['hypervisor'] == 'vagrant' ? 'tail' : 'head'} -1").strip
+
+      pipe_cmd = "#{self['hypervisor']}".include?('vagrant') ? 'tail' : 'head'
```
ekohl, line 142: Why is the string interpolation needed? Is it in case it's nil? Isn't it better to write:
`pipe_cmd = self['hypervisor']&.include?('vagrant') ? 'tail' : 'head'`

PR #1707, opened 2021-07-10
Align release setup with other gems

bastelfreak:
this setup is now identical to our setup in https://github.com/voxpupuli/puppet-lint_modulesync_configs


smortex (approved):
Looks fine

PR #1708, opened 2021-07-19
Use chrony on Fedora

ekohl:
Chrony became the default in Fedora 16[1]. It's so ancient that any older release can be ignored. This becomes particularly important on Fedora 34 where ntpdate is no longer available.

[1]: https://fedoraproject.org/wiki/Features/ChronyDefaultNTP


ekohl:
Turns out I needed to rebase because 553c96c98b154c97c6fe98bff9eb03a35dfc84a4 was already merged.

PR #1710, opened 2021-07-20
Implement codecov reporting

bastelfreak:
this also updates the gemspec/spec_helper to not depend on simplecov.
Right now we need to pull in the gem on all beaker-* projects because
they load the spec_helper.rb from beaker, which loads simplecov, which
isn't an explicit dependency. This PR cleans this up and also adds
coverage reporting.


alexjfisher (commented):
LGTM

PR #1712, opened 2021-07-21
Fix Platform version string comparison for install_local_package

da-ar:
The version string from Beaker::Platform was being compared to an integer. 

This was failing on fedora OS with:

```
An error occurred in a `before(:suite)` hook.
Failure/Error: host.install_local_package(package)
ArgumentError:
  comparison of Beaker::Platform with 21 failed
  
# ./.bundle/gems/ruby/2.5.0/bundler/gems/beaker-69705201a73f/lib/beaker/host/unix/pkg.rb:530:in `>'
# ./.bundle/gems/ruby/2.5.0/bundler/gems/beaker-69705201a73f/lib/beaker/host/unix/pkg.rb:530:in `install_local_package'
# ./spec/package/support/install_pdk.rb:34:in `install_local_pdk_package'
# ./spec/package/support/install_pdk.rb:9:in `install_pdk_on'
# ./spec/spec_helper_package.rb:18:in `block (3 levels) in <top (required)>'
# ./spec/spec_helper_package.rb:17:in `each'
# ./spec/spec_helper_package.rb:17:in `block (2 levels) in <top (required)>'
```

Unit tests were mocking out `Beaker::Platform` and returning an integer. This commit updates the tests to return a `Beaker::Platform` object rather than a mock.

The change in test from `huawei` to `cisco_nexus` was so that the initialisation of the `Beaker::Platform` is successful, while still allowing the test case condition to be met.

### DIFF
--- a/lib/beaker/host/unix/pkg.rb
+++ b/lib/beaker/host/unix/pkg.rb
@@ -527,7 +527,7 @@ module Unix::Pkg
     case variant
     when /^(fedora|el|redhat|centos)$/
       command_name = 'yum'
-      command_name = 'dnf' if variant == 'fedora' && version > 21
+      command_name = 'dnf' if variant == 'fedora' && version.to_i > 21
       execute("#{command_name} --nogpgcheck localinstall -y #{onhost_package_file}")
     when /^(opensuse|sles)$/
       execute("zypper --non-interactive --no-gpg-checks in #{onhost_package_file}")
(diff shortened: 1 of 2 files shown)


bastelfreak (approved):
looks good to me! I requested a review from a view people that have more knowledge about beaker/ruby.

ekohl (approved):

PR #1716, opened 2021-07-30
Drop yard dep

bastelfreak:
We don't use yard for generating anything so we don't need to include
it.


ekohl (changes requested):
I'm not sure this is correct. There are docs in the code and we use yard style docs (instead of rdoc). Why do you want to drop this?

Also, if you do drop this, you should also clean up Rakefile.

bastelfreak:
ah right. so lets keep it for now. but we should figure out how to build/publish those docs somewhere in the future.

ekohl:
It's weird that https://www.rubydoc.info/gems/beaker doesn't render anything.

PR #1718, opened 2021-09-09
Add packages_to_install method

ekohl:
This splits the existing validate_host into a method to just determine the packages to install and one that actually installs them. This makes it possible for beaker-docker to reuse the logic.


ekohl:
I forgot this was already a PR and merged another one.

PR #1719, opened 2021-09-09
Initial EL9 support

ekohl:
Currently CentOS Stream 9 is being built. This changes things so EL8 and EL9 are treated the same. Both use systemd and chrony.


ekohl:
@trevor-vaughan mind taking a look?

bastelfreak:
@trevor-vaughan can you take a look?

GabrielNagy (approved):
This looks good from our end :+1:

PR #1722, opened 2021-12-05
Arch Linux: install net-tools and openssh

bastelfreak:
Both packages are required. net-tools provides netstat

### DIFF
--- a/lib/beaker/host_prebuilt_steps.rb
+++ b/lib/beaker/host_prebuilt_steps.rb
@@ -17,7 +17,7 @@ module Beaker
     UNIX_PACKAGES = ['curl', 'ntpdate']
     FREEBSD_PACKAGES = ['curl', 'perl5|perl']
     OPENBSD_PACKAGES = ['curl']
-    ARCHLINUX_PACKAGES = ['curl', 'ntp']
+    ARCHLINUX_PACKAGES = ['curl', 'ntp', 'net-tools', 'openssh']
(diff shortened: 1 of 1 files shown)


bastelfreak:
tested this on https://github.com/voxpupuli/puppet-nginx/pull/1492

ekohl (approved):
I'm not strictly opposed to this, but I'd like to rant about something.

Right now these are only installed using SSH so installing SSH over SSH doesn't make sense. beaker-docker is in a weird state that it mostly duplicates this, but Arch reuses it. In https://github.com/voxpupuli/beaker/pull/1718 I started an effort to make this easier to reuse. That's where I'd like to end up.

With that in mind, :+1:

lib/beaker/host_prebuilt_steps.rb
```diff
@@ -17,7 +17,7 @@ module HostPrebuiltSteps
     UNIX_PACKAGES = ['curl', 'ntpdate']
     FREEBSD_PACKAGES = ['curl', 'perl5|perl']
     OPENBSD_PACKAGES = ['curl']
-    ARCHLINUX_PACKAGES = ['curl', 'ntp']
+    ARCHLINUX_PACKAGES = ['curl', 'ntp', 'net-tools', 'openssh']
```
ekohl, line 20: Btw, normally net-tools is only installed via beaker-hostgenerator. Beaker itself doesn't need net-tools and I already explained why openssh here doesn't make a lot of sense.

PR #1725, opened 2022-01-21
Add Ruby 3.1 to CI/ build gem during CI

bastelfreak:


ekohl (commented):
It looks like this is hit by Psych 4.

bastelfreak:
so we use psych in some places:
```
$ git grep -i psych
lib/beaker/options/hosts_file_parser.rb:      # @param [String] error_message Message to print if {::Psych::SyntaxError}
lib/beaker/options/hosts_file_parser.rb:        rescue Psych::SyntaxError => e
lib/beaker/options/parser.rb:      PARSE_ERROR = Psych::SyntaxError
lib/beaker/shared/fog_credentials.rb:        rescue Psych::SyntaxError, Errno::ENOENT => e
spec/beaker/options/hosts_file_parser_spec.rb:        class MockSyntaxError < Psych::SyntaxError
spec/beaker/shared/fog_credentials_spec.rb:          expect{ get_fog_credentials( fog_file_path = '/path/that/does/not/exist/.fog' ) }.to raise_error( ArgumentError, /Psych::SyntaxError/ )
```
but it's not a dependency in our gemspec

bastelfreak:
mhm
```
Psych::DisallowedClass:
       Tried to load unspecified class: Beaker::Options::OptionsHash
```

https://github.com/voxpupuli/beaker/blob/master/lib/beaker/options/options_hash.rb at least it exists

beaker.gemspec
```diff
@@ -47,6 +47,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'open_uri_redirections', '~> 0.2.1'
   s.add_runtime_dependency 'in-parallel', '~> 0.1'
   s.add_runtime_dependency 'thor', ['>= 1.0.1', '< 2.0']
+  s.add_runtime_dependency 'psych'
```
ekohl, line 50: This is a bad idea. It will pull in the latest Psych on older Rubies as well instead of using the vendored gem inside Ruby itself.

Instead, I'd update `Gemfile` to use `gem 'rdoc', '< 6.4' if RUBY_VERSION < '3.1'`

ekohl (changes requested):

ekohl:
And 2.4 / 2.5 jobs are probably now ruined with a Psych 4 in the cache while it's incompatible.

3.1 will probably need to modify the load. `load` changed to be `safe_load` in Pysch 4 and you need to select which classes can be instantiated

PR #1726, opened 2022-01-21
Add ed25519 as runtime dependency 

bastelfreak:
Previously we had this on voxpupuli-acceptance, but it makes more sense
here. See https://github.com/voxpupuli/voxpupuli-acceptance/pull/29

### DIFF
--- a/beaker.gemspec
+++ b/beaker.gemspec
@@ -39,7 +39,7 @@ Gem::Specification.new do |s|
 
   s.add_runtime_dependency 'hocon', '~> 1.0'
   s.add_runtime_dependency 'net-ssh', '>= 5.0'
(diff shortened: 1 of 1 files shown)


beaker.gemspec
```diff
@@ -39,6 +39,7 @@ Gem::Specification.new do |s|
 
   s.add_runtime_dependency 'hocon', '~> 1.0'
   s.add_runtime_dependency 'net-ssh', '>= 5.0'
+  s.add_runtime_dependency 'ed25519' # net-ssh needs this to connect to modern systems
```
ekohl, line 42: Do we want to limit this? And I wonder about this phrasing, even though it's a tiny difference.
```suggestion
  s.add_runtime_dependency 'ed25519', '~> 1.0' # net-ssh compatibility with ed25519 keys
```

PR #1735, opened 2022-05-03
Support an extension to tmpfile

ekohl:
Sometimes tools expect a certain extension or it is just cleaner to have one. This adds an optional parameter which keeps the API compatible.

### DIFF
--- a/lib/beaker/host/aix/file.rb
+++ b/lib/beaker/host/aix/file.rb
@@ -1,8 +1,8 @@
 module Aix::File
   include Beaker::CommandFactory
 
-  def tmpfile(name = '')
-    execute("rndnum=${RANDOM} && touch /tmp/#{name}.${rndnum} && echo /tmp/#{name}.${rndnum}")
+  def tmpfile(name = '', extension = nil)
+    execute("rndnum=${RANDOM} && touch /tmp/#{name}.${rndnum}#{extension} && echo /tmp/#{name}.${rndnum}#{extension}")
   end
 
   def tmpdir(name = '')
--- a/lib/beaker/host/unix/file.rb
+++ b/lib/beaker/host/unix/file.rb
@@ -1,8 +1,8 @@
 module Unix::File
   include Beaker::CommandFactory
 
-  def tmpfile(name = '')
-    execute("mktemp -t #{name}.XXXXXX")
+  def tmpfile(name = '', extension = nil)
+    execute("mktemp -t #{name}.XXXXXX#{extension}")
   end
 
   def tmpdir(name = '')
--- a/lib/beaker/host/pswindows/file.rb
+++ b/lib/beaker/host/pswindows/file.rb
@@ -1,7 +1,13 @@
 module PSWindows::File
   include Beaker::CommandFactory
 
-  def tmpfile(_name = '')
+  def tmpfile(_name = '', extension = nil)
+    if extension
+      # TODO: I do not have access to Windows, but the internet suggests this
+      # $newname = [System.IO.Path]::ChangeExtension($filename, "#{extension}") ; MoveItem $filename $newname
+      raise NotImplementedError, 'Passing an extension is not implemented'
+    end
+
     result = exec(powershell('[System.IO.Path]::GetTempFileName()'))
     result.stdout.chomp
   end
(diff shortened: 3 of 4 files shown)


lib/beaker/host/aix/file.rb
```diff
@@ -1,8 +1,8 @@
 module Aix::File
   include Beaker::CommandFactory
 
-  def tmpfile(name = '')
-    execute("rndnum=${RANDOM} && touch /tmp/#{name}.${rndnum} && echo /tmp/#{name}.${rndnum}")
+  def tmpfile(name = '', extension = nil)
+    execute("rndnum=${RANDOM} && touch /tmp/#{name}.${rndnum} && echo /tmp/#{name}.${rndnum}#{extension}")
```
smortex, line 5: This looks bad (regarding the *uniqueness* of the generated temporary file name) but I guess AIX is *that* bad :shrug: 
```suggestion
    execute("rndnum=${RANDOM} && touch /tmp/#{name}.${rndnum}#{extension} && echo /tmp/#{name}.${rndnum}#{extension}")
```

lib/beaker/host/unix/file.rb
```diff
@@ -1,8 +1,8 @@
 module Unix::File
   include Beaker::CommandFactory
 
-  def tmpfile(name = '')
-    execute("mktemp -t #{name}.XXXXXX")
+  def tmpfile(name = '', extension = nil)
+    execute("mktemp -t #{name}.XXXXXX#{extension}")
```
smortex, line 5: This is not portable, for example the FreeBSD man page for `mktemp(1)` says:

> The template may be any  file name with some number of ‘Xs’ appended to it, for example `/tmp/temp.XXXX`.  The trailing ‘Xs’ are replaced with the current process number and/or a unique letter combination.

And indeed, it does not do the right thing on FreeBSD:

```sh-session
romain@zappy ~ % mktemp -t foo.XXXX.gz
/tmp/foo.XXXX.gz.uAo9BCUu
```

I am not sure there is something guaranteed to be portable given `mktemp` is not defined by POSIX.  The GNU/Linux and FreeBSD man pages at least contain references to function that support this:

> The `mkstemps()` and `mkostemps()` functions act the same as `mkstemp()` and `mkostemp()` respectively, except they permit a suffix to exist in the template.  The template should be of the form `/tmp/tmpXXXXXXsuffix`.  The `mkstemps()` and `mkostemps()` function are told the length of the suffix string.

smortex (commented):
There seems to be some portability issue.  If we can live without the guaranteed uniqueness of the filenames crated with mktemp, a simple workaround might be to create a temporary file as before and rename it after that?

lib/beaker/host/unix/file.rb
ekohl: Thanks for that insight. I wonder what would be best to do here. Given Windows does not implement it either, would it be OK to mention it as a potential compatibility issue in a param documentation? I mean, in the end it still gives you a unique filename.

smortex: That looks great!

smortex: Wait wait wait, why don't we use `Tempfile` from stdlib which [already has support for this](https://ruby-doc.org/stdlib-2.7.0/libdoc/tempfile/rdoc/Tempfile.html#method-c-new):

> The basename parameter is used to determine the name of the temporary file. You can either pass a String or an Array with 2 String elements. In the former form, the temporary file’s base name will begin with the given string. In the latter form, the temporary file’s base name will begin with the array’s first element, and end with the second element. For example:

```ruby
file = Tempfile.new('hello')
file.path  # => something like: "/tmp/hello2843-8392-92849382--0"
         
# Use the Array form to enforce an extension in the filename:
file = Tempfile.new(['hello', '.jpg'])
file.path  # => something like: "/tmp/hello2843-8392-92849382--0.jpg"
```

ekohl: Because the Ruby part is executed on the host, not the target. You may need to do some wacky things to get Ruby to execute on the target host and in beaker we can't assume it's present since Puppet may not have been bootstrapped.

ekohl:
@smortex would you be OK with merging this?

bastelfreak (approved):

PR #1737, opened 2022-05-12
Drop pry dependency

ekohl:
Ruby 3.1 bundles debug.gem which has similar capabilities. This reduces the dependencies that beaker always pulls in.

### DIFF
--- a/beaker.gemspec
+++ b/beaker.gemspec
@@ -32,9 +32,6 @@ Gem::Specification.new do |s|
   # Run time dependencies
   s.add_runtime_dependency 'minitest', '~> 5.4'
   s.add_runtime_dependency 'minitar', '~> 0.6'
-  s.add_runtime_dependency 'pry-byebug', '~> 3.9'
-  # pry-byebug can have issues with native readline libs so add rb-readline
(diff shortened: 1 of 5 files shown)


ekohl:
I don't see a backwards-incompatible label, but I'd say this is a major version bump.

ekohl:
(I'm not sure what to do with the readline dependency so before I start fixing things, I'd first like some discussion on whether this is a good direction)

bastelfreak:
+100 for dropping it. I do not see any practical benefit and the dependencies caused some headaches in the past.

ekohl:
I should note that what triggered me was that on Ruby 3.1 you need to pull in rdoc manually in your Gemfile, so I went the other direction and dropped pry instead.

bastelfreak (approved):

ekohl:
Rebased because I couldn't see the logs anymore. No actual change.

ekohl:
So what do people think? Is this a major change or minor? On the one hand, it does make a change in functionality but debug_errors is a feature that isn't commonly used.

bastelfreak (approved):

PR #1742, opened 2022-06-06
(maint) Remove /etc/environment file for ubuntu2204

cthorn42:
Clear out the environment file for ubuntu 2204 so ~/.ssh/environment can be used.

### DIFF
--- a/lib/beaker/host/unix/exec.rb
+++ b/lib/beaker/host/unix/exec.rb
@@ -306,7 +306,7 @@ module Unix::Exec
       directory = tmpdir()
       exec(Beaker::Command.new("echo 'PermitUserEnvironment yes' | cat - /etc/ssh/sshd_config > #{directory}/sshd_config.permit"))
       exec(Beaker::Command.new("mv #{directory}/sshd_config.permit /etc/ssh/sshd_config"))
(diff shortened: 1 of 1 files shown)


lib/beaker/host/unix/exec.rb
```diff
@@ -307,6 +307,7 @@ def ssh_permit_user_environment
       exec(Beaker::Command.new("echo 'PermitUserEnvironment yes' | cat - /etc/ssh/sshd_config > #{directory}/sshd_config.permit"))
       exec(Beaker::Command.new("mv #{directory}/sshd_config.permit /etc/ssh/sshd_config"))
       exec(Beaker::Command.new("echo '' >/etc/environment")) if self['platform'] =~ /ubuntu-20.04/
+      exec(Beaker::Command.new('sed -i "s/UsePAM yes/UsePAM no/" /etc/ssh/sshd_config')) if self['platform'] =~ /ubuntu-22.04/
```
joshcooper, line 310: I think we're running into [BKR-1637](https://tickets.puppetlabs.com/browse/BKR-1637) and that the `UsePAM` change is somehow working around the issue. I think we want to apply the same change that was made for Ubuntu 20.04 to 22.04, maybe something like:

```ruby
exec(Beaker::Command.new("echo '' >/etc/environment")) if self['platform'] =~ /ubuntu-2/
```

joshcooper (approved):

PR #1744, opened 2022-06-23
Add support for Win32-OpenSSH

joshcooper:
Newer versions of Windows ship with Win32-OpenSSH (see https://github.com/PowerShell/openssh-portable) This PR makes it possible to detect when that implementation is used (based on the binary path of the `sshd` service) and it converts backslashes to forward slashes, such as when calling `scp_to`. This is similar to the work that was done to support BitVise and powershell (PSWindows)

### DIFF
--- a/lib/beaker/host/windows.rb
+++ b/lib/beaker/host/windows.rb
@@ -38,8 +38,14 @@ module Windows
       return @ssh_server if @ssh_server
       @ssh_server = :openssh
       status = execute('cmd.exe /c sc query BvSshServer', :accept_all_exit_codes => true)
-      @ssh_server = :bitvise if status =~ /4  RUNNING/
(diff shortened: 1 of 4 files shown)


smortex (approved):
The code looks good, windows is weird :zany_face:

joshcooper:
Still a problem trying to scp to a file containing spaces: 

```
    localhost $ scp /tmp/beaker20220622-2142744-1x2p96 jumpy-firepower.xxx:C:/Program Files/Puppet Labs/Puppet/puppet/lib/ruby/site_ruby/2.7.0/facter/custom_fact.rb {:ignore => }
    Warning: Net::SCP::Error error in scp'ing. Forcing the connection to close, which should raise an erro
```

PR #1745, opened 2022-06-27
Create ~/.ssh on Windows if it doesn't exist

joshcooper:
Beaker executed `mkdir -p "~/.ssh"` on Windows hosts as part of the prebuilt
steps sequence. However, when using cygwin bash, the double quotes behave like
single quotes, disabling tilde expansion:

   $ realpath "\~/.ssh"
   /home/Administrator/\~/.ssh

So '~' was a directory containing the '.ssh' directory. This was never noticed
because all of our VM templates already have the expected ~/.ssh directory. But
if the directory doesn't exist, then later steps would fail trying to write to
~/.ssh/environment.

This commit changes beaker behavior in the case that the path starts with tilde
and does not contain a space, so it's safe to not quote the path.

See commit 54b35b7603d25f4b38d11d0173ecd489ee96cad4 which double quoted unix &
windows, which caused a regression that was fixed in 9288623e6e95e5b144d565b2206aad994d15bdcb.

### DIFF
--- a/lib/beaker/host/windows/exec.rb
+++ b/lib/beaker/host/windows/exec.rb
@@ -120,10 +120,16 @@ module Windows::Exec
   end
 
   # Create the provided directory structure on the host
-  # @param [String] dir The directory structure to create on the host
+  # @param [String,Pathname] dir The directory structure to create on the host
   # @return [Boolean] True, if directory construction succeeded, otherwise False
-  def mkdir_p dir
-    cmd = "mkdir -p \"#{dir}\""
+  def mkdir_p(dir)
+    # single or double quotes will disable ~ expansion, so only quote if we have to
+    str = dir.to_s
+    cmd = if str.start_with?('~') && !str.include?(' ')
+            "mkdir -p #{str}"
+          else
+            "mkdir -p \"#{str}\""
+          end
     result = exec(Beaker::Command.new(cmd), :acceptable_exit_codes => [0, 1])
     result.exit_code == 0
   end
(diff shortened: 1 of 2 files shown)


joshcooper:
This is an example of how it behaved before. It just happened to work because ~/.ssh already existed:

```
03:51:09 orphic-fantasy (windows2022-64-1) 10:51:09$ mkdir -p "~/.ssh"
03:51:09 orphic-fantasy (windows2022-64-1) executed in 0.19 seconds
03:51:09 orphic-fantasy (windows2022-64-1) 10:51:09$ chmod 0600 ~/.ssh
03:51:10 orphic-fantasy (windows2022-64-1) executed in 0.22 seconds
```

This is after:

```
foggy-induction 12:12:35$ mkdir -p ~/.ssh
foggy-induction executed in 0.19 seconds
foggy-induction 12:12:35$ chmod 0600 ~/.ssh
foggy-induction executed in 0.30 seconds
```

ekohl (commented):
Thoughts on using `Shellwords.escape(str).sub("\\~", "~")` instead?

joshcooper:
> Thoughts on using Shellwords.escape(str).sub("\\~", "~") instead?

I think using Shellwords would be good (as in beaker should have shell escaped things from the beginning), but that's more than I want to take on right now.

mhashizume (approved):

joshcooper:
Hi @ekohl I don't want to dismiss your input, but we've got a looming deadline, so I'm going to merge. Happy to chat in the future about improvements to beaker.

ekohl:
No worries. I thought the inline comment was the more important one. The other has a lot of potential gotchas which is why I posed it much more as an open question.

PR #1755, opened 2022-09-21
Arch Linux: Ensure keyring is up2date

bastelfreak:
Before upgrading packages on Arch Linux, we need to ensure we've the latest keyring installed.


smortex (approved):
This looks reasonable.

PR #1758, opened 2022-11-14
Add to list of permitted classes and allow aliases

nmburgan:
When trying to load hosts_preserved.yml, all of these classes need to be able to be loaded. We should probably serialize fewer things to this file, but in the meantime, this allows things to work. This also uses strings for the list, since not all the classes are loaded and available at the time we are declaring this array. This also allows aliases, since we are using those in the hosts_preserved.yml file.

### DIFF
--- a/lib/beaker/options/hosts_file_parser.rb
+++ b/lib/beaker/options/hosts_file_parser.rb
@@ -2,7 +2,24 @@ module Beaker
   module Options
     #A set of functions to parse hosts files
     module HostsFileParser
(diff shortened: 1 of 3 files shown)


ekohl (commented):
Is there some test that exercises the code that requires the serialization? As you can see, we already had tests for 3.1 but since they passed, clearly there's a gap in coverage.

nmburgan:
Good call.  I added a hosts_preserved.yml to try to load in the spec.  It's not testing the specific permitted classes or anything, just verifying it can load the file. We can get fancier with it if you think it's important.

bastelfreak (approved):

PR #1760, opened 2022-12-28
Update usage of exists? -> exist? for ruby 3.2 compatibility

tvpartytonight:


spec/beaker/cli_spec.rb
```diff
@@ -426,7 +426,7 @@ module Beaker
             cli.execute!
 
             copied_hosts_file = File.join(File.absolute_path(dir), 'hosts_preserved.yml')
-            expect( File.exists?(copied_hosts_file) ).to be_truthy
+            expect( File.exist?(copied_hosts_file) ).to be_truthy
```
smortex, line 429: Line 80 and 81 above may need updating. Maybe related to the CI failure, but maybe there are more…

Maybe we can also add Ruby 3.2 to the CI matrix in `.github/workflows/test.yml`?

ekohl:
I decided to take another route: in https://github.com/voxpupuli/beaker/pull/1761 I enabled Rubocop. Rubocop can actually automatically find these and fix these. It has some failures which I need to look into.

bastelfreak:
@tvpartytonight I think this is obsolete? If not, please reopen.

PR #1761, opened 2022-12-28
Add Rubocop

ekohl:


bastelfreak (approved):

ekohl:
It is now much bigger because I forgot to enable rubocop-{performance,rake,rspec} so they weren't active.

tvpartytonight:
This looks good to me and probably supercedes #1760 ; tests are all green for this as is, so #1760 can be closed once this PR is merged.

smortex (approved):
Not sure if one commit per offense is the best (we can't distinguish automatic changes which are generally safe from manual ones). A quick look through the changes looks good.

joshcooper:
Thanks for all of your work @ekohl!

One thing, we're seeing some CI failures as `Beaker::Host.is_x86_64?` is returning false on Windows 2016 (x64). This version of beaker returns the correct value:

```
16:01:42 Using beaker 4.38.1 from https://github.com/voxpupuli/beaker (at master@5bbe0ff)
```

This one does not:

```
12:37:20 Using beaker 4.38.1 from https://github.com/voxpupuli/beaker (at master@b173e58)
```

ekohl:
I think https://github.com/voxpupuli/beaker/pull/1763 should fix that, but I don't have a Windows setup to verify. Mind reviewing it?

lib/beaker/host/unix.rb
```diff
@@ -35,12 +35,14 @@ def determine_ssh_server
...
 
     def external_copy_base
-      return @external_copy_base if @external_copy_base
-      @external_copy_base = '/root'
-      variant, version, arch, codename = self['platform'].to_array
-      # Solaris 10 uses / as the root user directory. Solaris 11 uses /root (like most).
-      @external_copy_base = '/' if variant == 'solaris' && version == '10'
-      @external_copy_base
+      @external_copy_base ||= begin
+        if self['platform'].variant == 'solaris' && self['platform'].version == '10'
```
bastelfreak, line 39: @ekohl I think this broke beaker-puppet:
https://github.com/voxpupuli/beaker-puppet/actions/runs/4261988674/jobs/7416955740#step:4:143

ekohl: Only the test suite, which used a mock object to pretend to be a platform. Fixed in https://github.com/voxpupuli/beaker-puppet/pull/209

PR #1759, opened 2022-11-23
Release 4.39.0

bastelfreak:

PR #1765, opened 2023-02-08
(maint) StringInclude Rubocop corrections

mhashizume:
rubocop-performance recently updated to 1.16.0 and its Performance/StringInclude cop identified several instances where Beaker was matching against regex when a less-expensive `String#include?` could be used.

This commit updates those regex instances.

See also: recent GitHub Actions Rubocop failures: https://github.com/voxpupuli/beaker/actions/runs/4119323790/jobs/7112852805#logs

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -214,7 +214,7 @@ namespace :history do
     Dir.chdir( File.expand_path(File.dirname(__FILE__)) )
     output = `bundle exec ruby history.rb .`
     puts output
-    if !/success/.match?(output)
+    if !output.include?('success')
       raise "History generation failed"
     end
     Dir.chdir( original_dir )
--- a/acceptance/tests/base/host/packages.rb
+++ b/acceptance/tests/base/host/packages.rb
@@ -3,23 +3,23 @@ confine :except, :platform => %w(osx)
 
 def get_host_pkg(host)
   case
-    when /sles-10/.match?(host['platform'])
+    when host['platform'].include?('sles-10')
       Beaker::HostPrebuiltSteps::SLES10_PACKAGES
     when /opensuse|sles-/.match?(host['platform'])
       Beaker::HostPrebuiltSteps::SLES_PACKAGES
-    when /debian/.match?(host['platform'])
+    when host['platform'].include?('debian')
       Beaker::HostPrebuiltSteps::DEBIAN_PACKAGES
-    when /cumulus/.match?(host['platform'])
+    when host['platform'].include?('cumulus')
       Beaker::HostPrebuiltSteps::CUMULUS_PACKAGES
     when (host['platform'].include?('windows') and host.is_cygwin?)
       Beaker::HostPrebuiltSteps::WINDOWS_PACKAGES
     when (host['platform'].include?('windows') and not host.is_cygwin?)
       Beaker::HostPrebuiltSteps::PSWINDOWS_PACKAGES
-    when /freebsd/.match?(host['platform'])
+    when host['platform'].include?('freebsd')
       Beaker::HostPrebuiltSteps::FREEBSD_PACKAGES
-    when /openbsd/.match?(host['platform'])
+    when host['platform'].include?('openbsd')
       Beaker::HostPrebuiltSteps::OPENBSD_PACKAGES
-    when /solaris-10/.match?(host['platform'])
+    when host['platform'].include?('solaris-10')
       Beaker::HostPrebuiltSteps::SOLARIS10_PACKAGES
     else
       Beaker::HostPrebuiltSteps::UNIX_PACKAGES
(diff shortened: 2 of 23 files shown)


acceptance/tests/base/host/packages.rb
```diff
@@ -3,23 +3,23 @@
 
 def get_host_pkg(host)
   case
-    when /sles-10/.match?(host['platform'])
+    when host['platform'].include?('sles-10')
```
ekohl, line 6: I wonder if this whole case statement should be rewritten to a `case host['platform']` and then use matchers in `when`.

lib/beaker/dsl/helpers/web_helpers.rb
```diff
@@ -92,7 +92,7 @@ def fetch_http_file(base_url, file_name, dst_dir)
         # @!visibility private
         def fetch_http_dir(url, dst_dir)
           logger.notify "fetch_http_dir (url: #{url}, dst_dir #{dst_dir})"
-          if !/\//.match?(url[-1, 1])
+          if !url[-1, 1].include?('/')
```
ekohl, line 95: Isn't this essentially:
```suggestion
          if !url.end_with('/')
```

lib/beaker/host/unix/pkg.rb
```diff
@@ -27,7 +27,7 @@ def check_for_package(name, opts = {})
     case self['platform']
       when /sles-10/
         result = execute("zypper se -i --match-exact #{name}", opts) { |result| result }
-        /No packages found/.match?(result.stdout) ? (return false) : (return result.exit_code == 0)
+        result.stdout.include?('No packages found') ? (return false) : (return result.exit_code == 0)
```
ekohl, line 30: Perhaps not for this PR, but this can be simplified to:
```suggestion
        return !result.stdout.include?('No packages found') && result.exit_code == 0
```

lib/beaker/host/windows.rb
```diff
@@ -38,11 +38,11 @@ def determine_ssh_server
       return @ssh_server if @ssh_server
       @ssh_server = :openssh
       status = execute('cmd.exe /c sc query BvSshServer', :accept_all_exit_codes => true)
-      if /4  RUNNING/.match?(status)
+      if status.include?('4  RUNNING')
         @ssh_server = :bitvise
       else
         status = execute('cmd.exe /c sc qc sshd', :accept_all_exit_codes => true)
-        if /C:\\Windows\\System32\\OpenSSH\\sshd\.exe/.match?(status)
+        if status.include?('C:\\Windows\\System32\\OpenSSH\\sshd.exe')
```
ekohl, line 45: I think you need to unescape the `\` if you're using single quotes
```suggestion
        if status.include?('C:\Windows\System32\OpenSSH\sshd.exe')
```

Rakefile
```diff
@@ -214,7 +214,7 @@ namespace :history do
     Dir.chdir( File.expand_path(File.dirname(__FILE__)) )
     output = `bundle exec ruby history.rb .`
     puts output
-    if !/success/.match?(output)
+    if !output.include?(success)
```
joshcooper, line 217:
```suggestion
    if !output.include?(''success')
```

One gotcha though is the old form was nil-safe, so this works `/success/.match?(nil)` but `nil.include?('success')` doesn't

ekohl: So should it use the safe navigator?
```suggestion
    if !output&.include?('success')
```

lib/beaker/dsl/helpers/web_helpers.rb
mhashizume: Good call!

lib/beaker/host/windows.rb
smortex: `\` in single quoted string allows to escape a `'`:

```ruby
'Hello \'world\'!' == "Hello 'world'!" #=> true
```

A `\` to escape a non special char will output a literal `\` before that char, so it feels more natural to me to use `\\`:

```ruby
'C:\Windows'  #=> "C:\\Windows"
'C:\\Windows' #=> "C:\\Windows"
```

TL;DR: I would keep the line as it is.

cthorn42 (approved):

smortex (approved):

bastelfreak (approved):

bastelfreak:
@ekohl I thing this is good to go, I will unblock our release PR #1759

acceptance/tests/base/host/packages.rb
ekohl: So what I meant (and may be a smaller diff)
```ruby
def get_host_pkg(host)
  case host['platform']
  when /sles-10/
    Beaker::HostPrebuiltSteps::SLES10_PACKAGES
  when /opensuse|sles-/
    Beaker::HostPrebuiltSteps::SLES_PACKAGES
  when /debian/
    Beaker::HostPrebuiltSteps::DEBIAN_PACKAGES
  when /cumulus/
    Beaker::HostPrebuiltSteps::CUMULUS_PACKAGES
  when /windows/
    host.is_cygwin? ? Beaker::HostPrebuiltSteps::WINDOWS_PACKAGES : Beaker::HostPrebuiltSteps::PSWINDOWS_PACKAGES
  when /freebsd/
    Beaker::HostPrebuiltSteps::FREEBSD_PACKAGES
  when /openbsd/
    Beaker::HostPrebuiltSteps::OPENBSD_PACKAGES
  when /solaris-10/
    Beaker::HostPrebuiltSteps::SOLARIS10_PACKAGES
  else
    Beaker::HostPrebuiltSteps::UNIX_PACKAGES
  end
end
```

but I'm fine with this too

mhashizume:
> So many more things that could be written cleaner, but this is a good start. Let me know what you think about the case statement.

Looks good to me, I've updated my branch with that addition. Thanks @ekohl

ekohl (approved):

ekohl:
Do you want to squash the changes on merge?

bastelfreak (approved):

mhashizume:
> Do you want to squash the changes on merge?

Sure, sounds good

ekohl:
Thanks!

PR #1767, opened 2023-02-18
Drop Ruby 2.4/2.5/2.6 support

bastelfreak:


bastelfreak:
To not cause a huge diff, I didn't fix rubocop violations, but updated the .rubocop_todo.yml. I would like to do some cleanup in a followup PR.

.github/workflows/test.yml
```diff
@@ -15,20 +15,18 @@ jobs:
       - name: Install Ruby ${{ matrix.ruby }}
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: "3.1"
+          # run rubocop with the ruby version we lint for
+          ruby-version: "2.7"
```
ekohl, line 19: Why? Normally you set target version in RuboCop (which you did) so why does it matter?

Rakefile
```diff
@@ -343,3 +343,15 @@ begin
 rescue LoadError
   # Optional group in bundler
 end
+
+require 'rubocop/rake_task'
+RuboCop::RakeTask.new(:rubocop) do |task|
+  # These make the rubocop experience maybe slightly less terrible
+  task.options = ['-D', '-S', '-E']
```
ekohl, line 350: Are there long form options?

.rubocop.yml
```diff
@@ -23,7 +24,7 @@ Layout:
 Metrics:
   Enabled: false
 
-Style:
+Style/TrailingCommaInArguments:
```
ekohl, line 27: This leads to a massively larger TODO file. Please keep all style changes for a separate PR.

Gemfile
```diff
@@ -20,11 +20,11 @@ if ENV['BEAKER_PE_PR_REPO_URL']
   gem lib, :git => "git@github.com:#{author}/#{lib}.git", :branch => ref
 end
 
-group :rubocop do
-  gem 'rubocop', '~> 1.12.0'
-  gem 'rubocop-performance'
-  gem 'rubocop-rake'
-  gem 'rubocop-rspec'
+group :tests do
```
ekohl, line 23: I liked the separate rubocop group. We don't do it today, but we can exclude RuboCop from our tests this way.

Rakefile
```diff
@@ -343,3 +343,15 @@ begin
...
+
+  # Use Rubocop's Github Actions formatter if possible
+  if ENV['GITHUB_ACTIONS'] == 'true'
+    rubocop_spec = Gem::Specification.find_by_name('rubocop')
+    task.formatters << 'github' if Gem::Version.new(rubocop_spec.version) >= Gem::Version.new('1.2')
```
ekohl, line 355: Since you require a new enough version, there's no need to try to be compatible with old versions.
```suggestion
    task.formatters << 'github'
```

lib/beaker/dsl/structure.rb
```diff
@@ -58,7 +58,7 @@ def step step_name, &block
               else
                 logger.info("Exception raised during step execution and debug-errors option is set, entering pry. Exception was: #{e.inspect}")
                 logger.info("HINT: Use the pry 'backtrace' and 'up' commands to navigate to the test code")
-                binding.pry # rubocop:disable Lint/Debugger
```
ekohl, line 61: Please keep this disable statement. We explicitly want pry, so moving it to TODO is counter-productive.

lib/beaker/options/parser.rb
```diff
@@ -297,7 +297,7 @@ def parse_hosts_options
         hosts_file_content = begin
           bhg_cli = BeakerHostGenerator::CLI.new(host_generator_options)
           bhg_cli.execute
-        rescue BeakerHostGenerator::Exceptions::Error,
+        rescue BeakerHostGenerator::Exceptions::Error, # rubocop:disable Lint/ShadowedException
           BeakerHostGenerator::Exceptions::InvalidNodeSpecError => e
```
ekohl, line 301: Please fix the cop instead. It's complaining (rightfully so) that `InvalidNodeSpecError` is a subclass of `Error` and thus the rescue is redundant.
```suggestion
        rescue BeakerHostGenerator::Exceptions::Error => e
```

Gemfile
```diff
@@ -20,11 +20,11 @@ if ENV['BEAKER_PE_PR_REPO_URL']
...
+  gem 'rubocop', '~> 1.45.0'
+  gem 'rubocop-rspec', '~> 2.18.1'
+  gem 'rubocop-rake', '~> 0.6.0'
+  gem 'rubocop-performance', '~> 1.16.0'
```
ekohl, line 27: Previously the lines were sorted (and there's a cop that enforces it). Please keep that.

ekohl (changes requested):

Rakefile
bastelfreak: updated

Gemfile
bastelfreak: done

Rakefile
bastelfreak: done

Gemfile
bastelfreak: done

lib/beaker/options/parser.rb
bastelfreak: done

lib/beaker/dsl/structure.rb
bastelfreak: We cannot use the statement here because we disable all `Lint` rubocops in `.rubocop.yml`:

```
Lint:
  Enabled: false
```

I would like to remove that from .rubocop.yml in a followup PR and will readd the comment to lib/beaker/dsl/structure.rb.

.rubocop.yml
```diff
@@ -2,11 +2,12 @@
 inherit_from: .rubocop_todo.yml
 
 require:
-  - rubocop-performance
   - rubocop-rake
   - rubocop-rspec
+  - rubocop-performance
```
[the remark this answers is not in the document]
bastelfreak: done

.rubocop_todo.yml
```diff
@@ -1,11 +1,40 @@
...
+# SupportedStyles: Gemfile, gems.rb, gemspec
+# Include: **/*.gemspec, **/Gemfile, **/gems.rb
+Gemspec/DevelopmentDependencies:
+  Exclude:
+    - 'beaker.gemspec'
+
+# This cop supports safe autocorrection (--autocorrect).
+# Configuration parameters: Severity, Include.
+# Include: **/*.gemspec
+Gemspec/RequireMFA:
```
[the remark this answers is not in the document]
bastelfreak: done

.rubocop_todo.yml
```diff
@@ -35,11 +69,21 @@ Lint/EmptyBlock:
     - 'spec/beaker/logger_junit_spec.rb'
     - 'spec/beaker/logger_spec.rb'
 
+# This cop supports safe autocorrection (--autocorrect).
+Lint/ErbNewArguments:
```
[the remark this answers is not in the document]
bastelfreak: done

.rubocop_todo.yml
```diff
@@ -1,11 +1,40 @@
...
+# using RuboCop version 1.45.1.
 # The point is for the user to remove these configuration records
 # one by one as the offenses are removed from the code base.
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
+# This cop supports safe autocorrection (--autocorrect).
+# Configuration parameters: Severity, Include.
+# Include: **/*.gemspec
+Gemspec/DeprecatedAttributeAssignment:
```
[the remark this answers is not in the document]
bastelfreak: done

lib/beaker/dsl/structure.rb
ekohl: I don't see that, and if we did then I wouldn't expect Lint to show up as a TODO item.

Rakefile
```diff
@@ -343,3 +343,17 @@ begin
...
+    # These make the rubocop experience maybe slightly less terrible
+    task.options = ['--display-cop-names', '--display-style-guide', '--extra-details']
+
+    # Use Rubocop's Github Actions formatter if possible
+    if ENV['GITHUB_ACTIONS'] == 'true'
+      task.formatters << 'github'
+    end
+  end
+rescue LoadError
+end
```
ekohl, line 359: I think you're introducing a RuboCop violation by not having a comment (nor code) in the rescue block. I always like to write it like this:

```suggestion
begin
  require 'rubocop/rake_task'
rescue LoadError
  # RuboCop is an optional group
else
  RuboCop::RakeTask.new(:rubocop) do |task|
    # These make the rubocop experience maybe slightly less terrible
    task.options = ['--display-cop-names', '--display-style-guide', '--extra-details']

    # Use Rubocop's Github Actions formatter if possible
    if ENV['GITHUB_ACTIONS'] == 'true'
      task.formatters << 'github'
    end
  end
end
```

IMHO the benefit of moving the actual code to `else` is that you better know where the exception can be triggered.

lib/beaker/dsl/structure.rb
bastelfreak: fixed now as well

.rubocop_todo.yml
```diff
@@ -55,6 +75,11 @@ Lint/ShadowingOuterLocalVariable:
     - 'lib/beaker/host_prebuilt_steps.rb'
     - 'lib/beaker/perf.rb'
 
+# Configuration parameters: AllowComments, AllowNil.
+Lint/SuppressedException:
```
ekohl, line 79: I think you solved this, or are there more cases of it?

bastelfreak: done

ekohl (approved):

PR #1769, opened 2023-02-20
Drop deprecated methods

ekohl:
These have been deprecated for a while and with a new major release it's time to drop them.

This may be incomplete and I didn't check for any failing tests. Opening so we don't forget.


ekohl:
I think this is now good to be merged. It doesn't remove all deprecations, but it does remove a bunch.

bastelfreak (approved):

bastelfreak:
Are you using the master branch of beaker? The last release is 4.39.0 and it doesnt contain these changes. They are not released yet.

PR #1774, opened 2023-03-02
Do not include Unix::Exec on Windows::Exec tests

ekohl:
I'm not 100% sure this is correct, but it feels wrong to have both in there.


smortex (approved):
Agreed, this feel wrong, and CI seems to not complain to not see the methods in `Unix::Exec` that have no equivalent in `Windows::Exec` to rewrite them, so LGTM1

PR #1788, opened 2023-03-18
Drop rspec-its dependency in favor of have_attributes

ekohl:


jay7x:
JFYI, transpec has this to convert from `its` to RSpec 3 syntax: https://github.com/yujinakayama/transpec/blob/master/lib/transpec/syntax/its.rb

ekohl:
I still question if the `RSpec/ImplicitSubject` cop makes sense.

bastelfreak (approved):

PR #1792, opened 2023-03-24
rubocop: Fix more Style cops

bastelfreak:


ekohl (commented):
I'm not really used to `%w[]` and `%i[]` so I wonder if that's really better or if we should disable the cops for those. Thought I don't feel too strong about it and think it can be merged.

bastelfreak:
I also don't have a strong opinion about it. My usual rule of thumb is: Follow upstream rubocop as much as possible.

PR #1798, opened 2023-04-04
(maint) Backport 'Removes open_uri_redirections'

joshcooper:
Prior to Ruby 2.4, the OpenURI module did not natively support HTTP => HTTPS redirections, necessitating a third-party gem, open_uri_redirections.

However, redirection functionality was added in Ruby 2.4:

https://github.com/ruby/ruby/commit/393ecc9f10016828a3713438ae216dc39b965364

And the `Kernel#open` method was deprecated in Ruby 2.7 and removed entirely in Ruby 3.0 in favor of the `URI#open` method:

https://bugs.ruby-lang.org/issues/15893

This commit removes the open_uri_redirections gem from the WebHelpers module and gemspec and replaces `Kernel#open` with `URI#open`.

(cherry picked from commit 46ba6bfaf4fe39b0ef4d73648fefe09a8aa36d22)

### DIFF
--- a/beaker.gemspec
+++ b/beaker.gemspec
@@ -41,7 +41,6 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'net-ssh', '>= 5.0'
 
   s.add_runtime_dependency 'in-parallel', '~> 0.1'
-  s.add_runtime_dependency 'open_uri_redirections', '~> 0.2.1'
   s.add_runtime_dependency 'rsync', '~> 1.0.9'
   s.add_runtime_dependency 'thor', ['>= 1.0.1', '< 2.0']
 
--- a/lib/beaker/dsl/helpers/web_helpers.rb
+++ b/lib/beaker/dsl/helpers/web_helpers.rb
@@ -62,7 +62,7 @@ module Beaker
             logger.notify "Fetching: #{src}"
             logger.notify "  and saving to #{dst}"
             begin
-              URI.open(src) do |remote|
+              uri_open(src) do |remote|
                 File.open(dst, "w") do |file|
                   FileUtils.copy_stream(remote, file)
                 end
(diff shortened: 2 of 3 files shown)


joshcooper:
When using ruby 3.1.2 and beaker 4.39.0, the `Kernel#open` method doesn't behave as desired:

```
irb(main):001:0> require 'open-uri'
irb(main):002:0> require 'open_uri_redirections'
irb(main):003:0> open('http://builds.delivery.puppetlabs.net/puppet-agent/dc2700c793f83cc41c62f54210d6ca7917d2a7f2/artifacts/dc2700c793f
83cc41c62f54210d6ca7917d2a7f2.yaml', :allow_redirections => :all) { |remote| puts remote }
(irb):3:in `initialize': No such file or directory @ rb_sysopen - http://builds.delivery.puppetlabs.net/puppet-agent/dc2700c793f83cc41c62f54210d6ca7917d2a7f2/artifacts/dc2700c793f83cc41c62f54210d6ca7917d2a7f2.yaml (Errno::ENOENT)
irb(main):004:0> RUBY_VERSION
=> "3.1.2"
```

Switching to `URI.open` works and is supported in Ruby 2.4 as @mhashizume mentioned.

```
irb(main):001:0> require 'open-uri'
irb(main):002:0> URI.open('http://builds.delivery.puppetlabs.net/puppet-agent/dc2700c793f83cc41c62f54210d6ca7917d2a7f2/artifacts/dc2700c793f83cc41c62f54210d6ca7917d2a7f2.yaml') { |remote| puts remote }
#<File:0x00007f686d564af0>
irb(main):003:0> RUBY_VERSION
=> "3.1.2"
```

I know this issue is fixed in beaker 5.x, but I'd like to take a more gradual approach so that moving to beaker 5 isn't a prerequisite to releasing puppet 8. If we can release this as 4.40.0, then we'll be able to move our internal CI systems to ruby 3.1 and unblock the puppet 8 release. Then we can incrementally move to beaker 5 in a future puppet 8.x release. And I don't think there are any semver concerns because beaker isn't a runtime dependency of puppet.

Also a secondary issue is puppet 7 will be supported for roughly two years, so we may need to support beaker 4.x releases in the future, e.g. for new platform additions.

joshcooper:
codecov is confused because only one branch can be tested by each ruby version.

ekohl (approved):
Codecov is not just confused by branches, and it's not just codecov that gets confused by branches. The GH changelog generator we use also gets lost. That should hold us back though.

PR #1800, opened 2023-04-05
roles are not optional

rwaffen:
if one does not set any role, you get an error

```bash
     Failure/Error: it { is_expected.to be_installed }
     Beaker::DSL::Outcomes::FailTest:
       There should be one host with default defined!
```

One and only one is allowed to be default.
```
Failure/Error: configure_beaker
ArgumentError:
  Only one host may have the role 'default', default roles assigned to [:"ubuntu-20-04", :"ubuntu-18-04", :"ubuntu-22-04"]
```

### DIFF
--- a/docs/tutorials/creating_a_test_environment.md
+++ b/docs/tutorials/creating_a_test_environment.md
@@ -45,7 +45,7 @@ To properly define a host you must provide:
 
 * name: The string identifying this host.
 * platform: One of the Beaker supported platforms.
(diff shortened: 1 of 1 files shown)


ekohl (changes requested):
I think this isn't exactly true. AFAIK roles are mandatory on multi-host setups, but optional on single host setups.

rwaffen:
ah okay. makes sense. in my single host setup it worked until i changed it to multihost. then i try to put this somehow into a sentence. 🤔

bastelfreak (approved):

PR #1818, opened 2023-07-26
(maint) Update Minitest capitalization

mhashizume:
Minitest v5.19.0, released today, includes a commit[1] that only calls its compatibility layer when using an environment variable ("MT_COMPAT").

This means that calling "MiniTest" (with a capital "T") instead of "Minitest" (with all lowercase "T"s) no longer works.

This commit replaces all instances of "MiniTest" with "Minitest."

[1] minitest/minitest@a2c6c18

### DIFF
--- a/acceptance/lib/helpers/test_helper.rb
+++ b/acceptance/lib/helpers/test_helper.rb
@@ -59,7 +59,7 @@ end
 #      assert_equal expected, user
 #    end
 #
-# Absorbs any MiniTest::Assertion from a failing test assertion in the block.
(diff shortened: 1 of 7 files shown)


bastelfreak:
do we have to merge this into the 4.x branch? that's kinda dead and maintaining and releasing it is tricky.

mhashizume:
We're still using 4.x internally at Puppet, so a 4.x release would be much appreciated

smortex (approved):
Let's try it :)

smortex:
> Let's try it :)

Note for myself next time:

run away!

PR #1822, opened 2023-09-11
(RE-15540) Add Debian 12 support

yachub:


ekohl (commented):
We already know the code names for future release. From https://wiki.debian.org/DebianReleases we know 13 will be Trixie and 14 will be Forky. Would it make sense to add these already?

yachub:
> We already know the code names for future release. From https://wiki.debian.org/DebianReleases we know 13 will be Trixie and 14 will be Forky. Would it make sense to add these already?

Good idea, sure!

bastelfreak (approved):

PR #1829, opened 2023-11-08
(PA-5938) Add support for Debian 12 - 14 & Amazon Linux 2023 in beaker 4.x

shubhamshinde360:


lib/beaker/host_prebuilt_steps.rb
```diff
@@ -12,7 +12,8 @@ module HostPrebuiltSteps
     NTPSERVER = 'pool.ntp.org'
     SLEEPWAIT = 5
     TRIES = 5
-    RHEL8_PACKAGES = ['curl', 'chrony']
+    AMAZON2023_PACKAGES = %w[curl-minimal chrony]
+    RHEL8_PACKAGES = %w[curl chrony]
```
ekohl, line 16: Nit: would be nice to also separate this change. A bit further down there are more `[]` to `%w[]` changes that are unrelated.

ekohl (commented):
For my understanding, will Amazon Linux also be its own OS family in Facter now or is it still considered a member of the Red Hat family? I know it's now Fedora derived, so it probably still is. Just no longer really close to RHEL + derivatives.

bastelfreak:
@shubhamshinde360 are there plans to move to beaker 5? We don't really have guidelines for projects with to maintained branches and beaker is currently the only project where we do this. It would be nice if we could get rid of it in the long term.

ekohl:
Updating to Beaker 5 would be preferred, but I don't mind maintaining a stable branch for a bit. If possible, it would be great to land the patches in `master` first and then use `git cherry-pick -x` on those so the history is clear. And we don't regress.

lib/beaker/host_prebuilt_steps.rb
shubhamshinde360: Hi @ekohl,
Yes, this hierarchy suggests that amazon facts are going to be derived from the redhat family.
https://github.com/puppetlabs/facter/blob/83ef0399cece0b475a97e04e2743561d703d7ca3/lib/facter/config.rb#L20C1-L22C21

shubhamshinde360:
Hi @bastelfreak,
Our CI is using beaker 4 currently and it would be a time consuming tasks to migrate to beaker 5.
We are having discussions around the migration. But as a workaround we made this cherry-pick to test out amazon 2023 platform addition to puppet-agent.

shubhamshinde360:
Closing this PR since it is a duplicate of https://github.com/voxpupuli/beaker/pull/1828

bastelfreak:
@shubhamshinde360 this is totally fine for now, because of that the 4.x branch exists, but in the long term it would be nice if we could get rid of it. Otherwise it diverges too much from the master branch and an upgrade just gets more painful.

PR #1831, opened 2023-11-16
(maint) Use DNF as Amazon Linux 2023 package manager

mhashizume:
AWS's documentation states: "In Amazon Linux 2 and Amazon Linux, the default software package management tool is YUM. In Amazon Linux 2023, the default software package management tool is DNF."[1]

This commit updates the install_package and uninstall_package methods to leverage DNF instead of Yum for Amazon Linux 2023 specifically (not Amazon Linux or Amazon Linux 2).

[1] https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/find-install-software.html

### DIFF
--- a/lib/beaker/host/unix/pkg.rb
+++ b/lib/beaker/host/unix/pkg.rb
@@ -96,7 +96,7 @@ module Unix::Pkg
           name = "#{name}-#{version}"
         end
         execute("dnf -y #{cmdline_args} install #{name}", opts)
-      when /amazon|cisco|fedora|centos|redhat|eos|el-/
+      when /amazon-2023|cisco|fedora|centos|redhat|eos|el-/
         if version
           name = "#{name}-#{version}"
         end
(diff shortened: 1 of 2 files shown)


mhashizume:
@bastelfreak Yeah, this is kind of a weird one.

Both `install_package` and `uninstall_package` use DNF for Amazon in `master`. I hadn't realized that https://github.com/voxpupuli/beaker/commit/73edd1a1a45213ced78fdd050ba00ea9f1ee3c5e didn't apply cleanly between the branches, I assume because `4.x` and `master` have diverged a significant amount by now.

The only significant difference I'm introducing is specifying Amazon 2023 to use DNF instead of Yum since the older Amazons use Yum.

I could split this up into two commits if you'd like--one in `4.x` to correct the DNF/Yum mixup and one in `master` to be more specific about 2023 and backport the latter commit.

Happy to do whatever makes the most sense

bastelfreak:
> I could split this up into two commits if you'd like--one in 4.x to correct the DNF/Yum mixup and one in master to be more specific about 2023 and backport the latter commit.

I think that's the best approach.

mhashizume:
I've split this into separate commits and opened a PR against `master`: https://github.com/voxpupuli/beaker/pull/1832

PR #1835, opened 2023-11-22
Use DNF for Fedora, newer Enterprise Linux

mhashizume:
Prior to this commit, the install_package and uninstall_package methods used Yum for all versions of Red Hat Enterprise Linux. However, RHEL has used DNF as its default package manager since version 8.

This commit updates those methods to use DNF for RHEL 8, 9, and any future versions and to explicitly use Yum on RHEL versions 1-7.

### DIFF
--- a/lib/beaker/host/unix/pkg.rb
+++ b/lib/beaker/host/unix/pkg.rb
@@ -87,10 +87,10 @@ module Unix::Pkg
       execute("zypper --non-interactive --gpg-auto-import-keys in #{name}", opts)
     when /el-4/
       @logger.debug("Package installation not supported on rhel4")
-    when /amazon-2023|el-(8|9|1[0-9])|fedora-(2[2-9]|3[0-9])/
+    when /amazon-2023|el-(8|9|1[0-9])|fedora/
       name = "#{name}-#{version}" if version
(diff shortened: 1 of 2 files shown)


lib/beaker/host/unix/pkg.rb
```diff
@@ -87,10 +87,10 @@ def install_package(name, cmdline_args = '', version = nil, opts = {})
       execute("zypper --non-interactive --gpg-auto-import-keys in #{name}", opts)
     when /el-4/
       @logger.debug("Package installation not supported on rhel4")
-    when /amazon-2023|fedora-(2[2-9]|3[0-9])/
+    when /amazon-2023|el-(8|9|1[0-9])|fedora-(2[2-9]|3[0-9])/
```
ekohl, line 90: Fedora 40 is about to come out. Any thoughts on assuming DNF?
```ruby
    when /cisco|eos|el-[1-7]-/
      # yum
    when /amazon|el|fedora/
      # dnf
```

mhashizume: Sounds good to me. I guess at this point,we can assume any Fedora < 22 machines running are long gone.

I'll add another commit

bastelfreak (approved):

bastelfreak (approved):

PR #1836, opened 2023-11-23
voxpupuli-rubocop: Require 2.x

bastelfreak:


beaker.gemspec
```diff
@@ -22,7 +22,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'fakefs', '~> 2.4'
   s.add_development_dependency 'rake', '~> 13.0'
   s.add_development_dependency 'rspec', '~> 3.0'
-  s.add_development_dependency 'voxpupuli-rubocop', '~> 1.1'
+  s.add_development_dependency 'voxpupuli-rubocop', '~> 2.0.0'
```
evgeni, line 25: That means 2.1.0 won't be allowed. Is that intentional?

bastelfreak: Yes. We don't want to pull them in automatically because that could introduce new cops that fail. Instead we will get a PR from dependabot that bumps the version. In that PR we see if it passes or not. And dependabot adds a specific label to it and I review all PRs in our org from time to time with that label.

evgeni (approved):

PR #1853, opened 2024-03-13
Only install timesync packages if needed

ekohl:
Chrony/NTP is only needed if timesync is requested. In some cases (like containers) it never makes sense.

This was inspired by https://github.com/voxpupuli/beaker-hostgenerator/pull/359#issuecomment-1995159331 and currently testing it out.


bastelfreak (approved):

lib/beaker/host_prebuilt_steps.rb
```diff
@@ -108,6 +109,16 @@ def validate_host host, opts
...
+      packages = host_base_packages(host)
+      packages += host_timesync_packages(host) if host[:timesync]
+      packages
+    end
+
+    # Return a list of packages that should be present.
+    #
+    # @param [Host] host A host return the packages for
+    # @return [Array<String>] A list of packages to install
+    def host_base_packages(host)
```
ekohl, line 121: I am quite tempted to also simplify this, since many of them are now simply `['curl']` and there's no need to duplicate that so often.

bastelfreak (approved):

ekohl:
This is still a draft. The fail because the tests don't ensure platform is an instance of `Beaker::Platform`, unlike the real application. https://github.com/voxpupuli/beaker/pull/1855 is a start for that.

ekohl:
Test changes split off to https://github.com/voxpupuli/beaker/pull/1900.

acceptance/tests/base/host/packages.rb
```diff
@@ -34,7 +11,7 @@ def get_host_pkg(host)
 
 step '#check_for_package : can determine if a package is installed'
 hosts.each do |host|
-  package = get_host_pkg(host)[0]
+  package = 'bash'
```
ekohl, line 14: This probably isn't correct for everything, but the old approach was getting invalid.

bastelfreak: at least for everything we test on, it should be fine.

PR #1863, opened 2024-05-06
Add extension parameter to parent tmpfile method signature

ekohl:
Even though this doesn't implement it, it's still good to use the same signature consistently.

Fixes: 37dd194f11eda4e39dd8e3d5f3cde5014e7e42e5 ("Support an extension to tmpfile")


bastelfreak:
mhm not sure if that's more of a bugfix or enhancement.

smortex (approved):

PR #1867, opened 2024-05-10
Drop EoL cumulus support 

bastelfreak:
cumulus isn't a supported platform anymore, we can remove the code.


ekohl (changes requested):
You missed one:
```console
$ rg -i cumulus
lib/beaker/host_prebuilt_steps.rb
25:    CUMULUS_PACKAGES = %w[curl ntpdate]
```

bastelfreak:
fixed it

PR #1868, opened 2024-05-10
drop sys-v leftovers; assume systemctl is available for unknown platforms

bastelfreak:
This also drops support for linux systems without systemd, for example EL6. But we don't support those anymore since a long time.


lib/beaker/host/unix/exec.rb
```diff
@@ -278,10 +278,6 @@ def clear_env_var key
   # @return [Result] result of restarting the SSH service
   def ssh_service_restart
     case self['platform']
-    when /debian|ubuntu|huaweios/
```
bastelfreak, line 281: huaweios is dropped at https://github.com/voxpupuli/beaker/pull/1869

spec/beaker/host/unix/exec_spec.rb
```diff
@@ -105,40 +105,15 @@ def to_s
       end
     end
 
-    describe '#ssh_service_restart' do
-      PlatformHelpers::SYSTEMDPLATFORMS.each do |platform|
+    describe '#ssh_service_restart' do # rubocop:disable RSpec/EmptyExampleGroup
+      PlatformHelpers::SYSTEMDPLATFORMS + PlatformHelpers::DEBIANPLATFORMS.each do |platform|
```
evgeni, line 109: Shouldn't `DEBIANPLATFORMS` be merged into `SYSTEMDPLATFORMS`?

evgeni: Ah no, it's also used for other things

spec/helpers.rb
```diff
@@ -125,24 +125,10 @@ module PlatformHelpers
...
-                      'redhat',
-                      'oracle',
-                      'scientific',
-                      'eos',].concat(FEDORASYSTEMV)
+                        el-
+                        centos
+                        redhat
+                        oracle
+                        scientific
+                        eos].concat(FEDORASYSTEMD).concat(DEBIANPLATFORMS)
```
ekohl, line 133: If you're going this route, just get rid of `FEDORASYSTEMD` too. Fedora < 14 is VERY old by now. You may also get rid of `DEBIANPLATFORMS` too. In that case, the distinction really becomes Linux vs non-Linux.

evgeni (approved):

ekohl (changes requested):
I'd prefer to first drop network devices and then revisit this.

bastelfreak:
will wait with merging until EOS/cisco are dropped

joshcooper:
We still support EL6 in puppet 7, though have plans to drop it soonish.

RHEL 7 is offering extended support until June 30, 2028, so we don't have plans on dropping it.

spec/helpers.rb
evgeni: Debian is used to test e.g. apt-get commands too

ekohl: https://github.com/voxpupuli/beaker/pull/1873

lib/beaker/host/unix/exec.rb
```diff
@@ -278,20 +278,12 @@ def clear_env_var key
   # @return [Result] result of restarting the SSH service
   def ssh_service_restart
     case self['platform']
-    when /debian|ubuntu/
-      exec(Beaker::Command.new("service ssh restart"))
-    when /amazon|(el|centos|redhat|oracle|scientific)-[7-9]|fedora|archlinux-/
-      exec(Beaker::Command.new("systemctl restart sshd.service"))
-    when /el-|centos|redhat|oracle|scientific/
-      exec(Beaker::Command.new("/sbin/service sshd restart"))
```
ekohl, line 286: How about keeping this but make it `/(el|centos|redhat|oracle|scientific)-[0-6]/` so we assume systemd going forward.

bastelfreak: done

arjenz (approved):

lib/beaker/host/unix/exec.rb
```diff
@@ -278,20 +278,14 @@ def clear_env_var key
...
     case self['platform']
-    when /debian|ubuntu/
-      exec(Beaker::Command.new("service ssh restart"))
-    when /amazon|(el|centos|redhat|oracle|scientific)-[7-9]|fedora|archlinux-/
-      exec(Beaker::Command.new("systemctl restart sshd.service"))
-    when /el-|centos|redhat|oracle|scientific/
+    when /(el|centos|redhat|oracle|scientific)-[0-6]/
       exec(Beaker::Command.new("/sbin/service sshd restart"))
-    when /opensuse|sles/
-      exec(Beaker::Command.new("/usr/sbin/rcsshd restart"))
```
ekohl, line 288: I think @joshcooper wanted to say that SLES 11 still hit this code. IMHO we should treat it like EL 6: special case it and less SLES 12+ fall through to systemd.

spec/beaker/host/unix/exec_spec.rb
```diff
@@ -114,31 +112,6 @@ def to_s
...
-      PlatformHelpers::DEBIANPLATFORMS.each do |platform|
-        it "calls the correct command for #{platform}" do
-          opts['platform'] = platform
-          expect(instance).to receive(:exec)
-          expect(Beaker::Command).to receive(:new).with("service ssh restart")
-          expect { instance.ssh_service_restart }.not_to raise_error
-        end
-      end
-
-      PlatformHelpers::SYSTEMVPLATFORMS.each do |platform|
```
ekohl, line 127: Would have been nice to keep this for EL6.

lib/beaker/host/unix/exec.rb
joshcooper: Yes, I'm good with defaulting to systemd and special casing older platforms. Can you restore the ability to start services on SLES 11 using `/usr/sbin/rcsshd restart`?

spec/beaker/host/unix/exec_spec.rb
bastelfreak: @joshcooper do you still rely on it? Vox Pupuli doesn't support EL6 anymore and I suspect you won't update to beaker 6 soonish and drop EL6 from your pipelines anyways?

lib/beaker/host/unix/exec.rb
bastelfreak: https://github.com/voxpupuli/beaker/pull/1883

spec/beaker/host/unix/exec_spec.rb
ekohl: We special cased it in the code. Either we should have dropped it or implemented tests. We kept it in at @joshcooper's request so I'd like to test it

lib/beaker/host/unix/exec.rb
joshcooper: RHEL 6 will have Extended LIfecycle Support until June 30, 2024 based on https://access.redhat.com/articles/4665701#6-how-can-i-continue-to-receive-software-maintenance-bug-and-security-fixes-and-full-technical-support-for-red-hat-enterprise-linux-6-6

But they could extend it, similar to how Oracle extended Solaris 10 until Jan 2027 https://blogs.oracle.com/solaris/post/new-end-date-of-extended-support-for-oracle-solaris-10-and-113

PR #1870, opened 2024-05-10
Drop EoL EL4 support

bastelfreak:


lib/beaker/options/parser.rb
```diff
@@ -364,10 +364,10 @@ def normalize_args
 
         @validator.validate_master_count(master)
 
-        # check that windows/el-4 boxes are only agents (solaris can be a master in foss cases)
+        # check that windows boxes are only agents (solaris can be a master in foss cases)
         @options[:HOSTS].each_key do |name|
           host = @options[:HOSTS][name]
-          test_host_roles(name, host) if /windows|el-4/.match?(host[:platform])
+          test_host_roles(name, host) if host[:platform] == 'windows'
```
evgeni, line 370: I think this should be `match?`?

bastelfreak: just fixed it

evgeni: haha, someone called the :police_officer:

evgeni (approved):

evgeni (approved):

PR #1871, opened 2024-05-10
Drop support for EoL Debian/Ubuntu versions

bastelfreak:
This PR isn't complete yet. I just saw the legacy code and wanted the PR as a reminder to get rid of it.


lib/beaker/platform.rb
```diff
@@ -9,32 +9,10 @@ class Platform < String
...
-                     "wheezy" => "7",
-                     "squeeze" => "6", },
+                     "bullseye" => "11", },
         :ubuntu => { "noble" => "2404",
                      "jammy" => "2204",
-                     "focal" => "2004",
-                     "eoan" => "1910",
-                     "disco" => "1904",
-                     "cosmic" => "1810",
-                     "bionic" => "1804",
```
joshcooper, line 24: we still support bionic

lib/beaker/platform.rb
```diff
@@ -9,32 +9,10 @@ class Platform < String
       { :debian => { "forky" => "14",
                      "trixie" => "13",
                      "bookworm" => "12",
-                     "bullseye" => "11",
-                     "buster" => "10",
```
joshcooper, line 13: we still support buster

joshcooper:
The osx codenames can also be dropped like highsierra, we don't support those versions anymore and our packages no longer contain the codename (which is why newer mac codenames are not in the hash)

lib/beaker/platform.rb
bastelfreak: added it back

lib/beaker/platform.rb
bastelfreak: added it back

TheMeier (approved):

bastelfreak:
@joshcooper I don't know much about the Mac OS ecosystem. Can we drop all references to their OS names?

joshcooper:
> Can we drop all references to their OS names?

Yes, please do! We used to include the mac code name in the puppet-agent dmg, e.g. https://downloads.puppet.com/mac/10.9/PC1/x86_64/puppet-agent-1.2.4-1.mavericks.dmg. So then beaker needed to know the codename to construct the URL. That was dumb, so we took the codename out of the package.

PR #1872, opened 2024-05-10
Drop EoL cisco support

bastelfreak:
Puppet doesnt support cisco anymore since some time.


spec/beaker/host/unix_spec.rb
```diff
@@ -37,13 +37,5 @@ module Unix
         expect(host.determine_ssh_server).to be === :openssh
       end
     end
-
-    describe '#validate_setup' do
```
ekohl, line 41: Should we drop `validate_setup` altogether, since it's not used anymore? We're doing breaking changes anyway.

spec/beaker/host/unix/exec_spec.rb
```diff
@@ -142,7 +142,7 @@ def to_s
     end
 
     describe '#prepend_commands' do
-      it 'returns the pc parameter unchanged for non-cisco platforms' do
+      it 'returns the pc parameter unchanged for all platforms' do
         allow(instance).to receive(:[]).with(:platform).and_return('notcisco')
```
ekohl, line 146: Can change this to something generic? Though this is also invalid, it's more reasonable IMHO.
```suggestion
        allow(instance).to receive(:[]).with(:platform).and_return('unix')
```

ekohl (commented):
In `spec/beaker/dsl/test_tagging_spec.rb` I still see some cisco stuff. Perhaps change those examples to

bastelfreak:
for `'deals with the same tag being used on multiple platforms correctly'` I'm not quite sure how to update it:

```
    it 'deals with the same tag being used on multiple platforms correctly' do
      @confines_array = [
        {
          :platform => /^el-/,
          :tag_reason_hash => {
            'tag1' => 'reason el 1',
            'tag2' => 'reason2',
          },
        }, {
          :platform => /^cisco-/,
          :tag_reason_hash => {
            'tag1' => 'reason cisco 1',
            'tag3' => 'reason3',
          },
        },
      ]

      internal_hash = confiner.instance_variable_get(:@tag_confine_details_hash)
      expect(internal_hash.keys).to include('tag1')
      expect(internal_hash.keys).to include('tag2')
      expect(internal_hash.keys).to include('tag3')
      expect(internal_hash.keys.length).to be === 3

      shared_tag_array = internal_hash['tag1']
      expect(shared_tag_array.length).to be === 2

      platform_el_found = false
      platform_cisco_found = false
      shared_tag_array.each do |confine_details|
        case confine_details[:log_message]
        when /\ el\ 1/
          platform_el_found = true
          platform_to_match = /^el-/
          reason_to_match = /reason\ el\ 1/
        when /\ cisco\ 1/
          platform_cisco_found = true
          platform_to_match = /^cisco-/
          reason_to_match = /reason\ cisco\ 1/
        else
          log_msg = "unexpected log message for confine_details: "
          log_msg << confine_details[:log_message]
          fail(log_msg)
        end

        expect(confine_details[:platform_regex]).to eql(platform_to_match)
        expect(confine_details[:log_message]).to match(reason_to_match)
      end
      expect(platform_el_found).to be === true
      expect(platform_cisco_found).to be === true
    end
  end
```

spec/beaker/host/unix_spec.rb
bastelfreak: sounds good to me. https://github.com/voxpupuli/beaker/pull/1875

PR #1873, opened 2024-05-10
Drop Fedora < 22 support

ekohl:
Those are long EOL and this makes sure the right commands on Fedora 40 are used.


lib/beaker/host/unix/pkg.rb
```diff
@@ -193,7 +193,7 @@ def upgrade_package(name, cmdline_args = '', opts = {})
     case self['platform']
     when /opensuse|sles-/
       execute("zypper --non-interactive --no-gpg-checks up #{name}", opts)
-    when /fedora-(2[2-9]|3[0-9])/
+    when /fedora/
       execute("dnf -y #{cmdline_args} update #{name}", opts)
     when /cisco|fedora|centos|redhat|eos|el-/
```
evgeni, line 198: This Fedora will never match now ;)

PR #1874, opened 2024-05-10
Drop EoL Arista EOS support

bastelfreak:
Puppet doesn't support Arista anymore.


spec/beaker/host/unix/file_spec.rb
```diff
@@ -75,13 +68,6 @@ def to_s
         @platform = 'opensuse-15-x86_64'
         expect(instance.package_config_dir).to be === '/etc/zypp/repos.d/'
       end
-
-      it 'errors for all other platform types' do
-        @platform = 'eos-4-x86_64'
```
bastelfreak, line 80: I wasn't sure if I should update this to anything else or if we should drop it

ekohl (approved):

PR #1877, opened 2024-05-10
Rubocop: Enable Naming/VariableNumber

bastelfreak:
I would like to clean up our .rubocop.yaml to follow our standards. I think enabling this cop makes sense, except for the `determine_if_x86_64` method. rubocop prefers `determine_if_x8664` but since `x86_64` is it's own name I disable the cop for this function.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -22,9 +22,6 @@ Naming/MethodParameterName:
 Naming/PredicateName:
   Enabled: false
 
-Naming/VariableNumber:
-  Enabled: false
-
 RSpec/DescribeClass:
   Exclude:
     - acceptance/fixtures/module/spec/**/*.rb
@@ -49,3 +46,6 @@ RSpec/NestedGroups:
 
 Style:
   Enabled: false
+
+Naming/VariableNumber:
+  AllowedPatterns: ['x86_64']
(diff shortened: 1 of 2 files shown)


lib/beaker/host/mac/pkg.rb
```diff
@@ -43,7 +43,7 @@ def upgrade_package(name, _cmdline_args = '')
 
   # Examine the host system to determine the architecture
   # @return [Boolean] true if x86_64, false otherwise
-  def determine_if_x86_64
+  def determine_if_x86_64 # rubocop:disable Naming/VariableNumber
```
ekohl, line 46: https://docs.rubocop.org/rubocop/cops_naming.html#namingvariablenumber says `AllowedIdentifiers` already allows `x86_64` so I wonder why it's not working.

ekohl (changes requested):
I'd prefer to use either `AllowedIdentifiers` or `AllowedPatterns` rather than disabling every instance.

lib/beaker/host/mac/pkg.rb
bastelfreak: I tested a bit around and the default isn't working, also `AllowedIdentifiers: ['x86_64']` isn't working. I will test a newer rubocop and then raise an upstream issue.

bastelfreak: `AllowedIdentifiers` is only for whole method names. It works when we set `AllowedPatterns`

TheMeier (approved):

PR #1885, opened 2024-05-20
(PA-6032) Add ssh restart for Ubuntu

skyamgarp:
(PA-6032) Changes for Ubuntu 24

This is to resolve error:
`
far-ceiling.delivery.puppetlabs.net (ubuntu2404-64-1) executed in 1.12 seconds
Exited: 5
Failed: errored in CLI.provision
#<Beaker::Host::CommandFailure: Host 'far-ceiling.delivery.puppetlabs.net' exited with 5 running:
 systemctl restart sshd.service
Last 10 lines of output were:
	Failed to restart sshd.service: Unit sshd.service not found.>
`


lib/beaker/host/unix/exec.rb
```diff
@@ -278,6 +278,8 @@ def clear_env_var key
   # @return [Result] result of restarting the SSH service
   def ssh_service_restart
     case self['platform']
+    when /ubuntu-24.04-amd64/
```
joshcooper, line 281:
```suggestion
    when /ubuntu-24.04-/
```

skyamgarp: Added suggested change

lib/beaker/host/unix/exec.rb
```diff
@@ -278,6 +278,8 @@ def clear_env_var key
   # @return [Result] result of restarting the SSH service
   def ssh_service_restart
     case self['platform']
+    when /ubuntu-24.04-/
+      exec(Beaker::Command.new("systemctl restart ssh"))
```
bastelfreak, line 282: did they actually name the unit `ssh.service` and not `sshd.service` anymore?

skyamgarp: Yes it seems they have and sshd.service throws error.

joshcooper: Indeed `ssh.service` is the sshd server. As for why Ubuntu would rename the service, I have no idea

```
# cat  /usr/lib/systemd/system/ssh.service
[Unit]
Description=OpenBSD Secure Shell server
Documentation=man:sshd(8) man:sshd_config(5)
After=network.target auditd.service
ConditionPathExists=!/etc/ssh/sshd_not_to_be_run

[Service]
EnvironmentFile=-/etc/default/ssh
ExecStartPre=/usr/sbin/sshd -t
ExecStart=/usr/sbin/sshd -D $SSHD_OPTS
ExecReload=/usr/sbin/sshd -t
ExecReload=/bin/kill -HUP $MAINPID
KillMode=process
Restart=on-failure
RestartPreventExitStatus=255
Type=notify
RuntimeDirectory=sshd
RuntimeDirectoryMode=0755

[Install]
WantedBy=multi-user.target
Alias=sshd.service
```

bastelfreak: but if it has `Alias=sshd.service` it should be possible to run `systemctl restart sshd.service`?

joshcooper: Hmm, note this line `Alias=sshd.service`. For some reason earlier versions of ubuntu allow the service to managed via `ssh` or `sshd`. For example, Ubuntu 22.04 accepts either:

```
root@forlorn-field:~# systemctl status sshd | head -2
● ssh.service - OpenBSD Secure Shell server
     Loaded: loaded (/lib/systemd/system/ssh.service; enabled; vendor preset: enabled)
root@forlorn-field:~# systemctl status ssh | head -2
● ssh.service - OpenBSD Secure Shell server
     Loaded: loaded (/lib/systemd/system/ssh.service; enabled; vendor preset: enabled)
```

Perhaps this is a bug in our 24.04 images?

joshcooper: This appears to be a regression in the version of systemd that ships in noble https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/2057768 Since "ssh" is the correct name (whereas "sshd" is the alias) and noble is a new platform I think this PR is fine.

joshcooper (approved):
Not a blocking issue but it would be great to future proof beaker, so make "systemctl ssh" the default for all future Ubuntu, Debian, ??? versions. That way we won't need to update beaker for each new platform addition.

lib/beaker/host/unix/exec.rb
```diff
@@ -302,7 +304,7 @@ def ssh_permit_user_environment
       directory = tmpdir
       exec(Beaker::Command.new("echo 'PermitUserEnvironment yes' | cat - /etc/ssh/sshd_config > #{directory}/sshd_config.permit"))
       exec(Beaker::Command.new("mv #{directory}/sshd_config.permit /etc/ssh/sshd_config"))
-      exec(Beaker::Command.new("echo '' >/etc/environment")) if /ubuntu-2(0|2).04/.match?(self['platform'])
+      exec(Beaker::Command.new("echo '' >/etc/environment")) if /ubuntu-2(0|2|4).04/.match?(self['platform'])
```
bastelfreak, line 307: @skyamgarp does it make sense to match for all ubuntus here? (and why is that even required at all?)

skyamgarp:
> Not a blocking issue but it would be great to future proof beaker, so make "systemctl ssh" the default for all future Ubuntu, Debian, ??? versions. That way we won't need to update beaker for each new platform addition.

Earlier it was 
`    when /debian|ubuntu/
      exec(Beaker::Command.new("service ssh restart"))`
      
      Can we just go back to this instead of doing it for versions specifically?

lib/beaker/host/unix/exec.rb
skyamgarp: This is to add puppet to `~/.ssh/environment`. Without this change we get error as `puppet: command not found`. I am not sure how does this work exactly for earlier versions.

lib/beaker/host/unix/exec.rb
bastelfreak: I am a bit torn between using this for every Ubuntu+Debian vs using it only as a workaround until the systemd bug is fixed. I tend to the latter because it keeps our code shorter

lib/beaker/host/unix/exec.rb
bastelfreak: Is there a chance you can test that on other Ubuntu platforms you've to support? Then we could remove the version check and run it for every ubuntu. So we don't have to update it for ubuntu 26.04 again.

lib/beaker/host/unix/exec.rb
skyamgarp: Can we have check for just Ubuntu-24 and move forward with Ubuntu 24 support as this is blocking further development and fix this later?
@joshcooper @bastelfreak

lib/beaker/host/unix/exec.rb
```diff
@@ -278,6 +278,8 @@ def clear_env_var key
   # @return [Result] result of restarting the SSH service
   def ssh_service_restart
     case self['platform']
+    when /ubuntu-24.04-/
```
ekohl, line 281: Since all of Debian (and its derivatives) use `ssh` instead of `sshd`, I'd prefer to make this:
```suggestion
    when /debian|ubuntu/
```

Just like it was prior to 3fb9a1bd857d21995d15d6cf876ce5c64fda0bb0.

ekohl (changes requested):
Thanks for spotting this regression.

Edit: please drop the `PA-6032` from the commit. This isn't a Puppetlabs project.

spec/helpers.rb
```diff
@@ -129,5 +129,5 @@ module PlatformHelpers
                         redhat
                         oracle
                         scientific
-                        archlinux].concat(DEBIANPLATFORMS)
+                        archlinux]
```
joshcooper, line 132: Why remove `DEBIANPLATFORMS`?

lib/beaker/host/unix/exec.rb
```diff
@@ -302,7 +304,7 @@ def ssh_permit_user_environment
       directory = tmpdir
       exec(Beaker::Command.new("echo 'PermitUserEnvironment yes' | cat - /etc/ssh/sshd_config > #{directory}/sshd_config.permit"))
       exec(Beaker::Command.new("mv #{directory}/sshd_config.permit /etc/ssh/sshd_config"))
-      exec(Beaker::Command.new("echo '' >/etc/environment")) if /ubuntu-2(0|2).04/.match?(self['platform'])
+      exec(Beaker::Command.new("echo '' >/etc/environment")) if self['platform'].include?('ubuntu')
```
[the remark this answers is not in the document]
bastelfreak: I think it's fine to match on just `ubuntu`, so hopefully we won't have to touch it for Ubuntu 26.04

bastelfreak (approved):

spec/helpers.rb
```diff
@@ -122,12 +122,14 @@ def make_instance instance_data = {}
...
 
-  SYSTEMDPLATFORMS = %w[amazon-2023
-                        fedora
-                        el-
-                        centos
-                        redhat
-                        oracle
-                        scientific
-                        archlinux].concat(DEBIANPLATFORMS)
+  RHELPLATFORMS = %w[
```
ekohl, line 125: I'd say REDHATPLATFORMS is better because Fedora and Amazon 2023 aren't RHEL. Amazon 2023 (and beyond) is based on Fedora, just like RHEL is based on Fedora.

ekohl (commented):
Small naming nit, otherwise :+1:

skyamgarp:
Can we please merge this PR? @joshcooper , @bastelfreak , @ekohl

PR #1890, opened 2024-07-23
Debian: Replace apt-get with apt for updates

bastelfreak:
This replaces the old `apt-get update` with the newer `apt update`. This works on all modern Debian versions we support. Also `-qq` was added, to only output errors. This makes the beaker run a bit more readable.

### DIFF
--- a/lib/beaker/host/unix/pkg.rb
+++ b/lib/beaker/host/unix/pkg.rb
@@ -62,7 +62,8 @@ module Unix::Pkg
     return unless /debian|ubuntu/.match?(self['platform'])
     return unless @apt_needs_update
 
-    execute("apt-get update")
+    # -qq: Only output errors to stdout
+    execute("apt-get update -qq")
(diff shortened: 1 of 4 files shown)


ekohl (commented):
On my Debian 12 I see `apt-get update -qq` also works. I didn't test this, but I know with plain apt you always see `WARNING: apt does not have a stable CLI interface. Use with caution in scripts.` when you pipe things so for scripts I'd have a slight preference to stay with `apt-get` if `apt-get update -qq` also works on all stable version we support.

Another difference: `apt` tells me how many packages can be upgraded, `apt-get` doesn't.

bastelfreak:
@ekohl updated it for apt-get update -qq

ekohl (approved):
Verified this at least works on Debian 10. Debian 9 is no longer on the mirrors so that errors, out, but not because it doesn't understand `-qq`.

PR #1897, opened 2024-08-12
Arch Linux: Generate keyring before installing packages

bastelfreak:
This will ensure we've a local keyring available. That's a requirement to instal and import keys. According to
https://gitlab.archlinux.org/archlinux/archlinux-docker#principles, the container images don't ship a local keyring by default.


bastelfreak:
I don't understand how the arch linux beaker jobs passed in the past, but I also didn't have much time today to dig into the container history. I didn't see any obvious recent changes.

ekohl (approved):

bastelfreak:
tested at https://github.com/voxpupuli/puppet-example/pull/70

PR #1902, opened 2024-08-27
platform: don't install `curl-minimal` on EL family

kenyon:
`curl-minimal` was mistakenly added in commit 18e6e2292aeed5557c6a85e45ba86365ca43d6f6. Attempting to install it on EL 8 and newer causes errors because there is no such package.

`curl` is installed by default. See
https://fedoraproject.org/wiki/Changes/CurlMinimal_as_Default and #1854,
for example.

### DIFF
--- a/lib/beaker/platform.rb
+++ b/lib/beaker/platform.rb
@@ -118,7 +118,7 @@ module Beaker
     def base_packages
       case @variant
       when 'el'
-        @version.to_i >= 8 ? ['curl-minimal', 'iputils'] : %w[curl]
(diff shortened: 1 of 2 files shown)


kenyon:
Example PR where this caused test failures:

* https://github.com/voxpupuli/puppet-nginx/pull/1610
* https://github.com/voxpupuli/puppet-nginx/actions/runs/10583820842/job/29326722509?pr=1610#step:6:94

silug (approved):

kenyon:
Hit the merge button when I should have hit the rebase button :( Rebase incoming....

PR #1903, opened 2024-08-27
platform: Debian: don't install apt-transport-https

kenyon:
This package is not needed and does nothing on the versions of
Debian-family operating systems that we support (only non-EOL versions;
see #1871, for other removal of EOL Debian-family code).

### DIFF
--- a/lib/beaker/platform.rb
+++ b/lib/beaker/platform.rb
@@ -120,7 +120,7 @@ module Beaker
       when 'el'
         @version.to_i >= 8 ? ['curl-minimal', 'iputils'] : %w[curl]
       when 'debian'
-        %w[curl lsb-release apt-transport-https]
+        %w[curl lsb-release]
       when 'freebsd'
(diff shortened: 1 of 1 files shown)


ekohl (approved):

lib/beaker/platform.rb
```diff
@@ -120,7 +120,7 @@ def base_packages
       when 'el'
         @version.to_i >= 8 ? ['curl-minimal', 'iputils'] : %w[curl]
       when 'debian'
-        %w[curl lsb-release apt-transport-https]
+        %w[curl lsb-release]
```
ekohl, line 123: Unrelated, but do we still need `lsb-release`? In the past it was needed for proper facts on Debian but I think current Facter uses `/etc/os-release` to determine the values.

kenyon: Made #1904 to track this.

PR #1906, opened 2024-09-06
Require Ruby 3.2

bastelfreak:
Ruby 3.0 and older are dead upstream. Requiring 3.1 or newer allows us to safely use latest gem dependencies. Beaker isn't tied to the Puppet AIO Ruby versions, so we don't need to stay on Ruby 2.6 or 2.7.


rwaffen (approved):

bastelfreak:
Tested in https://github.com/voxpupuli/puppet-example/pull/78

TheMeier (approved):

bastelfreak:
Now that we don't use Ruby 2.7 anymore, we can switch to 3.2.

TheMeier (approved):

PR #1907, opened 2025-03-05
Changes to add powershell commands as wmic is deprecated in windows 2025

skyamgarp:
With this PR trying to use powershell commands for `group_list` and `user_list` methods for windows as wmic is deprecated with windows 2025.

### DIFF
--- a/lib/beaker/host/windows/group.rb
+++ b/lib/beaker/host/windows/group.rb
@@ -14,6 +14,20 @@ module Windows::Group
     end
   end
 
+  # using powershell commands as wmic is deprecated in windows 2025
+  def group_list_using_powershell
+    execute('cmd /c echo "" | powershell.exe "Get-LocalGroup | Select-Object -ExpandProperty Name"') do |result|
+      groups = []
+      result.stdout.each_line do |line|
+        groups << line.strip or next
+      end
+
+      yield result if block_given?
+
+      groups
+    end
+  end
+
   def group_get(name)
     execute("net localgroup \"#{name}\"") do |result|
       fail_test "failed to get group #{name}" if result.exit_code != 0
--- a/lib/beaker/host/windows/user.rb
+++ b/lib/beaker/host/windows/user.rb
@@ -14,6 +14,20 @@ module Windows::User
     end
   end
 
+  # using powershell commands as wmic is deprecated in windows 2025
+  def user_list_using_powershell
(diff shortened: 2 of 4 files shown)


lib/beaker/host/windows/group.rb
```diff
@@ -14,6 +14,20 @@ def group_list
     end
   end
 
+  # using powershell commands as wmic is deprecated in windows 2025
+  def group_list_using_powershell
+    execute('cmd /c echo "" | powershell.exe "Get-LocalGroup | Select-Object -ExpandProperty Name"') do |result|
```
joshcooper, line 19: should we be using the `powershell` dsl method to handle things like `-NoProfile`, etc?

```suggestion
    execute(powershell("Get-LocalGroup | Select-Object -ExpandProperty Name")) do |result|
```

joshcooper (approved):

lib/beaker/host/windows/group.rb
```diff
@@ -14,6 +15,17 @@ def group_list
     end
   end
 
+  # using powershell commands as wmic is deprecated in windows 2025
+  def group_list_using_powershell
+    result = exec(powershell('"Get-LocalGroup | Select-Object -ExpandProperty Name"'))
```
bastelfreak, line 20: I haven't used powershell in a decade, so maybe a stupid question: Do we need double quotes within the single quotes?

skyamgarp: yes, we need to pass this command as string instead of command hence have added quotes.

joshcooper (approved):
If we called `powershell.exe -NonInteractive ... ` then I wonder if we could drop the `echo "" | `? But it looks the `echo` pattern is used elsewhere, so at least we're consistent.

skyamgarp:
@bastelfreak , can you please review and merge the PR? Thanks.

bastelfreak:
I'm still not quite sure if this is a breaking change or not?

skyamgarp:
> I'm still not quite sure if this is a breaking change or not?

I did not understand the question, as `wmic` is deprecated we get error like  `'WMIC' is not recognized as an internal or external command`

joshcooper:
> I'm still not quite sure if this is a breaking change or not?

This PR is purely additive (because we backed out the including the DSL wrapper), so should not be breaking.

skyamgarp:
@bastelfreak could you please release this change?

PR #1915, opened 2025-04-09
Fix macOS package installation when DMG content differs from filename

gavindidrichsen:
## Problem

The current macOS package installation code assumes that

* **(1)** the `name` input to `Mac::Pkg.install_package(name,...)` will be a filename without an extension
* **(2)** the `*.pkg` file inside a DMG will have the same base name as the DMG itself.

**(1)** caused failures in our CI because the subsequent code was mismatching some of the extensions so making the code look for a non-existent file/directory.
**(2)** caused failures because some of our DMGs do not have a `*pkg` name with the same basename as the `*.dmg`

For example:

```bash
# mount the dmg before installation
utter-suzerain:~ root# hdiutil attach puppet-bolt-4.0.0-1.osx14.dmg 
Checksumming Protective Master Boot Record (MBR : 0)…
...
...
utter-suzerain:~ root#     

# notice internal *.pkg has a different basename
utter-suzerain:~ root# ls /Volumes/puppet-bolt-4.0.0-1.osx14/
puppet-bolt-4.0.0-1-installer.pkg
utter-suzerain:~ root# 
```

## Solution

This PR makes the following changes:

1. Strips any `.dmg` extension--if it exists--from the package `name` input to `Mac::Pkg.install_package(name,...)`
2. ONLYIF `#{pkg_name}` doesn't exist, then do a wildcard search for the pkg name.  If nothing is found even with this, then error out.

These changes ensure we can find and install the correct package regardless of naming differences between the DMG file and its contents.

# Testing

* Added new `spec/beaker/host/mac/pkg_spec.rb`
* Manually tested the PR condition: the pkg name is different from the dmg filename (current PR allows this)
* Manually tested backwards compatibility: (1) the pkg name is the same as dmg name; (2) the pkg doesn't exist at all within the dmg

### DIFF
--- a/lib/beaker/host/mac/pkg.rb
+++ b/lib/beaker/host/mac/pkg.rb
@@ -18,7 +18,10 @@ module Mac::Pkg
   # @param [String] pkg_base      The base name of the directory that the dmg
   #                               attaches to under `/Volumes`
   # @param [String] pkg_name      The name of the package file that should be
-  #                               used by the installer
+  #                               used by the installer. If the specified
+  #                               package is not found, a wildcard search will be
+  #                               performed to locate and install the first `.pkg`
+  #                               file in the volume.
   # @example: Install vagrant from URL
   #   mymachost.generic_install_dmg('https://releases.hashicorp.com/vagrant/1.8.4/vagrant_1.8.4.dmg', 'Vagrant', 'Vagrant.pkg')
   def generic_install_dmg(dmg_file, pkg_base, pkg_name)
@@ -36,7 +39,7 @@ module Mac::Pkg
         execute("installer -pkg #{specific_pkg_path} -target /")
       else
         # else find and install the first *.pkg file in the volume
-        execute(<<~SCRIPT
+        execute <<~SCRIPT
           found=0
           for pkg in /Volumes/#{pkg_base}/*.pkg; do
             if [ -f "$pkg" ]; then
@@ -53,7 +56,6 @@ module Mac::Pkg
             exit 1
           fi
         SCRIPT
-               )
       end
     end
   end
--- a/lib/beaker/host/mac/pkg.rb
+++ b/lib/beaker/host/mac/pkg.rb
@@ -6,6 +6,8 @@ module Mac::Pkg
   end
 
   def install_package(name, _cmdline_args = '', _version = nil)
+    # strip off any .dmg extension, if it exists
+    name = File.basename(name, '.dmg')
     generic_install_dmg("#{name}.dmg", name, "#{name}.pkg")
   end
 
@@ -25,7 +27,35 @@ module Mac::Pkg
     end
     dmg_name = File.basename(dmg_file, '.dmg')
     execute("hdiutil attach #{dmg_name}.dmg")
-    execute("installer -pkg /Volumes/#{pkg_base}/#{pkg_name} -target /")
+
+    # First check if the specific package exists, otherwise use wildcard
+    specific_pkg_path = "/Volumes/#{pkg_base}/#{pkg_name}"
+    execute("test -f #{specific_pkg_path}", :accept_all_exit_codes => true) do |result|
+      if result.exit_code == 0
+        # $pkg_name package found so install it
+        execute("installer -pkg #{specific_pkg_path} -target /")
+      else
+        # else find and install the first *.pkg file in the volume
+        execute(<<~SCRIPT
+          found=0
+          for pkg in /Volumes/#{pkg_base}/*.pkg; do
+            if [ -f "$pkg" ]; then
+              echo "Installing $pkg"
+              installer -pkg "$pkg" -target /
+              found=1
+              break  # Only install the first package found
+            fi
+          done
+
+          # Return non-zero exit code if no packages were found
+          if [ $found -eq 0 ]; then
+            echo "ERROR: No .pkg files found in /Volumes/#{pkg_base}/"
+            exit 1
+          fi
+        SCRIPT
+               )
+      end
+    end
   end
 
   def uninstall_package(name, _cmdline_args = '')
--- a/lib/beaker/host/mac/pkg.rb
+++ b/lib/beaker/host/mac/pkg.rb
@@ -40,18 +40,12 @@ module Mac::Pkg
       else
         # else find and install the first *.pkg file in the volume
         execute <<~SCRIPT
-          found=0
-          for pkg in /Volumes/#{pkg_base}/*.pkg; do
-            if [ -f "$pkg" ]; then
-              echo "Installing $pkg"
-              installer -pkg "$pkg" -target /
-              found=1
-              break  # Only install the first package found
-            fi
-          done
-
-          # Return non-zero exit code if no packages were found
-          if [ $found -eq 0 ]; then
+          # find the first .pkg file in the mounted volume
+          pkg=$(find /Volumes/#{pkg_base} -name "*.pkg" -type f -print -quit)
+          if [ -n "$pkg" ]; then
+            echo "Installing $pkg"
+            installer -pkg "$pkg" -target /
+          else
             echo "ERROR: No .pkg files found in /Volumes/#{pkg_base}/"
             exit 1
           fi
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -272,6 +272,7 @@ RSpec/SpecFilePathFormat:
     - 'spec/beaker/dsl/wrappers_spec.rb'
     - 'spec/beaker/host/mac/group_spec.rb'
     - 'spec/beaker/host/mac/user_spec.rb'
+    - 'spec/beaker/host/mac/pkg_spec.rb'
     - 'spec/beaker/host/pswindows/user_spec.rb'
     - 'spec/beaker/host/windows/user_spec.rb'
     - 'spec/beaker/host_prebuilt_steps_spec.rb'
--- a/spec/beaker/host/mac/pkg_spec.rb
+++ b/spec/beaker/host/mac/pkg_spec.rb
@@ -1,0 +1,85 @@
+require 'spec_helper'
+
+class MacPkgTest
+  include Mac::Pkg
+  # CommandFactory is included in Mac::Pkg to provide us with the execute method
+  include Beaker::CommandFactory
+
+  def initialize
+    @logger = RSpec::Mocks::Double.new('logger').as_null_object
+
+    # Set up the test host to be used
+    @hostname = "mac.test.com"
+    @options = {}
+    @test_host = Beaker::Host.create(@hostname, @options, {})
+  end
+
+  def logger
+    @logger
+  end
+end
+
+describe MacPkgTest do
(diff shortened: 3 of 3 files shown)


lib/beaker/host/mac/pkg.rb
```diff
@@ -25,7 +27,8 @@ def generic_install_dmg(dmg_file, pkg_base, pkg_name)
     end
     dmg_name = File.basename(dmg_file, '.dmg')
     execute("hdiutil attach #{dmg_name}.dmg")
-    execute("installer -pkg /Volumes/#{pkg_base}/#{pkg_name} -target /")
+    # install the mounted *.pkg even if its basename is different from the *.dmg
+    execute("ls /Volumes/#{pkg_base}/*.pkg | xargs -t -I{} installer -pkg {} -target /")
```
ekohl, line 31: It's generally considered a bad practice to use `ls` in scripts. Since you're using a shell anyway, can't you use a for loop?

gavindidrichsen: Hi @ekohl.  Actually, just making a few more amendments to this..so that it will work if the pkg_name is valid.  Only if the pkg_name is not there, then search and install the *.pkg.  I'm just testing the change now.  

With regard to the `ls`; can you give me an example of how you mean with a for-loop?  I just want to find the *.pkg file that lives beneath /Volumes/#{pkg_base} and then install that.

gavindidrichsen: In other words, something like this:

```ruby
    execute("test -f #{specific_pkg_path}", :accept_all_exit_codes => true) do |result|
      if result.exit_code == 0
        # $pkg_name package found so install it
        execute("installer -pkg #{specific_pkg_path} -target /")
      else
        # else find and install any .pkg file in the volume
        execute("ls /Volumes/#{pkg_base}/*.pkg | xargs -t -I{} installer -pkg {} -target /")
      end
    end
```

gavindidrichsen:
Hi @ekohl.  I've:
* replaced the `ls` with a for-loop.  The new "script" does a bit more so I've laid it out with a HEREDOC for clarity.
* added a new `pkg_spec.rb`
* manually tested a bunch of scenarios as in the PR description

lib/beaker/host/mac/pkg.rb
```diff
@@ -25,7 +27,35 @@ def generic_install_dmg(dmg_file, pkg_base, pkg_name)
...
+
+    # First check if the specific package exists, otherwise use wildcard
+    specific_pkg_path = "/Volumes/#{pkg_base}/#{pkg_name}"
+    execute("test -f #{specific_pkg_path}", :accept_all_exit_codes => true) do |result|
+      if result.exit_code == 0
+        # $pkg_name package found so install it
+        execute("installer -pkg #{specific_pkg_path} -target /")
+      else
+        # else find and install the first *.pkg file in the volume
+        execute(<<~SCRIPT
```
ekohl, line 39: Perhaps leave out the brace since it's not needed?
```suggestion
        execute <<~SCRIPT
```

lib/beaker/host/mac/pkg.rb
```diff
@@ -25,7 +27,35 @@ def generic_install_dmg(dmg_file, pkg_base, pkg_name)
     end
     dmg_name = File.basename(dmg_file, '.dmg')
     execute("hdiutil attach #{dmg_name}.dmg")
-    execute("installer -pkg /Volumes/#{pkg_base}/#{pkg_name} -target /")
+
+    # First check if the specific package exists, otherwise use wildcard
```
ekohl, line 31: IMHO this should at least be documented in the `@param` documentation for `pkg_name`. Perhaps it should even start to default to `nil` and this method should handle it.

Still, I wonder in which case you can't predict the `pkg_name` for a dmg.

lib/beaker/host/mac/pkg.rb
```diff
@@ -25,7 +27,35 @@ def generic_install_dmg(dmg_file, pkg_base, pkg_name)
...
+    execute("test -f #{specific_pkg_path}", :accept_all_exit_codes => true) do |result|
+      if result.exit_code == 0
+        # $pkg_name package found so install it
+        execute("installer -pkg #{specific_pkg_path} -target /")
+      else
+        # else find and install the first *.pkg file in the volume
+        execute(<<~SCRIPT
+          found=0
+          for pkg in /Volumes/#{pkg_base}/*.pkg; do
+            if [ -f "$pkg" ]; then
```
bastelfreak, line 42: that if shouldn't be required. The globbing from the `for` loop above only returns files that exist?

gavindidrichsen: Hi @bastelfreak .  I wanted to keep the if so that it "breaks" out of the for-loop on find of first `*.pkg`.  I suspect a `*.dmg' would only have one internal `*.pkg` but I didn't want to take the chance of multiple installs occurring.

lib/beaker/host/mac/pkg.rb
gavindidrichsen: Changed to for-loop

lib/beaker/host/mac/pkg.rb
gavindidrichsen: Removed un-needed paranthesis.

lib/beaker/host/mac/pkg.rb
gavindidrichsen: Hi @ekohl.  I updated the  `@param` documentation for `pkg_name` but didn't make it `nil` by default.  My thinking here is that `Mac::Pkg.install_package()` seems to be the "public" interface into this at which point it always passes the `pkg_name` to `generic_install_dmg()`.  You ok with this?

lib/beaker/host/mac/pkg.rb
gavindidrichsen: Hi @bastelfreak . I tested just the `for` loop and the process continues without error.  For example, I changed the `execute` to glob on a non-existent file:

```bash
        execute <<~SCRIPT
          for pkg in /Volumes/#{pkg_base}/*.poo; do
            installer -pkg "$pkg" -target /
            break  # Only install the first package found
          done
        SCRIPT
      end
```

and the flow does not exit as I would have expected; it fails with something else downstream.  For example:

```bash
Install local package puppet-bolt-4.0.0-1.osx14.dmg
    localhost $ scp puppet-bolt-4.0.0-1.osx14.dmg osx14-64-1:puppet-bolt-4.0.0-1.osx14.dmg {:ignore => }
    
    hard-pepperoni.delivery.puppetlabs.net (osx14-64-1) 09:38:49$ test -f puppet-bolt-4.0.0-1.osx14.dmg
    
    hard-pepperoni.delivery.puppetlabs.net (osx14-64-1) executed in 0.44 seconds
    
    hard-pepperoni.delivery.puppetlabs.net (osx14-64-1) 09:38:50$ hdiutil attach puppet-bolt-4.0.0-1.osx14.dmg
      Checksumming Protective Master Boot Record (MBR : 0)…
      Protective Master Boot Record (MBR :: verified   CRC32 $A63E199D
      Checksumming GPT Header (Primary GPT Header : 1)…
       GPT Header (Primary GPT Header : 1): verified   CRC32 $D9C1386A
      Checksumming GPT Partition Data (Primary GPT Table : 2)…
      GPT Partition Data (Primary GPT Tabl: verified   CRC32 $E8EBFC16
      Checksumming  (Apple_Free : 3)…
                          (Apple_Free : 3): verified   CRC32 $00000000
      Checksumming disk image (Apple_APFS : 4)…
               disk image (Apple_APFS : 4): verified   CRC32 $88F0A322
      Checksumming  (Apple_Free : 5)…
                          (Apple_Free : 5): verified   CRC32 $00000000
      Checksumming GPT Partition Data (Backup GPT Table : 6)…
      GPT Partition Data (Backup GPT Table: verified   CRC32 $E8EBFC16
      Checksumming GPT Header (Backup GPT Header : 7)…
        GPT Header (Backup GPT Header : 7): verified   CRC32 $460C67C4
      verified   CRC32 $1223A72E
      /dev/disk2                GUID_partition_scheme          
      /dev/disk2s1              Apple_APFS                     
      /dev/disk3                EF57347C-0000-11AA-AA11-0030654
      /dev/disk3s1              41504653-0000-11AA-AA11-0030654 /Volumes/puppet-bolt-4.0.0-1.osx14
    
    hard-pepperoni.delivery.puppetlabs.net (osx14-64-1) executed in 1.94 seconds
    
    hard-pepperoni.delivery.puppetlabs.net (osx14-64-1) 09:38:52$ test -f /Volumes/puppet-bolt-4.0.0-1.osx14/puppet-bolt-4.0.0-1.osx14.pkg
    
    hard-pepperoni.delivery.puppetlabs.net (osx14-64-1) executed in 0.47 seconds
    Exited: 1
    
    hard-pepperoni.delivery.puppetlabs.net (osx14-64-1) 09:38:52$ for pkg in /Volumes/puppet-bolt-4.0.0-1.osx14/*.poo; do
      installer -pkg "$pkg" -target /
      break  # Only install the first package found
    done
      installer: Error - the package path specified was invalid: '/Volumes/puppet-bolt-4.0.0-1.osx14/*.poo'.
    
    hard-pepperoni.delivery.puppetlabs.net (osx14-64-1) executed in 0.64 seconds
Begin teardown
End teardown
setup/package/pre-suite/020_install.rb passed in 29.29 seconds
Begin setup/common/pre-suite/030_set_password.rb

Set root/Administrator password to a known value
  
  hard-pepperoni.delivery.puppetlabs.net (osx14-64-1) 09:38:53$ echo "PermitRootLogin yes" >> /etc/ssh/sshd_config
...
...
Warning: setup/common/pre-suite/031_add_local_nix_user.rb errored in 4.59 seconds
      Test Suite: pre_suite @ 2025-04-14 09:38:23 +0100

      - Host Configuration Summary -


              - Test Case Summary for suite 'pre_suite' -
       Total Suite Time: 36.48 seconds
      Average Test Time: 12.16 seconds
              Attempted: 3
                 Passed: 2
                 Failed: 0
                Errored: 1
                Skipped: 0
                Pending: 0
                  Total: 5

      - Specific Test Case Status -
        
Failed Tests Cases:
Errored Tests Cases:
  Test Case setup/common/pre-suite/031_add_local_nix_user.rb reported: #<Beaker::Host::CommandFailure:"Host 'hard-pepperoni.delivery.puppetlabs.net' exited with 127 running:\n /opt/puppetlabs/bolt/bin/puppet apply /var/folders/zz/zyxvpxvq6csfxvn_n0000000000000/T/osx_manifest.XXXXXX.hLD5zUfTEH/osx_user.pp\nLast 10 lines of output were:\n\tsh: /opt/puppetlabs/bolt/bin/puppet: No such file or directory">
    Test line: setup/common/pre-suite/031_add_local_nix_user.rb:29:in `block (3 levels) in run_test'
Skipped Tests Cases:
Pending Tests Cases:
```

I think we should keep the error trap so that if no DMG is installed then we error out.

bastelfreak (approved):

PR #1917, opened 2025-05-15
Add Ruby 3.4 to CI & Pin to Ubuntu 24.04 runners

bastelfreak:


beaker.gemspec
```diff
@@ -26,6 +26,7 @@ Gem::Specification.new do |s|
 
   # Run time dependencies
   # we cannot require 1.0.2 because that requires Ruby 3.1
+  s.add_dependency 'base64', '~> 0.2.0'
```
ekohl, line 29: The comment above this line refers to `minitar`. You should keep those together

acceptance/tests/base/dsl/structure_test.rb
```diff
@@ -73,7 +73,7 @@
 
       fail "#confine did not skip test but should have."
     rescue Beaker::DSL::Outcomes::SkipTest => e
-      assert_match(/No suitable hosts found with {:platform=>"test"}/, e.message, "#confine raised an unexpected skip_test")
+      assert_match(/No suitable hosts found with.*platform.*"test"}/, e.message, "#confine raised an unexpected skip_test")
```
ekohl, line 76: Wouldn't it at least keep the `{` from before?
```suggestion
      assert_match(/No suitable hosts found with {.*platform.*"test"}/, e.message, "#confine raised an unexpected skip_test")
```

And I guess it's `:platform=>` vs `platform:` so can we capture it as follows?
```suggestion
      assert_match(/No suitable hosts found with {(:platform=>|platform:)"test"}/, e.message, "#confine raised an unexpected skip_test")
```

acceptance/tests/base/dsl/structure_test.rb
```diff
@@ -83,7 +83,7 @@
 
       fail "#confine did not skip test but should have."
     rescue Beaker::DSL::Outcomes::SkipTest => e
-      assert_match(/No suitable hosts found without {:platform=>"#{default['platform']}"}/, e.message, "#confine raised an unexpected
+      assert_match(/No suitable hosts found without {.*platform.*"#{default['platform']}"}/, e.message, "#confine raised an unexpected
```
ekohl, line 86: Same comment as above

rwaffen (approved):

PR #1924, opened 2025-05-27
excon: Exclude 1.2.6

bastelfreak:
excon is a dependency of docker-api
1.2.6 is broken and needs to be excluded
we need to allow 1.2.5 because newer versions require Ruby 3.1 https://github.com/excon/excon/issues/884


bastelfreak:
https://github.com/voxpupuli/beaker/pull/1925 & https://github.com/voxpupuli/beaker-docker/pull/156

PR #1926, opened 2025-05-27
Fix wrong variable assignment in multiple places

bastelfreak:
We discovered those bugs during https://github.com/voxpupuli/beaker/pull/1919/


ekohl (commented):
Do we have test coverage for this? I don't think we run the acceptance tests now.

bastelfreak:
I mean it didn't raise failing tests in the past, so I guess there are no tests :(

ekohl (approved):
Previously it worked because it was just passing positional parameters. I'm a bit afraid of breaking that accidentally. Keyword arguments in Ruby are less intuitive than Python. Just making sure for myself:

```ruby
def x(value, print = nil)
  puts value if print
end

x('should not print')
x('should print', true)
x('should print', print: true)
```

This works on Ruby 2.7:

```console
$ ruby --version
ruby 2.7.8p225 (2023-03-30 revision 1f4d455848) [x86_64-linux]
$ ruby test.rb 
should print
should print
```

And on Ruby 3.4:

```console
$ ruby --version
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +PRISM [x86_64-linux]
$ ruby test.rb 
should print
should print
```

So I'm confident this is ok.

PR #1935, opened 2025-06-20
Update RHEL10 packages

skyamgarp:
This is to use chrony as RHEL 10 failing with installing `ntpdate`

### DIFF
--- a/lib/beaker/host_prebuilt_steps.rb
+++ b/lib/beaker/host_prebuilt_steps.rb
@@ -119,7 +119,7 @@ module Beaker
         AMAZON2023_PACKAGES
       when /el-8/
         RHEL8_PACKAGES
-      when /el-9/
+      when /el-(9|[1-9][0-9])/
         RHEL9_PACKAGES
       when /sles-10/
         SLES10_PACKAGES


skyamgarp:
@bastelfreak Can you please review?

lib/beaker/host_prebuilt_steps.rb
```diff
@@ -119,7 +119,7 @@ def host_packages(host)
         AMAZON2023_PACKAGES
       when /el-8/
         RHEL8_PACKAGES
-      when /el-9/
+      when /el-(9|10)/
```
bastelfreak, line 122: Does it make sense to stay future proof and match on el9 and any newer version? With a bit of luck we won't have to touch it again on EL11.

skyamgarp: Updated.

ekohl: This is the 4.x branch. In master it's already future proof via 18e6e2292aeed5557c6a85e45ba86365ca43d6f6, which was part of 6.4.0.

skyamgarp: > Do you actually need this in the beaker 4 branch? Can't you update to latest? It not, can you try to cherry-pick [18e6e22](https://github.com/voxpupuli/beaker/commit/18e6e2292aeed5557c6a85e45ba86365ca43d6f6)?

Hi @bastelfreak , Yes we need this in 4.x. It might take some time to upgrade to beaker latest.

PR #1946, opened 2025-08-12
Move noask_file_text to platform

ekohl:
This really depends on the platform and doesn't belong in the file abstraction.


lib/beaker/host/unix/pkg.rb
```diff
@@ -107,7 +107,7 @@ def install_package(name, cmdline_args = '', version = nil, opts = {})
     when /solaris-10/
       if !check_for_command('pkgutil')
         # https://www.opencsw.org/package/pkgutil/
-        noask_text = self.noask_file_text
+        noask_text = self['platform'].noask_file_text
```
ekohl, line 110: This is the only use. Should it instead move to a private helper in this class?

PR #1950, opened 2025-08-12
Drop Solaris support

ekohl:
As suggested in https://github.com/voxpupuli/beaker/pull/1946#issuecomment-3179524063. I'm not sure how much Solaris is still being used and it's a lot of code to keep around.


bastelfreak (approved):

joshcooper:
@bastelfreak fwiw perforce still supports Solaris (and AIX)

PR #1954, opened 2025-09-18
Avoid duplicating PermitUserEnvironment

### LINKED ISSUE #1953: Duplicate PermitUserEnvironment in /etc/ssh/sshd_config after multiple run
I'm now using the local mode to run beaker tests and after running the tests multiple times the sshd_config file gets duplicated `PermitUserEnvironment yes` lines.

```
[root@localhost ~]# head -12 /etc/ssh/sshd_config
PermitUserEnvironment yes
PermitUserEnvironment yes
PermitUserEnvironment yes
PermitUserEnvironment yes
PermitUserEnvironment yes
PermitUserEnvironment yes
PermitUserEnvironment yes
PermitUserEnvironment yes
#	$OpenBSD: sshd_config,v 1.104 2021/07/02 05:11:21 dtucker Exp $
```

kajinamit:
Do not add a new line if the option is already defined. This avoids adding dupulicated lines and also make sure that the existing explicit setting to disable PermitUserEnvironment is purged.

Closes #1953

### DIFF
--- a/lib/beaker/host/unix/exec.rb
+++ b/lib/beaker/host/unix/exec.rb
@@ -302,7 +302,7 @@ module Unix::Exec
     case self['platform']
     when /amazon|debian|ubuntu|archlinux|el-|centos|fedora|redhat|oracle|scientific|opensuse|sles|solaris/
       directory = tmpdir
-      exec(Beaker::Command.new("echo 'PermitUserEnvironment yes' | cat - /etc/ssh/sshd_config > #{directory}/sshd_config.permit"))
+      exec(Beaker::Command.new("sed -e 's/^PermitUserEnvironment .*/PermitUserEnvironment yes/' -e t -e '1s/^/PermitUserEnvironment yes\\n/' /etc/ssh/sshd_config > #{directory}/sshd_config.permit"))
       exec(Beaker::Command.new("mv #{directory}/sshd_config.permit /etc/ssh/sshd_config"))
       exec(Beaker::Command.new("echo '' >/etc/environment")) if self['platform'].include?('ubuntu-')
     when /(free|open)bsd/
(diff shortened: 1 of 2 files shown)


lib/beaker/host/unix/exec.rb
```diff
@@ -301,9 +301,7 @@ def ssh_service_restart
   def ssh_permit_user_environment
     case self['platform']
     when /amazon|debian|ubuntu|archlinux|el-|centos|fedora|redhat|oracle|scientific|opensuse|sles|solaris/
-      directory = tmpdir
-      exec(Beaker::Command.new("echo 'PermitUserEnvironment yes' | cat - /etc/ssh/sshd_config > #{directory}/sshd_config.permit"))
-      exec(Beaker::Command.new("mv #{directory}/sshd_config.permit /etc/ssh/sshd_config"))
+      exec(Beaker::Command.new("sed -i -e 's/^PermitUserEnvironment .*/PermitUserEnvironment yes/' -e t -e '1s/^/PermitUserEnvironment yes\\n/' /etc/ssh/sshd_config"))
```
smortex, line 304: I am concerned because `sed -i` is not portable across sed(1) implementations.  Specifically, FreeBSD use standard getopt(3) and the `-i` option _require_ a value (so it will use `-e` in this case, shifting all arguments and making the command-line invalid).  GNU sed(1) do some magic and only use anything that is "glued" to `-i` and does not support passing an empty value "the regular way" with `-i ''`.  So the only common supported syntax is `-isomething` where something is not blank.

Rather than in-place edits, when I write portable code, I generally prefer to have sed print the modified content to stdout, redirect it to a temporary file and move that file over the original one like it was done before.

Currently we do not support running acceptance tests on FreeBSD, but I would love to see that one day

PR #1958, opened 2025-10-16
Remove default use of wmic to determine arch

silug:
WMIC has been deprecated for some time, and it has been removed in
Windows 2025.

With this change, the default path for `identify_windows_architecture`
avoids WMIC.

### DIFF
--- a/lib/beaker/host/pswindows/pkg.rb
+++ b/lib/beaker/host/pswindows/pkg.rb
@@ -36,63 +36,7 @@ module PSWindows::Pkg
   def identify_windows_architecture
     arch = nil
     execute('echo %PROCESSOR_ARCHITECTURE%', :accept_all_exit_codes => true) do |result|
-      arch = if result.exit_code == 0
-               result.stdout.strip
-             else
-               identify_windows_architecture_wmic
-             end
-    end
-
-    arch = 'AMD64' if arch.include?('EM64T') # Normalize Intel 64-bit architecture name
-
-    arch
-  end
-
-  # @api private
-  def identify_windows_architecture_wmic
-    execute("wmic cpu get architecture", :accept_all_exit_codes => true) do |result|
-      if result.exit_code == 0
-        cpu = result.lines(chomp: true).grep(%r{^\d+$}).first
-        case cpu
-        # We only care about '9' (x64) and '0' (x86) here,
-        # but others are included for completeness
-        when '0'
-          return 'X86'
-        when '1'
-          return 'MIPS'
-        when '2'
-          return 'ALPHA'
-        when '3'
-          return 'POWERPC'
-        when '5'
-          return 'ARM'
-        when '6'
-          return 'IA64'
-        when '9'
-          return 'AMD64'
-        when '12'
-          return 'ARM64'
-        end
-      end
-    end
-
-    # Fallback for backwards compatibility
-    arch = nil
-    execute("wmic os get osarchitecture", :accept_all_exit_codes => true) do |result|
-      arch = if result.exit_code == 0
-               result.stdout.include?('64') ? 'AMD64' : 'X86'
-             else
-               identify_windows_architecture_from_os_name_for_win2003
-             end
-    end
-    arch
-  end
-
-  # @api private
-  def identify_windows_architecture_from_os_name_for_win2003
-    arch = nil
-    execute("wmic os get name", :accept_all_exit_codes => true) do |result|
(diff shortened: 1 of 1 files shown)


michael-riddle (approved):

lib/beaker/host/pswindows/pkg.rb
```diff
@@ -34,6 +34,19 @@ def determine_if_x86_64
 
   # @api private
   def identify_windows_architecture
+    arch = nil
+    execute('echo %PROCESSOR_ARCHITECTURE%', :accept_all_exit_codes => true) do |result|
+      arch = if result.exit_code == 0
+               result.stdout.include?('64') ? '64' : '32'
```
ekohl: This can be confused by ARM64, which isn't x86_64

ekohl: On my phone now, but when I looked it up I found it returns `ARM64` which also includes `64`. That's what I meant with my comment.

ekohl: https://learn.microsoft.com/en-us/windows/win32/winprog64/wow64-implementation-details#environment-variables lists the following valid values:
* `AMD64`
* `IA64`
* `ARM64`
* `x86`

So how about `identify_windows_architecture` only returns the output of `echo %PROCESSOR_ARCHITECTURE%` (and returns `nil` if the command failed). Then `determine_if_x86_64` is updated to something like:

```ruby
def determine_if_x86_64
  case identify_windows_architecture
  when 'AMD64'
    true
  when nil
    identify_windows_architecture_wmic.include?('64')
  else
    false
  end
end
```

Though another question we should ask ourselves: do we even need this whole method to determine if a host is x86_64? It's only used in [beaker-puppet](https://github.com/puppetlabs/beaker-puppet) and [beaker-pe](https://github.com/puppetlabs/beaker-pe). Does Perforce even support 32bit Windows? AFAIK all those versions are pretty much EOL now.

Short term I don't mind fixing this, but IMHO we can also deprecate the Beaker methods and remove them in the next major version.

silug: I'm sorry, I completely missed that you said `ARM64`, not `AMD64`.

I can update my change to be more correct, but I have no idea if that will be enough to make ARM work.

lib/beaker/host/pswindows/pkg.rb
```diff
@@ -27,17 +27,60 @@ def uninstall_package(_name, _cmdline_args = '')
...
+    arch = nil
+    execute('echo %PROCESSOR_ARCHITECTURE%', :accept_all_exit_codes => true) do |result|
+      arch = if result.exit_code == 0
+               result.stdout.strip
+             else
+               identify_windows_architecture_wmic
+             end
+    end
+
+    arch = 'AMD64' if arch.include?('EM64T') # Normalize Intel 64-bit architecture name
```
ekohl: The documentation said that was only for Windows XP so I'm not sure we need to include this.

lib/beaker/host/pswindows/pkg.rb
```diff
@@ -27,17 +27,60 @@ def uninstall_package(_name, _cmdline_args = '')
...
   private
 
   # @api private
   def identify_windows_architecture
+    arch = nil
+    execute('echo %PROCESSOR_ARCHITECTURE%', :accept_all_exit_codes => true) do |result|
+      arch = if result.exit_code == 0
+               result.stdout.strip
+             else
+               identify_windows_architecture_wmic
```
ekohl: Do you know when this would fail? I get the impression that `PROCESSOR_ARCHITECTURE` should always be available. I'd be OK with dropping any `wmic` code if we can. Reading your comment I get the feeling we can.

lib/beaker/host/pswindows/pkg.rb
```diff
@@ -27,17 +27,60 @@ def uninstall_package(_name, _cmdline_args = '')
...
+  # @api private
+  def identify_windows_architecture_wmic
+    execute("wmic cpu get architecture", :accept_all_exit_codes => true) do |result|
+      if result.exit_code == 0
+        cpu = result.lines(chomp: true).grep(%r{^\d+$}).first
+        case cpu
+        # We only care about '9' (x64) and '0' (x86) here,
+        # but others are included for completeness
+        when '0'
+          return 'X86'
```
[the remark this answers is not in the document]
silug: I dropped this whole code path.

PR #1996, opened 2026-04-01
Fix Debian 13 NTP setup by using chrony

treydock:
This is the error Debian 13 gives:

{"stream":"Package ntpdate is not available, but is referred to by another package.\nThis may mean that the package is missing, has been obsoleted, or\nis only available from another source\nHowever the following packages replace it:\n  ntpsec-ntpdate\n\n"}

For Debian 13+, use chrony

### DIFF
--- a/lib/beaker/platform.rb
+++ b/lib/beaker/platform.rb
@@ -108,6 +108,8 @@ module Beaker
         true
       when 'el'
         @version.to_i >= 8
+      when 'debian'
+        @version.to_i >= 13
       else
         false
       end


bastelfreak:
This makes sense to me, but I'm wondering why we never ran into this? Vox Pupuli uses Debian 13 acceptance tests singe many months now.

treydock:
Not sure, we use beaker and beaker-docker to run E2E tests for our Open OnDemand product and ran into it here: https://github.com/OSC/ondemand/actions/runs/23854542491/job/69543747938

We were on somewhat old versions of both beaker and beaker-docker but from what I can see the ntpdate logic affects recent versions too.

treydock:
What Puppet modules have passing Debian 13 tests?  I can take a look at their Github Action output to see if anything jumps out at me.

bastelfreak:
One example is the aptly one: https://github.com/voxpupuli/puppet-aptly/actions/runs/23663978270/job/68941305013

treydock:
The apt installs on those looks a lot different than what I saw and don't install NTP or chrony things.  I updated things on our end to latest beaker and beaker-docker and didn't see the problem, so seems like maybe this NTP package code path isn't even used.  I'm fine if you'd rather just close this PR as not needed.

ekohl:
> This makes sense to me, but I'm wondering why we never ran into this? Vox Pupuli uses Debian 13 acceptance tests singe many months now.

I don't think we actually enable timesync in our Beaker testing because it's off by default. Since 18e6e2292aeed5557c6a85e45ba86365ca43d6f6 the package is only installed when needed.

bwitt:
yes this also affects Ubuntu 26.04.

bwitt:
See also https://github.com/voxpupuli/beaker/pull/2036 for Ubuntu 26.04

ekohl (approved):

PR #2008, opened 2026-07-12
Update puppet_configprint helper

Sharpie:
This commit updates the `puppet_configprint` helper method to call through to `puppet config print` instead of `puppet <subcommand> --configprint`. The `--configprint` option was deprecated in Puppet 5.x and was finally retired in OpenVoxProject/openvox#374.

### DIFF
--- a/lib/beaker/host.rb
+++ b/lib/beaker/host.rb
@@ -21,15 +21,15 @@ module Beaker
     class RebootFailure < CommandFailure; end
     class RebootWarning < StandardError; end
 
-    # This class provides array syntax for using puppet --configprint on a host
+    # This class provides array syntax for using "puppet config print" on a host
     class PuppetConfigReader
-      def initialize(host, command)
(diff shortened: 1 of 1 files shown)


lib/beaker/host.rb
```diff
@@ -21,23 +21,23 @@ class CommandFailure < StandardError; end
...
+        cmd = PuppetCommand.new('config', 'print', "--section=#{@section}")
         keys = @host.exec(cmd).stdout.split("\n").collect do |x|
           x[/^[^\s]+/]
         end
         keys.include?(k)
       end
 
       def [](k)
-        cmd = PuppetCommand.new(@command, "--configprint #{k}")
+        cmd = PuppetCommand.new('config', 'print' "--section=#{@section}", k)
```
ekohl:
```suggestion
        cmd = PuppetCommand.new('config', 'print', "--section=#{@section}", k)
```

PR #2029, opened 2026-08-24
Fix Solaris sshd_config editing in ssh_permit_user_environment

shubhamshinde360:
## Summary

Solaris 10's `/usr/bin/sed` is POSIX-only and does not support `\n` in replacement strings (a GNU sed extension). The sed one-liner introduced in #1953 (v7.4.0) silently produces a corrupt `sshd_config` on Solaris 10, causing sshd to fail to restart and leaving port 22 unreachable.

Replace the GNU sed idiom with a POSIX-compatible `grep` + `sed`/`cp`+`echo` approach:
- If `PermitUserEnvironment` is already present: use a simple `sed` substitution (no `\n` needed)
- If absent: `cp` the config and `echo >>` to append the directive

## Root cause

Commit 60e3e664 (shipped in v7.4.0) changed:
```sh
# beaker 6 / pre-7.4.0 — POSIX-compatible, worked on Solaris 10
echo 'PermitUserEnvironment yes' | cat - /etc/ssh/sshd_config > sshd_config.permit
```
to:
```sh
# v7.4.0+ — GNU sed extension, breaks Solaris 10
sed -e 's/^PermitUserEnvironment .*/PermitUserEnvironment yes/' -e t -e '1s/^/PermitUserEnvironment yes\n/' /etc/ssh/sshd_config
```

## Test plan
- [ ] Run beaker against a Solaris 10 host and verify `ssh_permit_user_environment` completes successfully
- [ ] Verify `PermitUserEnvironment yes` is correctly set in `/etc/ssh/sshd_config` both when the directive already exists and when it doesn't
- [ ] Confirm non-Solaris platforms (Linux, BSD) are unaffected

🤖 Generated with [Claude Code](https://claude.com/claude-code)


amitkarsale (approved):

lib/beaker/host/unix/exec.rb
```diff
@@ -302,7 +308,15 @@ def ssh_permit_user_environment
     case self['platform']
     when /amazon|debian|ubuntu|archlinux|el-|centos|fedora|redhat|oracle|scientific|opensuse|sles|solaris/
       directory = tmpdir
-      exec(Beaker::Command.new("sed -e 's/^PermitUserEnvironment .*/PermitUserEnvironment yes/' -e t -e '1s/^/PermitUserEnvironment yes\\n/' /etc/ssh/sshd_config > #{directory}/sshd_config.permit"))
+      # Use grep to check if PermitUserEnvironment is already in the config.
+      # Solaris 10 sed does not support \n in replacement strings, so we
+      # cannot use a single sed to both substitute and insert.
+      cmd = "if grep -q '^PermitUserEnvironment' /etc/ssh/sshd_config; " \
```
bastelfreak, line 314: This looks a bit overly complicated. I'm wondering why `\n` is even in the original string. Is that really required?

shubhamshinde360: Hi @bastelfreak 
Yes,
'1s/^/PermitUserEnvironment yes\n/' inserts the directive as its own line at the top of the file. Without the \n, the prepended text would be concatenated directly onto the first line of the file with no newline separator.

joshcooper: Can we leave this as-is and use perl on Solaris 10 only, like we do for free/openbsd:

```
-bash-3.2# uname -a
SunOS pix-sol10-1 5.10 Generic_147147-26 sun4v sparc sun4v
-bash-3.2# grep PermitUser sshd_config 
PermitUserEnvironment no
-bash-3.2# perl -pi -e 's/^#?PermitUserEnvironment no/PermitUserEnvironment yes/' sshd_config
-bash-3.2# grep PermitUser sshd_config 
PermitUserEnvironment yes
```

joshcooper (approved):

shubhamshinde360:
Thanks @bastelfreak,
I have signed it. Github shows the commit now as verified.
Could you please review. Thanks!

joshcooper (approved):
This looks fine to me but could you amend the commit to mention this affects all Solaris platforms, not just 10 SPARC?

shubhamshinde360:
Hey @joshcooper, 
Updated the commit message. Thanks!

PR #2037, opened 2026-09-09
Create root's dotfiles on EL 8+ hosts (the other half of #2033)

silug:
Pull request #2033 added `rootfiles` to the EL 8+ base packages so that `/root/.bashrc` would exist and commands run over ssh would pick up `/etc/profile.d` (and with it PATH additions such as `/etc/profile.d/puppet-agent.sh`). That was only half of the fix: on the CentOS Stream 9 container image the package installs cleanly and `rpm -V rootfiles` is silent, yet `/root/.bashrc` is still absent and `puppet` is still "command not found".

Since rootfiles 8.1-32 (RHEL-58760) the dotfiles are no longer shipped as regular files. They are `%ghost` entries that systemd-tmpfiles copies out of `/usr/share/rootfiles` at boot, per `/usr/lib/tmpfiles.d/rootfiles.conf`. Nothing else runs that rule: the EL 9 systemd package carries no tmpfiles file trigger, so installing the package on a running system creates the templates and the rule but never the files. A container is never booted, so in a container that is where it stays. The earlier change assumed a trigger would cover the image build path; it does not.

This adds the missing step to validate_host: on EL 8+, after the base packages are in place, apply the rule the way boot would with `systemd-tmpfiles --create rootfiles.conf`. Where systemd-tmpfiles is not installed (minimal images), fall back to copying the templates from `/usr/share/rootfiles` into `/root`. Both paths create only files that are missing, so the step behaves the same whether `rootfiles.conf` exists or not:

  * container, tmpfiles-based rootfiles ....... files created (the fix)
  * booted VM, tmpfiles-based rootfiles ....... no-op, tmpfiles already ran
  * hosts whose rootfiles ships the files ..... no-op, no rule, no templates
  * a customised /root/.bashrc ................ left untouched

Verified in a quay.io/centos/centos:stream9 container built the way beaker-docker builds it (systemd is present there, pulled in through openssh-server): before the step an ssh session has PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin and no LANG; after it, PATH includes /root/bin, LANG=C.UTF-8, and a profile.d PATH drop-in makes its command resolvable over ssh.

The exact command string was then run on ten container images after `dnf install rootfiles`, with a marker line appended to any existing `.bashrc` so that a rewrite would show up as a checksum change. Every run exited 0 with no output on stdout or stderr:

| image            | rootfiles | systemd | rule | dotfiles | result                |
| -----            | --------- | ------- | ---- | -------- | ------                |
| Rocky 8          | 8.1-22    | yes     | no   | 5 -> 5   | unchanged             |
| Alma 8           | 8.1-22    | yes     | no   | 5 -> 5   | unchanged             |
| Rocky 9          | 8.1-35    | no      | yes  | 5 -> 5   | unchanged (copy path) |
| Alma 9           | 8.1-35    | yes     | yes  | 5 -> 5   | unchanged             |
| UBI 9            | 8.1-35    | yes     | yes  | 5 -> 5   | unchanged             |
| Oracle 9         | 8.1-35    | yes     | yes  | 0 -> 5   | created               |
| CentOS Stream 9  | 8.1-35    | yes     | yes  | 0 -> 5   | created               |
| Rocky 10         | 8.1-54    | no      | yes  | 5 -> 5   | unchanged (copy path) |
| Alma 10          | 8.1-54    | yes     | yes  | 5 -> 5   | unchanged             |
| CentOS Stream 10 | 8.1-54    | yes     | yes  | 0 -> 5   | created               |

Oracle Linux 9 turns out to ship without the dotfiles too, so it was affected in the same way as CentOS Stream. Nothing about where `puppet` lives is assumed, so OS packages in /usr/bin and any future OpenVox layout work the same way.

_Generated by Claude Code_

### DIFF
--- a/lib/beaker/host_prebuilt_steps.rb
+++ b/lib/beaker/host_prebuilt_steps.rb
@@ -85,13 +85,13 @@ module Beaker
       report_and_raise(logger, e, "validate")
     end
 
-    # Make sure root's shell dotfiles exist on EL 8+ hosts.
+    # Make sure root's shell dotfiles exist on EL 8+ and Fedora hosts.
     #
-    # Since rootfiles 8.1-32 (EL 9+), `/root/.bashrc` and friends are no
-    # longer shipped as regular files: they are %ghost entries that
-    # systemd-tmpfiles copies out of `/usr/share/rootfiles` at boot, per
-    # `/usr/lib/tmpfiles.d/rootfiles.conf`. Nothing else runs that rule --
-    # the EL 9 systemd package carries no tmpfiles file trigger -- so in a
+    # Since rootfiles 8.1-32 (EL 9+) and 9.0 (Fedora 43+), `/root/.bashrc`
+    # and friends are no longer shipped as regular files: they are %ghost
+    # entries that systemd-tmpfiles copies out of `/usr/share/rootfiles` at
+    # boot, per `/usr/lib/tmpfiles.d/rootfiles.conf`. Nothing else runs that
+    # rule -- the systemd package carries no tmpfiles file trigger -- so in a
     # container, which is never booted, installing the package leaves /root
     # without the dotfiles.
     #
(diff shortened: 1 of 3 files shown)


bastelfreak (approved):
@joshcooper might be relevant for you as well

